PR #101, opened 2015-03-03
Add npm start so gulp global isn't necessary

gesa:


toolness:
Good idea! I actually tried doing this but my machine was having problems finding gulp... which was odd because `npm test` runs gulp and that works... Anyhow, would you be interested in submitting a PR for this?

gesa:
I did! running npm scripts automatically adds `node_modules` to your `$PATH` so this should run fine. (thanks @k88hudson for teaching me that)

toolness:
ohhh sweet. sorry, didn't realize this was actually a PR--Merging now!

PR #111, opened 2015-03-03
Support multiple sidebar colours w/ transition effect

toolness:
This attempts to add support for multiple sidebar colors.

If the user's browser supports `pushState`, we don't actually need to unload the page, and can therefore take advantage of CSS transitions, as shown in this video:

<a href="https://toolness-media.herokuapp.com/v/mozilla-learning-sidebar-transition/play/loop">Play 'Mozilla Learning Sidebar Transition' video</a>

### DIFF
--- a/lib/homepage.jsx
+++ b/lib/homepage.jsx
@@ -66,19 +66,15 @@ var Values = React.createClass({
   }
 });
 
-var Homepage = React.createClass({
-  render: function() {
-    return (
-      <Page>
-        <HeroUnit image="img/hero-unit.jpg">
-          <h1>Unlock opportunities for all citizens of the Web.</h1>
-          <div><Ia href="/foo/" className="btn btn-awsm">Join Us</Ia></div>
-        </HeroUnit>
-        <Values/>
-        <CaseStudies/>
-      </Page>
-    );
-  }
-});
-
-module.exports = Homepage;
+module.exports = function() {
+  return (
+    <Page>
+      <HeroUnit image="img/hero-unit.jpg">
+        <h1>Unlock opportunities for all citizens of the Web.</h1>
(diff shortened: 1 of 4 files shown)


lib/homepage.jsx
```diff
@@ -66,19 +66,15 @@ var Values = React.createClass({
...
-        </HeroUnit>
-        <Values/>
-        <CaseStudies/>
-      </Page>
-    );
-  }
-});
-
-module.exports = Homepage;
+module.exports = function() {
```
toolness: So this is definitely the most unintuitive part of this commit.

Due to the way that [React's reconciliation algorithm](http://facebook.github.io/react/docs/reconciliation.html) works, we need to have a `<Page>` component be the top-level component of all our pages if we want to have cool transitions between them. If we don't, React's reconciliation algorithm will think we're rendering completely different pages and rebuild the DOM of the new one from scratch, thus not triggering any CSS transition. Hence we need to get rid of the `<Homepage>` top-level component.

That said, I'm not a big fan of the particular *way* that I got rid of the `Homepage` component. Might be better to have `Homepage` be a component that goes *inside* a page (and is wrapped by a `<div>`); that way it can still be a React component rather than a weird factory function, and we can still have our cool transition effects.

toolness:
Also, if someone wants to argue that the CSS transition is super distracting, that's fine! I partly just implemented it because I thought it would look cool from a technical standpoint--it may not actually be that great from a usability perspective.

lib/homepage.jsx
toolness: This gets solved easily by migrating to the ol' `react-router`, which we're doing in #132. So hooray for that.

toolness:
Ok, now that we've migrated to `react-router`, the stuff accomplished in this PR will need to be done differently. I'll close this one for now and re-open a new one in a little bit.

PR #128, opened 2015-03-04
[WIP] Integrate react-a11y

toolness:
Currently in the process of integrating [`react-a11y`](https://github.com/rackt/react-a11y#readme) into the build system, and ensuring that no accessibility warnings are raised during the site build process. If they do, travis will complain.

Tasks left to do:
- [ ] Fix all occurrences of `You have a click handler on a non-interactive element but no 'role' DOM property. It will be unclear what this element is supposed to do to a screen-reader user. http://www.w3.org/TR/wai-aria/roles#role_definitions`
- [ ] Fix all occurrences of `You have a click handler on a non-interactive element but no 'tabIndex' DOM property. The element will not be navigable or interactive by keyboard users. http://www.w3.org/TR/wai-aria-practices/#focus_tabindex`
- [ ] Fix all occurrences of `You have a click handler on an element with no screen-readable text. Add 'aria-label' or 'aria-labelled-by' attribute, or put some text in the element.`

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -10,6 +10,7 @@ var webpack = require('gulp-webpack');
 var plumber = require('gulp-plumber');
 
 require('node-jsx').install();
+require('react-a11y')();
 
 var IndexFileStream = require('./lib/gulp-index-file-stream');
 var webpackConfig = require('./webpack.config');
--- a/lib/footer.jsx
+++ b/lib/footer.jsx
@@ -50,7 +50,7 @@ var HiveCities = React.createClass({
           {cities.map(function(city, i) {
             return (
               <li key={i}>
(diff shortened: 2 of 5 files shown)


toolness:
(Note that the Travis build is failing because the `react-a11y` library is properly raising warnings about accessibility. Hooray, it's [working](https://travis-ci.org/mozilla/teach.webmaker.org/builds/53053665)!)

For dealing w/ the accessibility of the hamburger menu, we might want to look at 
[what Bootstrap does](http://getbootstrap.com/components/#navbar-default).

toolness:
Hmm, with the changes introduced by #132, it might be easier to just redo this PR from scratch, not sure.

toolness:
This is superseded by #137.

PR #129, opened 2015-03-04
Fix #110 - Add gulp webserver in watch task

### LINKED ISSUE #110: 'gulp watch' might as well start a server up
While we do currently support loading of the site via a `file:` URL, a bunch of stuff (like `pushState`) doesn't work on it, so we might as well have `gulp watch` start a dev server.

Not really sure which dev server is best here... There's [webpack-dev-server](http://webpack.github.io/docs/webpack-dev-server.html), which apparently gives us some webpack bonuses like auto-reload via socket.io, but I'm not sure if that will get triggered whenever we do something that's *not* webpack related.

Also worth looking into whatever [mofo-example-app](https://github.com/MozillaFoundation/mofo-example-app) uses.

@alicoding, are you interested in looking into this?

alicoding:
closes #110

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,4 +1,5 @@
 var path = require('path');
+var webserver = require('gulp-webserver');
 var _ = require('underscore');
 var gulp = require('gulp');
 var gutil = require('gulp-util');
@@ -18,7 +19,8 @@ var BUILD_TASKS = [
   'copy-dirs',
   'less',
   'webpack',
-  'generate-index-files'
+  'generate-index-files',
+  'server'
 ];
 
 var COPY_DIRS = [
@@ -127,6 +129,16 @@ gulp.task('watch', _.without(BUILD_TASKS, 'webpack'), function(cb) {
 
   gulp.watch(COPY_DIRS, ['copy-dirs']);
   gulp.watch(LESS_FILES, ['less']);
+});
+
+gulp.task('server', function () {
+  return gulp.src('dist')
+    .pipe(webserver({
+      livereload: {
+        enable: true
+      },
+      port: 8008
+    }));
 });
 
 gulp.task('s3', BUILD_TASKS, function() {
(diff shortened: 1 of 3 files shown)


alicoding:
@toolness R?

gulpfile.js
```diff
@@ -18,7 +19,8 @@ var BUILD_TASKS = [
   'copy-dirs',
   'less',
   'webpack',
-  'generate-index-files'
+  'generate-index-files',
+  'server'
```
toolness: oh, I don't think this is something we want in `BUILD_TASKS`... that array is only meant to hold all the tasks necessary to build the static site... I should probably add a comment or rename the constant to make that more obvious, sorry!

gulpfile.js
```diff
@@ -129,6 +131,16 @@ gulp.task('watch', _.without(BUILD_TASKS, 'webpack'), function(cb) {
   gulp.watch(LESS_FILES, ['less']);
 });
 
+gulp.task('server', function () {
```
toolness: Hmm would it be better to just add this to the `watch` task? That way one could just run `gulp watch` in one Terminal window rather than having to run `gulp watch` in one, and then `gulp server` in another? Or is it possible to do `gulp watch server` and have both run simultaneously?

We should also add information about how to run/access this server to the quick start in the README. I can do that in a separate commit if you want, or you can do it, whatever your preference is.

alicoding:
@toolness I updated a README file and would love to get you to look it before we merge this.

toolness:
Cool!

Does gulp have to be globally installed? @gesa had contributed #101 so it wasn't needed, and I like anything that reduces setup time for devs... Also see this stackoverflow answer: http://stackoverflow.com/a/22115441

alicoding:
@toolness done. Now we don't have to install gulp globally.

toolness:
Hooray!!!! Thanks Ali!!!

gesa:
Thanks guys, this is great!

PR #132, opened 2015-03-04
Migrate to react-router

toolness:
This gets rid of a bunch of my hacky `pushState` code, the weird `pages.PAGES` object, and the `<Ia>` component in favor of [`react-router`](https://github.com/rackt/react-router).

There are a few downsides to this approach:

* Currently, the generated static site uses relative URLs, which enables the site to be mounted anywhere in a directory structure. `react-router` really wants our site to be mounted at the root of a directory structure, so we'll likely lose this flexibility (or we'll need to implement some funky work-arounds, like hard-coding a directory prefix at build time).

* The current static site supports being accessed over the `file:` protocol but the new one won't.

* `react-router` doesn't seem to expose any APIs that allow us to traverse a site's URL structure. Right now `pages.URLS`, which is a list of all the URLs in the site, is hard-coded; I'll likely be able to iterate through the React structure of a `<Route>` element to generate a list of URLs but it will be hacky and might break with new versions of React or `react-router`.

I think that's about it, though. In exchange for this, though, we throw out a lot of my hacky/undocumented code, get hash-based navigation for free, and other [documented benefits](https://github.com/rackt/react-router#benefits-of-this-approach). It also means we won't have to implement a bunch of stuff in the future as the site grows, since `react-router` likely already supports our needs.

TODOs:

- [x] We're still setting the `<base href>` in generated static pages. Since we're no longer supporting being mounted at any URL, we can get rid of this, as it just complicates things. This also means providing absolute URLs to static assets like images and such.


toolness:
Ok, there's a number of things I'm not a huge fan of with the state of this code right now, but I think it's better to file bugs for them and merge this PR so we can unblock others from landing.

The main issue that will need to be filed is having `routes.URLS` be auto-populated, rather than manually kept in-sync with the routes React element.

alicoding:
This is pretty good @toolness ! r+ 

I think this might solve some of my problem that I'm facing right now regarding sourcemap (not entirely sure if it is related, but having a strange feeling that it will solve my problem).

davidascher:
BTW: I wouldn't be surprised if the react-router folks would be happy with
a PR to expose the routing tables.  Could help w/ @pomax's
https://github.com/rackt/react-router/issues/842

Pomax:
yeah, this would be superhandy as part of react-router for everyone else!

PR #137, opened 2015-03-04
Integrate react-a11y, but don't test against it yet

toolness:
Ok, this is another attempt at #128, with a few changes:

* It's based on the codebase post-#132.
* It doesn't *force* `react-a11y` warnings on ... yet. Since fixing some of them is non-trivial, I just want to merge this PR ASAP to lay the groundwork for solid accessibility support. We can file individual bugs for the existing warnings (they're all the checkboxes in the description for #128), and once those are all resolved, then we can have travis/gulp always test against `react-a11y`.

So, for now, in order to enable the accessibility warnings, you have to put `ENABLE_REACT_A11Y` in your OS environment (it doesn't matter what it's set to, it just has to be in the environment). Then running `gulp test-react-warnings` (or `npm test`) should raise any relevant accessibility issues.

This PR supersedes #128.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,6 +12,9 @@ var plumber = require('gulp-plumber');
 
 require('node-jsx').install();
 
+if ('ENABLE_REACT_A11Y' in process.env)
+  require('react-a11y')();
+
 var IndexFileStream = require('./lib/gulp-index-file-stream');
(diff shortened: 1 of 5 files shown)


toolness:
@alicoding, can you take a quick look at this?

alicoding:
r+

Issue #135, opened 2015-03-04
Disable minification of JS during development


toolness:
Our [current `webpack.config.js`](https://github.com/mozilla/teach.webmaker.org/blob/603f1f378707b060d01f287bb4124ccec27813fb/webpack.config.js#L19-L23) always minifies the JS it outputs.

Obviously this is a particularly unfortunate thing until #102 is fixed, but even once we have source maps, it can still be useful to not minify our development code, simply because minification takes extra time and can really slow down older systems. Also, we may eventually be testing and debugging our code on older browsers that don't have source map support!

@mmmavis, @gesa, @alicoding - Do you have any opinions on this?

alicoding:
I think we should always disable minification of JS when we run `gulp dev` or `npm start`.

toolness:
Good plan! I notice you've done this in #139, but since there are currently issues w/ that PR and ease of development is more important to us than production efficiencies, I've gone ahead and committed only your changes to the `webpack.config.js` file from your PR, so that we never minify at all for now. I've also filed #140 so that we remember to bring back minification for production deploys.

PR #139, opened 2015-03-04
Issue102

alicoding:
This patch should fix the issue we are having in #135 as well.

Right now by default when you run `gulp dev` or `npm start` it shouldn't uglify the JS for you since it's for development. The downside here as we know it will slow down the browser load time, but obviously it will increase our productivity significantly.

We still have the useful webpack plugin enabled when you run `gulp` only which should be the build step for us on production.

I'm still having trouble getting `LESS` to work on Firefox for some reason but other than that this is in a good state at the moment.


alicoding:
@toolness R?

toolness:
Awesome thanks Ali!

Hmm, I am seeing the sourcemaps for the CSS/LESS, but not for the JS... When I intentionally add an exception in the code, for instance, it just shows me `bundle.js` rather than the original file it came from in Chrome:

screen shot 2015-03-04 at 6 04 55 pm

Loading `bundle.js` in my text editor, I'm also not seeing any mention of `sourceMappingURL` like I see in `styles.css`...

toolness:
Oh, so I'm now noticing that when I run `gulp`, a source map for `bundle.js` *is* being generated, but `bundle.js` isn't being uglified.

And when I run `npm start` (aka `gulp dev`), no source map for `bundle.js` is generated, nor is `bundle.js` uglified.

So I guess things need to change so that the bundle **is** uglified for production, and source maps **are** generated for development?

This is getting a little confusing--I'm starting to think we need a test suite for our own build system! :sweat_smile:

gulpfile.js
```diff
@@ -30,9 +29,39 @@ var COPY_DIRS = [
...
 var LESS_FILES = './less/**/*.less';
 
+function webpackTask(options) {
+  console.log(options)
+  options = options || {};
+  var srcFile = './lib/index-static.jsx';
+  var outputPath = path.join(__dirname, COMPILED_DIR, 'js');
+  fs.removeSync(outputPath);
+  var outputName = 'bundle.js';
+  var config = require('./webpack.config');
```
toolness: hmm, since we are essentially dynamically creating our webpack config, perhaps we should just do away with `webpack.config.js` entirely?

gulpfile.js
```diff
@@ -52,96 +89,38 @@ gulp.task('copy-dirs', function() {
...
-    .pipe(gulp.dest('./dist'));
-});
-
-gulp.task('test', BUILD_TASKS.concat([
-  'test-react-warnings'
-]), function() {
-  gutil.log(gutil.colors.green.bold("Yay, tests pass!"));
-});
+// gulp.task('webpack', webpackTask({sourcemaps: true}));
+// gulp.task('webpack-optimized', webpackTask({optimize: true}));
```
toolness: should these comments be removed?

gulpfile.js
```diff
@@ -52,96 +89,38 @@ gulp.task('copy-dirs', function() {
...
     .pipe(gulp.dest('./dist'));
 });
 
-gulp.task('webpack', function() {
-  return gulp.src(webpackConfig.entry)
-    .pipe(webpack(webpackConfig))
-    .pipe(gulp.dest('./dist'));
-});
-
-gulp.task('test', BUILD_TASKS.concat([
```
toolness: er, why is this test suite going away?

gulpfile.js
alicoding: ideally I want this back, but not sure why gulp is firing these tasks even though I didn't have it anywhere...

toolness:
So this PR is doing lots of stuff that I'm confused about... I think it might be better to split this into two different objectives:

1. Adding JS and CSS source maps to *all* builds, regardless of whether they're dev or prod. Since the source maps can be in a separate file that only devtools access, they shouldn't impact the load times of production builds, and they also encourage anyone to delve into our site and see how it works.
2. Uglifying JS and CSS output for production builds *only*.

Is it ok if I take the very basics of what you've done to address (1) and issue a really simple separate PR to address them? Then we can address (2) anytime before launch.

alicoding:
Closing this in favour of #147

Issue #112, opened 2015-03-04
Implement Automated Testing


simonwex:
* [x] Implement testing framework ( #143)
* [X] Pick two open bugs
  *  Bug: #138 (tests added in 54b8ce0da26b1d8b281004988ae7e9e61d1a9750)
  *  Bug: #166 (tests added in #167)
* [x] implement tests that validate fixes for two open bugs
* [x] Document how to add a new test
* [x] Set up project on TravisCI
* [x] Add Travis status Image to README
* [x] Add mofo-style linter (#142)
* [x] Ensure all tests are passing

https://github.com/MozillaFoundation/plan/issues/351

toolness:
So one big advantage of having a static site generator is that the build phase kind of doubles as a testing phase.

I've just added some functionality for #117 that raises errors during `npm test` if any React warnings are raised. This includes:

* warnings that come with React core, such as forgetting to put a `key` attribute on components in an array;
* warnings raised by [prop validation](http://facebook.github.io/react/docs/reusable-components.html#prop-validation);
* anything logged in our own code via `console.warn()` during component initialization or rendering, like f28ef4624c1df17b946f0d941cb7352a2a8a2d24.

So I highly recommend we use these sorts of features to make sure our code is up-to-snuff at build/test time!

toolness:
That said, dynamic functionality, such as event handlers, isn't exercised at all during the build step, so writing tests for those would be super useful. <code>[Reacts.addons.TestUtils](http://facebook.github.io/react/docs/test-utils.html)</code> could come in handy here, and this blog post might too:

http://www.jackcallister.com/2014/12/01/building-a-test-suite-in-react-js.html

Currently our code doesn't actually include much dynamic event-handling yet. Two things that would be great to test are the collapsible hamburger code and the `pushState` support.

If we can't effectively test these things in an automated way, it might be worth writing up instructions for manually testing them.

toolness:
It seems like there exist modules like `react-a11y` (see #126) that we can add to our build-time testing, which effectively serve as "linters" for our React code. Cool!

toolness:
Woot, this is done!

PR #143, opened 2015-03-05
Add mocha tests for server/site-generator and browser.

toolness:
This attempts to resolve a lot of #112 by adding simple [mocha](http://mochajs.org/)-based test frameworks for:

* **The static site generator.** This is our version of the "server", the program that runs on a build machine to generate the static website. These tests are located in the `test` directory.
* **The browser.** This is the code that runs in user's browsers to progressively enhance their experience. These tests are located in the `test/browser` directory.

Currently there's only trivial test suites for [`lib/config.js`](https://github.com/mozilla/teach.webmaker.org/blob/master/lib/config.js), which behaves differently in the two environments.

In both environments, [`should`](https://www.npmjs.com/package/should) is used for test assertions. This assertion library only works on IE9+, but that should be OK for us; I'm assuming at the moment that we'll essentially not progressively enhance anything below IE9, thus leaving nothing to test on the browser.

This PR also introduces Webpack's [code splitting](http://webpack.github.io/docs/code-splitting.html) functionality to make builds faster and ensure that the browser-side tests use the exact same code as the app.

Tasks left to do:
- [X] Formally decide what browsers to support on the client-side and document it. If we *are* supporting client-side JS on browsers below IE9, we should switch from `should` to something else like [`expect`](https://github.com/Automattic/expect.js). **This decision has been moved to #151.**
- [x] Document the testing system and process of adding new tests to `README.md`.
- [x] Use [`mocha-phantomjs`](https://github.com/metaskills/mocha-phantomjs) to run the browser tests from the command-line, and as part of the Travis CI build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,21 @@
 [![Build Status](https://travis-ci.org/mozilla/teach.webmaker.org.svg)](https://travis-ci.org/mozilla/teach.webmaker.org)
 
 This is an initial attempt at implementing the Webmaker Learning
-front-page.
+website.
+
+# Overview
+
+This software consists of two major parts:
+
+* A **static site generator** that creates a number of
+  `index.html` files in various directories which can be viewed
+  in any browser, including ones that don't support JavaScript.
+* Client-side JavaScript code that **progressively enhances**
+  the user experience based on browser capabilities.
+
+It should be noted that, based on the
+[product roadmap][roadmap], the static site generator
+may eventually evolve into becoming a dynamic server.
 
 # Get started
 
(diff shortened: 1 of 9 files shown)


toolness:
I should note that I did take a brief look at using [jest](https://facebook.github.io/jest/) to do testing, but while its approach to [automatic mocking](https://facebook.github.io/jest/docs/automatic-mocking.html) does seem really useful, it only supports the command-line at the moment, and it tries to force you into a simulated browser environment via jsdom.

This seems like the *worst* of both worlds since there's no way to functionally test that your code works on all the browsers you want to support; nor is there any way to easily test code in a non-simulated-browser environment.

it seems like other people ran into the [same problem](https://github.com/facebook/jest/issues/139) and a lot of them switched to mocha. Since I already like mocha a lot, I figured it'd be a good solution to run with.

toolness:
Hmm, as for formally deciding whether to support browsers below IE9, that's out of my control and I don't want to block on it... I'd rather file a bug for it and if we need to support < IE9, it won't be hard to convert the assertions. So I've filed #151 for that.

@alicoding, can you review this PR?

alicoding:
r+ and my thought on IE9 might be better to wait until we launch the site and look at the statistics of our user's useragent to find out if it is even worth it to support IE9 at all. Lets land this first anyway.

Issue #145, opened 2015-03-05
'gulp dev' can start the web server before the 'dist' directory exists


toolness:
Today @adamlofting and I ran into the same race condition: because `gulp dev` starts `gulp watch` and `gulp server` simultaneously, `gulp server` might start itself before the `dist` directory is created by `gulp watch`. This means the user won't be able to see anything in it once it *is* created.

Since there isn't a compelling use case for running `gulp watch` independently from `gulp server`, I think the easiest way to fix this bug is to simplify the code by just having `gulp watch` start the server too. (It already waits for an initial build before doing anything, so the `dist` directory will definitely exist by the time the webserver starts.)

alicoding:
This is fixed so I'm closing the ticket.

PR #146, opened 2015-03-05
Coalesce the 'watch' and 'server' tasks, fixing #145.

toolness:
If a compelling use case for running `watch` without `server` (or vice versa) arises, we can always bring back the functionality too, so long as we don't regress back to #145 when we do.

@alicoding can you review this?


alicoding:
r+

PR #147, opened 2015-03-05
Add JS and CSS sourcemaps.

toolness:
As mentioned in https://github.com/mozilla/teach.webmaker.org/pull/139#issuecomment-77400812, this adds JS and CSS source maps to *all builds*, regardless of whether they're dev or prod.

Since the source maps are provided in separate `.map.` files that only dev tools access, they shouldn't impact the load times of production builds, and they also encourage anyone to delve into our site and see how it works!

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -6,6 +6,7 @@ var index = require('./lib/index-static.jsx');
 
 module.exports = {
   entry: './lib/main.jsx',
+  devtool: 'source-map',
   output: {
     path: __dirname + '/dist',
     filename: index.JS_FILENAME
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,6 +9,7 @@ var less = require('gulp-less');
 var prettify = require('gulp-prettify');
 var webpack = require('gulp-webpack');
 var plumber = require('gulp-plumber');
+var sourcemaps = require('gulp-sourcemaps');
 
 require('node-jsx').install();
 
@@ -54,9 +55,11 @@ gulp.task('copy-dirs', function() {
 gulp.task('less', function() {
   return gulp.src(LESS_FILES)
     .pipe(handleError())
+    .pipe(sourcemaps.init())
     .pipe(less({
       paths: [path.join(__dirname, 'less')]
     }))
+    .pipe(sourcemaps.write('./'))
     .pipe(gulp.dest('./dist'));
 });
 
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "gulp-plumber": "^0.6.6",
     "gulp-prettify": "^0.3.0",
     "gulp-s3": "^0.3.0",
+    "gulp-sourcemaps": "^1.5.0",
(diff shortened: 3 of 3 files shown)


toolness:
@alicoding does this look OK to you? Aside from production-related tasks like minification, is there anything #139 does that this doesn't do?

webpack.config.js
```diff
@@ -6,6 +6,7 @@ var index = require('./lib/index-static.jsx');
 
 module.exports = {
   entry: './lib/main.jsx',
+  devtool: 'source-map',
```
alicoding: this doesn't work for Firefox for some reason and I have to set it to `eval` to get it to work.

toolness: Hmm, what do you mean by "work for Firefox"? In general I've found that Firefox has very poor support for source maps--they work fine in the debugger, but not the console. 

For instance, here's what a `debugger` statement does in Firefox, with the code in this PR applied:



As you can see there it's correctly mapping to the original JSX source.

However, see that `console.log("HI")` right before the `debugger` statement? This is what that looks like in the console:



This is particularly frustrating for me because I spend much more of my time in the console than I do in the debugger. But I've never found a project that works any differently on Firefox, so when I really need source-mapped `console.log()` statements, I just switch to Chrome. It's kind of Chrome devtools' killer feature for me. :(

alicoding: Exactly the problem I was facing and had to set to `eval` to get this working both on debugger and console, but lets land this one for now and document that in order to get the console to work they either have to set `eval` manually for Firefox or use the debugger.

alicoding:
r+ with updating the README regarding Firefox users how can they debug here.

webpack.config.js
toolness: Whoa, yeah, you're totally right, setting `devtool` to `eval` *does* make console logging work on Firefox!!!!



WOW.

Unfortunately, it's totally dumb that this won't work on production, since it uses `eval()` to evaluate the code and thus won't work w/ minified stuff. It really sucks that Firefox's support for `.map.` files is limited to the debugger only.

Another thing I've just noticed about the `eval` approach is that it's actually only showing us the JSX file *after* the [JSX transform](http://facebook.github.io/react/docs/jsx-in-depth.html#the-transform). Which for Firefox console logging is still way better than referring to `bundle.js`, but for all other use cases (e.g. Chrome, the FF debugger) it's *worse*, since those other use cases show us the original JSX source when we use `.map.` files. So there is no clear winner here, which is lame.

toolness:
Good idea, I've added an optional `WEBPACK_DEVTOOL` env var in 223d71cd0b37a0c540a62d0574bfe391277a41ef and documented it in the readme, so folks can change the setting if they want.

webpack.config.js
toolness: Hmm, I just noticed that the exception tracebacks shown in mocha browser tests are also un-sourcemapped when using `source-map`, **even in Chrome**, but they work fine with `eval`.

PR #149, opened 2015-03-05
Fix #138 - Accessibility yay

### LINKED ISSUE #138: Fix existing react-a11y warnings
As of the merging of #137, this is what we need to do:

- [x] Fix all occurrences of `You have a click handler on a non-interactive element but no 'role' DOM property. It will be unclear what this element is supposed to do to a screen-reader user. http://www.w3.org/TR/wai-aria/roles#role_definitions`
- [x] Fix all occurrences of `You have a click handler on a non-interactive element but no 'tabIndex' DOM property. The element will not be navigable or interactive by keyboard users. http://www.w3.org/TR/wai-aria-practices/#focus_tabindex`
- [x] Fix all occurrences of `You have a click handler on an element with no screen-readable text. Add 'aria-label' or 'aria-labelled-by' attribute, or put some text in the element.`
- [x] Once the above are done, get rid of the `ENABLE_REACT_A11Y` environment variable check in `gulpfile.js` so that **accessibility warnings always cause the test suite to fail**.

alicoding:
Closes #138

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -13,8 +13,7 @@ var sourcemaps = require('gulp-sourcemaps');
 
 require('node-jsx').install();
 
-if ('ENABLE_REACT_A11Y' in process.env)
-  require('react-a11y')();
+require('react-a11y')();
 
 var IndexFileStream = require('./lib/gulp-index-file-stream');
 var webpackConfig = require('./webpack.config');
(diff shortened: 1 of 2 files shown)


alicoding:
@toolness R?

toolness:
Woot thanks!

PR #150, opened 2015-03-05
Fix #136 - dynamic route

### LINKED ISSUE #136: routes.URLS should be dynamically generated
As a result of #132, we've migrated to `react-router` but it doesn't provide a great way of iterating through all routes. Currently `lib/routes.js` exports an array called `URLS` that is essentially hard-coded; it'd be nice if we dynamically generated it from traversing the `<Routes>` React element in that module.

This is something we should have tests for, though, as it's likely to break when we upgrade to new versions of `react-router` and React. Another alternative is to actually add some route iteration functionality to `react-router` itself, and push it upstream.

alicoding:
Closes #136

### DIFF
--- a/lib/homepage.jsx
+++ b/lib/homepage.jsx
@@ -77,7 +77,7 @@ var Homepage = React.createClass({
       <div>
         <HeroUnit image="/img/hero-unit.jpg">
           <h1>Unlock opportunities for all citizens of the Web.</h1>
-          <div><Link to="foo/" className="btn btn-awsm">Join Us</Link></div>
+          <div><Link to="/foo/" className="btn btn-awsm">Join Us</Link></div>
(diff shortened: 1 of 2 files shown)


alicoding:
@toolness R?

toolness:
Niice!

I'm noticing that the code you wrote to dynamically generate `URLS` doesn't seem to support [nested routes](https://github.com/rackt/react-router/blob/master/docs/guides/overview.md#more-nesting)... which is fine, I think, because we don't currently have any nested routes. But do you think you can file a bug so it's documented and we address it in the future? Maybe also add a comment above the definition of `routes` reminding coders that nesting won't be reflected in `URLS` currently?

Other than that this looks great! Feel free to merge it.

PR #153, opened 2015-03-05
Add sidebar browser tests.

toolness:
This will test the meager client-side JS we've got and move us towards fixing #112!

@alicoding what do you think?! This uses [`React.addons.TestUtils`](http://facebook.github.io/react/docs/test-utils.html).

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
   },
   "devDependencies": {},
   "scripts": {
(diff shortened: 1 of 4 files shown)


alicoding:
test failing here @toolness :D

toolness:
@alicoding sorry, it should work now!

alicoding:
r+ looking good!

PR #159, opened 2015-03-06
Fix #125 - Build sitemap.xml

### LINKED ISSUE #125: Build sitemap.xml
A  sitemap is being designed in #124. It would be cool if the site build process generated a [sitemap XML](http://www.sitemaps.org/) so search engines could provide extra helpful results.

For example, I believe it's due to mozilla.org's own sitemap that these rich search results are provided by Google:

Aside from that, the code could potentially help new developers acquaint themselves with the project, maybe.

alicoding:
Closes #125

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -11,6 +11,7 @@ var webpack = require('gulp-webpack');
 var plumber = require('gulp-plumber');
 var merge = require('merge-stream');
 var sourcemaps = require('gulp-sourcemaps');
+var sitemap = require('gulp-sitemap');
 
 require('node-jsx').install();
 
@@ -24,6 +25,7 @@ var BUILD_TASKS = [
   'copy-dirs',
   'less',
   'webpack',
+  'sitemap',
   'generate-index-files'
 ];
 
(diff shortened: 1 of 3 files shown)


alicoding:
@toolness R?

gulpfile.js
```diff
@@ -47,6 +49,14 @@ function handleError() {
   });
 }
 
+gulp.task('sitemap', function() {
+  gulp.src('dist/**/*.html')
+    .pipe(sitemap({
+      siteUrl: process.env.teachURL || 'https://teach.webmaker.org'
```
toolness: do you mind changing this env var to be called `ORIGIN`? That way if we ever add persona etc it can just use that URL... and I'd also like to refer to all env vars in uppercase w/ underscores as a general matter of style.

Can you also document the env var in `README.md`?

package.json
```diff
@@ -10,6 +10,7 @@
     "gulp-plumber": "^0.6.6",
     "gulp-prettify": "^0.3.0",
     "gulp-s3": "^0.3.0",
+    "gulp-sitemap": "^2.0.1",
```
toolness: heh that's cool, didn't know there was a plugin that automates this for us!!

neat.

toolness:
Ultra dudical, aside from the comment about the environment variable I think this looks great.

toolness:
Hmm, I guess it should be noted that your PR doesn't seem to add anything to our `watch` task, which means that the sitemap won't be regenerated on-the-fly during development.

Do you think it should be? I'm on a laptop right now without A/C power so I have to admit that I am thinking about how much battery life all our watch tasks might suck out, lol. If we decide not to regenerate it on-the-fly, though, I guess we should document it...

PR #160, opened 2015-03-06
Sitemap watch

alicoding:
@toolness R?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,7 +25,6 @@ var BUILD_TASKS = [
   'copy-dirs',
   'less',
   'webpack',
-  'generate-index-files',
   'sitemap'
 ];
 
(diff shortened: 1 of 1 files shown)


toolness:
Oh, so this is making me realize something... there could be a doggone race condition here, because your `sitemap` task depends on `generate-index-files` running before it--which I assume is why you reorganized the `BUILD_TASKS` array in this PR. But just listing it after `generate-index-files` in that array won't actually do anything useful, because gulp executes all those tasks *in parallel*. So I think that you might have to specify `generate-index-files` as a prerequisite for `sitemap`, e.g. change the task definition to something like this:

```js
gulp.task('sitemap', ['generate-index-files'], function() {
  /* ... */
});
```

Does that make sense?

Aside from that it might make sense to actually replace all occurrences of `generate-index-files` with `sitemap`, since `sitemap` includes it as a dependency... Then you could just remove the `gulp.watch('dist/**/*.html', ['sitemap']);` entirely. Does that make sense?

Oy, I really wish that gulp managed dependency trees better. In a [blog post from April 2014](https://medium.com/@contrahacks/gulp-3828e8126466) one of gulp's creators claims that orchestrator will be going over a major overhaul to make dependency management easier, but I suspect that hasn't landed yet...

alicoding:
@toolness good catch! I have updated as you suggested and tested it working expected!

toolness:
Hooray! Looks good :+1:

PR #167, opened 2015-03-06
Get rid of baseURL in IndexFileStream, add tests. fixes #166

### LINKED ISSUE #166: Get rid of IndexFileStream's baseURL code
We're not using base URLs in the index files anymore as of aff1f5c79d319c23a1298847140f234d3a1df167, so we should get rid of the relevant code in `IndexFileStream`. I think it's also our only use of `path-posix` so we can probably get rid of that too.

AND WE'RE GONNA NEED A TEST!

toolness:

### DIFF
--- a/lib/gulp-index-file-stream.js
+++ b/lib/gulp-index-file-stream.js
@@ -1,5 +1,5 @@
 var Readable = require('stream').Readable;
-var posixPath = require('path-posix');
+var path = require('path');
 var util = require('util');
 var File = require('vinyl');
 
@@ -13,7 +13,7 @@ function IndexFileStream(indexStatic) {
   });
   this._baseDir = __dirname;
   this._indexStatic = indexStatic;
-  this._urls = indexStatic.URLS.slice();
+  this._urls = indexStatic.URLS.slice().reverse();
 }
 
 util.inherits(IndexFileStream, Readable);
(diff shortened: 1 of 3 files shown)


lib/gulp-index-file-stream.js
```diff
@@ -1,5 +1,5 @@
 var Readable = require('stream').Readable;
-var posixPath = require('path-posix');
```
alicoding: can you remove `path-posix` from package.json too?

toolness: Ah good catch! Done.

lib/gulp-index-file-stream.js
```diff
@@ -22,14 +22,13 @@ IndexFileStream.prototype._read = function() {
...
   var url = this._urls.pop();
-  var indexFile = posixPath.join(
+  var indexFile = path.join(
     this._baseDir,
-    posixPath.join.apply(posixPath, url.split('/').slice(1, -1)),
+    path.join.apply(path, url.split('/').slice(1, -1)),
     'index.html'
   );
 
   this._indexStatic.generate(url, {
```
alicoding: I don't know if we need second parameter here anymore? What do you think @toolness ?

alicoding:
Left one question regarding the removal of `options` object parameter. If you wish to fix then good to land otherwise also good to land. Your call :)

lib/gulp-index-file-stream.js
toolness: Yeah, I'm just suspecting we'll actually need options at some point in the future and didn't want to deal w/ removing it for now :)

PR #170, opened 2015-03-07
Use only named links. Fixes #161.

### LINKED ISSUE #161: Make 'gulp smoketest' raise React warnings when <Link> tags don't go to known links
So back when we had the `<Ia>` component, we had some [code](https://github.com/mozilla/teach.webmaker.org/blob/672941d06ed495e22a7af4f90b6053310d8764e9/lib/ia.jsx#L34-L36) that raised a React warning when the link didn't go to a known URL.

We since moved to `react-router` in #132, which replaced `<Ia>` with `<Link>`.

I thought that `<Link>` had the same functionality built-in, but it appears to only raise a warning when the dead link is *clicked*, rather than at instantiation or render time, which means that our test suite doesn't fail if the site contains dead links. We should change that!

toolness:
This fixes #161 by going back to only using route names, rather than URL paths, as the `to` property in `<Link>` elements, which ensures that URL validity is checked at build time rather than when a user clicks on a link.


test/browser/main.js
```diff
@@ -1,2 +1,5 @@
+process.env.NODE_ENV = 'test';
```
toolness: This makes it so `react-router` doesn't complain when we call `Router.run()` with a static location in the browser.

test/browser/routes.test.jsx
```diff
@@ -0,0 +1,29 @@
...
+var TestUtils = React.addons.TestUtils;
+var Router = require('react-router');
+var Link = Router.Link;
+
+var routes = require('../../lib/routes.jsx');
+
+describe('routes', function() {
+  routes.URLS.forEach(function(url) {
+    describe(url, function() {
+      it('should contain 0 non-named links', function(done) {
```
toolness: This test ensures we never regress back to using URL paths in the `to` property of `<Link>` components.

toolness:

alicoding:
r+

PR #178, opened 2015-03-09
Basic Google Analytics pageview tracking

adamlofting:
I want to talk more about this, but thought some reference code as a starting point would be useful.

I'll start a separate issue for the general discussion.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -7,3 +7,6 @@ var ENABLE_PUSHSTATE = (IN_STATIC_SITE &&
 exports.IN_STATIC_SITE = IN_STATIC_SITE;
 exports.GENERATING_STATIC_SITE = GENERATING_STATIC_SITE;
 exports.ENABLE_PUSHSTATE = ENABLE_PUSHSTATE;
+
+// this is a dummy account until we choose a domain name
+exports.GA_ACCOUNT='UA-59356678-3';
--- a/lib/googleanalytics.jsx
+++ b/lib/googleanalytics.jsx
@@ -1,0 +1,27 @@
+var React = require('react');
+var config = require('./config');
+
+var GoogleAnalyticsBasicTag = React.createClass({
+  render: function() {
+    if (config.GA_ACCOUNT) {
+      return (
+        <script dangerouslySetInnerHTML={{
+          __html: "(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){ \
+                  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o), \
+                  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m) \
+                  })(window,document,'script','//www.google-analytics.com/analytics.js','ga'); \
+                  \
+                  ga('create', '" + config.GA_ACCOUNT + "', 'auto');"
+        }}></script>
+      );
+    } else {
+      return (
+        <script dangerouslySetInnerHTML={{
+          __html: "// GA_ACCOUNT not set in config.js"
+        }}></script>
+      );
+    }
+  }
+});
+
+module.exports = GoogleAnalyticsBasicTag;
--- a/lib/index-static.jsx
+++ b/lib/index-static.jsx
@@ -1,5 +1,6 @@
 var React = require('react');
 
+var GoogleAnalyticsBasicTag = require('./googleanalytics.jsx');
 var routes = require('./routes.jsx');
 
 function generateWithPageHTML(url, options, pageHTML) {
@@ -26,6 +27,7 @@ function generateWithPageHTML(url, options, pageHTML) {
           __html: pageHTML
         }}></div>
         <script src='https://api.tiles.mapbox.com/mapbox.js/v2.1.5/mapbox.js'></script>
+        <GoogleAnalyticsBasicTag/>
         <script src="/commons.bundle.js"></script>
         <script src="/app.bundle.js"></script>
       </body>
--- a/lib/googleanalytics.js
+++ b/lib/googleanalytics.js
@@ -1,0 +1,13 @@
+module.exports = {
+  /**
+   * pageview:
+   * a proxy to basic GA pageview tracking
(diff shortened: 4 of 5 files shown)


adamlofting:
To note, this code errors in the console `ga not defined`, but does send the pageview events to GA. Which is weird in itself. 

screen shot 2015-03-09 at 13 53 52

lib/index-static.jsx
```diff
@@ -26,6 +27,7 @@ function generateWithPageHTML(url, options, pageHTML) {
         }}></div>
         <script src="/commons.bundle.js"></script>
         <script src="/app.bundle.js"></script>
+        <GoogleAnalyticsBasicTag/>
```
alicoding: move this two line up and your error should be fixed.

adamlofting:
Following discussion here: https://github.com/mozilla/teach.webmaker.org/issues/180 

I think this implements the basic level of GA pageview tracking we need right now.

adamlofting:
I've just noticed we're using `develop` as the branch rather than `master` so I'll rebase to that now.

adamlofting:
Just rebased this to develop now.
R? @toolness

lib/config.js
```diff
@@ -7,3 +7,6 @@ var ENABLE_PUSHSTATE = (IN_STATIC_SITE &&
 exports.IN_STATIC_SITE = IN_STATIC_SITE;
 exports.GENERATING_STATIC_SITE = GENERATING_STATIC_SITE;
 exports.ENABLE_PUSHSTATE = ENABLE_PUSHSTATE;
+
+// this is a dummy account until we choose a domain name
+exports.GA_ACCOUNT='UA-59356678-3';
```
toolness: So as of #194, we're actually storing configuration in environment variables and transferring the whitelisted ones over to the client-side via webpack's `DefinePlugin`... #194 might be enough to show you how to integrate it into this PR, but if not, that's fine, we can just land this and I can integrate the `DefinePlugin` stuff in a separate PR.

lib/googleanalytics.jsx
```diff
@@ -0,0 +1,27 @@
...
+                  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m) \
+                  })(window,document,'script','//www.google-analytics.com/analytics.js','ga'); \
+                  \
+                  ga('create', '" + config.GA_ACCOUNT + "', 'auto');"
+        }}></script>
+      );
+    } else {
+      return (
+        <script dangerouslySetInnerHTML={{
+          __html: "// GA_ACCOUNT not set in config.js"
```
toolness: Do you think it's useful to define a no-op `ga()` function instead of not defining it at all? That way the rest of our analytical code could just call `ga()` with reckless abandon rather than constantly checking to see if it's defined, but I leave it up to you!

lib/config.js
adamlofting: I'm dropping offline now, so if you have time to implement today that would be great. Otherwise I can pick up again tomorrow.

Either way, I'm happy to see environment variables in use :+1:

lib/googleanalytics.jsx
```diff
@@ -0,0 +1,27 @@
+var React = require('react');
+var config = require('./config');
+
+var GoogleAnalyticsBasicTag = React.createClass({
+  render: function() {
+    if (config.GA_ACCOUNT) {
+      return (
+        <script dangerouslySetInnerHTML={{
+          __html: "(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){ \
```
toolness: Oh, so by putting the GA code into an actual module, I meant actually taking the value of `__html` and turning it into its own module. For instance, something like this:

```javascript
exports.initialize = function() {
  if (!config.GA_ACCOUNT) return;

  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
})(window,document,'script','//www.google-analytics.com/analytics.js','ga');

  ga('create', config.GA_ACCOUNT, 'auto');
  ga('send', 'pageview');
};
```

Then the first line of `routes.jsx`'s `run()` function could just be `ga.initialize()`, and we wouldn't have to deal with putting a script tag with "dangerous inner HTML" in at all... But we could either land this and I could issue a PR, or we could just go with your version.

toolness:
I'm interested in knowing your thoughts on the points I brought up, but other than that, this looks good overall!

toolness:
Ok, I'm going to merge this now, and we can address any of the other things in future PRs!

lib/googleanalytics.jsx
adamlofting: Ahh, that's interesting. 

But do you know if this would this reload `www.google-analytics.com/analytics.js` on every page load?

lib/googleanalytics.jsx
adamlofting: Yeah, that's a good call.

But I'd like to do this in a way that keeps the no-op `ga()` 'near' the pageview / event tracking module. 

Moving forward, I'll modify or build a version of https://github.com/mozilla/webmaker-analytics that works better with this React setup.

lib/googleanalytics.jsx
toolness: Oh, sorry, I missed this somehow... Nope, it won't reload `analytics.js` on every page load if we do it at the beginning of `run()`... The `run()` function only gets called once per initial page load, while the callback passed to `Router.run()` is called once on initial page load, and then again every time the user navigates to a new page (if they're using hash-based or pushState navigation). I actually changed the code to work this way in bd0a8b3b15276d31615e57f0f4501a790790e168, apologies for not asking for code review from you first. If you think there's anything amiss there I will revert the commit immediately.

adamlofting: That's awesome! :+1: 
Absolutely no apologies, or reverts needed here. Thanks for working on that.

toolness: Hooray! Awesome :smile:

PR #179, opened 2015-03-09
Add manual test suite

toolness:
This just adds a page at `/test/manual/` that shows all the static pages of the site in iframes at different sizes, which should make it easier for us to verify that every that everything looks okay. A checkbox is provided to allow the tester to see the thumbnails with JS enabled or disabled (requires a browser with support for the [iframe sandbox attribute](http://caniuse.com/#feat=iframe-sandbox)).

Here's what it looks like right now:

At present all the iframes are loaded at once, but as the site grows we can change the manual test suite to only render iframes that are within the visible viewport.

### DIFF
--- a/test/browser/manual-main.jsx
+++ b/test/browser/manual-main.jsx
@@ -1,0 +1,110 @@
+var React = require('react');
+
+var routes = require('../../lib/routes.jsx');
+
+var RouteThumbnail = React.createClass({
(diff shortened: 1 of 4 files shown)


alicoding:
r+! This is great and I think we should do this kind of test more <3

toolness:
Woot, thanks for the feedback Ali :smile:

PR #182, opened 2015-03-09
Adding icons to navigation sidebar

### LINKED ISSUE #113: Sidebar menu items need icons
_From @toolness on February 26, 2015 0:18_

_Copied from original issue: toolness/mozilla-learning-html#1_

vazquez:
Fixes #113


less/styles.less
```diff
@@ -211,6 +211,26 @@ html.no-js .sidebar {
...
+    float: none;
+    width: 5.4rem;
+    opacity: 0.6;
+    margin: 0 1.5rem 0 0;
+    @media (min-width: @screen-md-min) {
+      float: left;
+    }
+  }
+
+  .sidebar-menu li:nth-child(3) img {
```
toolness: Hmm, I am a bit wary of `:nth-child` because changing the DOM structure just a little bit could throw off the styling in hard-to-debug ways. Do you think maybe we could add a `.sidebar-icon` class instead? Unless `:nth-child` is an accepted best practice and I'm just being paranoid, which is entirely possible. I guess I'd defer to @gesa for best judgement on this.

lib/sidebar.jsx
```diff
@@ -27,24 +27,28 @@ var TriangleCorner = React.createClass({
 var Sidebar = React.createClass({
   MENU_ENTRIES: [
     {
-      name: "Teaching Activities",
+      name: "Teaching Materials",
```
toolness: Hmm I thought @hannahkane and @iamjessklein  said that it was called "Teaching Activites" now? It used to be "Materials" but I changed it in a commit late last week...

toolness:
This looks splendid in my browser! Aside from the two issues I mentioned above, I think it is radical.

lib/sidebar.jsx
vazquez: Ahh! My bad, I got confused. I will switch it back.

less/styles.less
vazquez: I wanted to avoid using superflous classes, but what you say makes sense. I will defer to target with a class.

toolness:
Cool I dig it! Thanks.

PR #185, opened 2015-03-09
Add Mapbox component

alicoding:
This PR only adds MapComponent to the repo. I will file a ticket to work on getting it to display correctly in the next ticket or maybe after @mmmavis or @gesa finish with the CSS work on the page before I start working on the CSS for map itself.

### DIFF
--- a/lib/index-static.jsx
+++ b/lib/index-static.jsx
@@ -15,6 +15,8 @@ function generateWithPageHTML(url, options, pageHTML) {
         <link rel="stylesheet" href="http://fonts.googleapis.com/css?family=Open+Sans:400,300,300italic,400italic,600,700,600italic,700italic,800,800italic"/>
         <link rel="stylesheet" href="/vendor/bootstrap/css/bootstrap.min.css"/>
         <link rel="stylesheet" href={'/' + exports.CSS_FILENAME}/>
+        <script src='https://api.tiles.mapbox.com/mapbox.js/v2.1.5/mapbox.js'></script>
+        <link href="https://api.tiles.mapbox.com/mapbox.js/v2.1.5/mapbox.css" rel="stylesheet" />
         <script dangerouslySetInnerHTML={{
           __html: "document.documentElement.setAttribute('class', '');"
         }}></script>
--- a/lib/index-static.jsx
+++ b/lib/index-static.jsx
@@ -15,7 +15,6 @@ function generateWithPageHTML(url, options, pageHTML) {
         <link rel="stylesheet" href="http://fonts.googleapis.com/css?family=Open+Sans:400,300,300italic,400italic,600,700,600italic,700italic,800,800italic"/>
         <link rel="stylesheet" href="/vendor/bootstrap/css/bootstrap.min.css"/>
         <link rel="stylesheet" href={'/' + exports.CSS_FILENAME}/>
-        <script src='https://api.tiles.mapbox.com/mapbox.js/v2.1.5/mapbox.js'></script>
         <link href="https://api.tiles.mapbox.com/mapbox.js/v2.1.5/mapbox.css" rel="stylesheet" />
         <script dangerouslySetInnerHTML={{
           __html: "document.documentElement.setAttribute('class', '');"
@@ -26,6 +25,7 @@ function generateWithPageHTML(url, options, pageHTML) {
         <div id="page-holder" dangerouslySetInnerHTML={{
           __html: pageHTML
         }}></div>
+        <script src='https://api.tiles.mapbox.com/mapbox.js/v2.1.5/mapbox.js'></script>
         <script src="/commons.bundle.js"></script>
         <script src="/app.bundle.js"></script>
       </body>
--- a/lib/map.jsx
+++ b/lib/map.jsx
@@ -12,7 +12,6 @@ var MapComponent = React.createClass({
       this.map = L.mapbox.map('map', this.props.mapId, {accessToken: this.props.accessToken});
     },
     componentWillUnmount: function() {
-      delete L.mapbox.accessToken;
       this.map.remove();
     },
     // Called on initialization and after each change to the components
--- a/lib/map.jsx
+++ b/lib/map.jsx
@@ -8,14 +8,13 @@ var MapComponent = React.createClass({
     },
 
     componentDidMount: function () {
-      console.log('here1')
-      // Provide your access token
-      L.mapbox.accessToken = this.props.accessToken;
-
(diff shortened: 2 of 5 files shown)


alicoding:
@toolness can you review just the component code part first? I just want to at least get this part finish before start working on the CSS as mentioned above.

lib/index-static.jsx
```diff
@@ -15,6 +15,8 @@ function generateWithPageHTML(url, options, pageHTML) {
         <link rel="stylesheet" href="http://fonts.googleapis.com/css?family=Open+Sans:400,300,300italic,400italic,600,700,600italic,700italic,800,800italic"/>
         <link rel="stylesheet" href="/vendor/bootstrap/css/bootstrap.min.css"/>
         <link rel="stylesheet" href={'/' + exports.CSS_FILENAME}/>
+        <script src='https://api.tiles.mapbox.com/mapbox.js/v2.1.5/mapbox.js'></script>
```
toolness: Hmm, at the very least we should move the script tag to the end of the document for faster page loading... After we land this, though, we might want to file a ticket later to only load this on the Clubs page.

lib/map.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+var MapComponent = React.createClass({
+
+    propTypes: {
+      accessToken: React.PropTypes.string.isRequired,
+      mapId: React.PropTypes.string.isRequired,
+    },
+
+    componentDidMount: function () {
+      console.log('here1')
```
toolness: Remove this debug line?

lib/map.jsx
```diff
@@ -0,0 +1,28 @@
+var React = require('react');
+
+var MapComponent = React.createClass({
+
+    propTypes: {
+      accessToken: React.PropTypes.string.isRequired,
+      mapId: React.PropTypes.string.isRequired,
```
toolness: I think `mapboxId` might be a better name for this one, to disambiguate it from the actual `id` of the `<div>`?

lib/map.jsx
```diff
@@ -0,0 +1,28 @@
...
+    },
+
+    componentDidMount: function () {
+      console.log('here1')
+      // Provide your access token
+      L.mapbox.accessToken = this.props.accessToken;
+
+      // Create a map in the div #map
+      var map = L.mapbox.map('map', this.props.mapId);
```
toolness: We might want to assign the map to `this.map` so the component can use it later?

lib/map.jsx
```diff
@@ -0,0 +1,28 @@
...
+    },
+
```
toolness: We should probably add a [`componentWillUnmount`](http://facebook.github.io/react/docs/component-specs.html#unmounting-componentwillunmount) that calls [`remove()`](http://leafletjs.com/reference.html#map-remove) on the map here.

lib/map.jsx
```diff
@@ -0,0 +1,28 @@
...
+    // Called on initialization and after each change to the components
+    // props or state
+    render: function () {
+      return (
+        <div id="map"></div>
```
toolness: Hmm, so a downside of this is that we're using a global ID for the map here... While that's not altogether bad, since our current design only puts one map on the page, the assumption is generally that React components can be instantiated multiple times on the page. I recommend either changing the name of this component to `SingletonMapComponent`, to indicate that it's a singleton, or removing the `id` and instead passing an element reference (via `this.getDOMNode()`) into the [`L.mapbox.map`](https://www.mapbox.com/mapbox.js/api/v2.1.5/l-mapbox-map/) constructor.

lib/map.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+    propTypes: {
+      accessToken: React.PropTypes.string.isRequired,
+      mapId: React.PropTypes.string.isRequired,
+    },
+
+    componentDidMount: function () {
+      console.log('here1')
+      // Provide your access token
+      L.mapbox.accessToken = this.props.accessToken;
```
toolness: Rather than setting the global `accessToken`, I think it's better to pass it into the [`L.mapbox.map`](https://www.mapbox.com/mapbox.js/api/v2.1.5/l-mapbox-map/) constructor, so as to avoid messing with global state.

lib/map.jsx
toolness: (Actually, if we end up getting rid of the `id` attribute, I think it's fine to keep this name.)

alicoding:
@toolness any suggestion on how to deal with this on the test? https://travis-ci.org/mozilla/teach.webmaker.org/builds/53690110#L347

lib/map.jsx
```diff
@@ -0,0 +1,27 @@
...
+      accessToken: React.PropTypes.string.isRequired,
+      mapId: React.PropTypes.string.isRequired,
+    },
+
+    componentDidMount: function () {
+      // Create a map in the div #map
+      this.map = L.mapbox.map('map', this.props.mapId, {accessToken: this.props.accessToken});
+    },
+    componentWillUnmount: function() {
+      delete L.mapbox.accessToken;
```
toolness: Hmm, I don't think there's a need to delete `L.mapbox.accessToken` since we're not touching it anymore...

lib/map.jsx
```diff
@@ -0,0 +1,27 @@
+var React = require('react');
+
+var MapComponent = React.createClass({
```
toolness: Can we change `MapComponent` to `SingletonMapComponent` here too just so it's clear to folks who are reading this file?

toolness:
~~Hmmm good question... one option for the time being might be to just test to see if `typeof(L) == 'undefined'` on `componentDidMount` and, if so, exit immediately, with a comment explaining what kind of use case it addresses.~~

~~The other option is to actually add the MapBox `script` tag to the testing suite's `index.html` page, but I'm not sure how I feel about that because it means you have to be connected to the internet for the tests to work...~~

Oh, wait! It looks like `mapbox.js` is available on npm:

https://github.com/mapbox/mapbox.js/#usage-with-browserify

If we just add it to our `package.json` then we can simply have the map component `require('mapbox.js')` and we should be good to go for everything!

alicoding:
Ughh mapbox vs mapbox.js lol I looked it up and thought why don't they publish to npm haha. Ty Atul!

toolness:
Heh, yeah, I found `mapbox` first too and was confused. I actually found the official github repo by doing a search for "mapbox js test" to see if there was any advice on testing with mapbox, and the official github repo was the first hit.

toolness:
Nice, this looks good, thanks @alicoding !!

Issue #142, opened 2015-03-05
Integrate mofo-style


toolness:
I'm not sure if this is a duplicate or not, but we should integrate with [mofo-style](https://www.npmjs.com/package/mofo-style) and run it as part of our test suite.

toolness:
@gesa are you still interested in taking this one on?

gesa:
So someone in IRC mentioned that @alicoding might already be on this? Additionally, @gvn & @Pomax are currently applying mofo-style to all the repos. If this isn't on anyone's radar already, I'll **totally** take a stab at it—I just don't want to duplicate work.

Gentlemen, could you all confirm that this isn't already in progress somewhere?

Pomax:
This is a currently heartbeat over at https://github.com/MozillaFoundation/plan/issues/351 -- we're on the job.

Pomax:
closing this as a duplicate of https://github.com/mozilla/teach.webmaker.org/issues/112 that we're working on this heartbeat

toolness:
Er, my understanding was that I was on the hook for integrating mofo-style, and aki expressed interest in it so I delegated it to her. Hmmmm.

toolness:
Actually, #142 is quite a doozy and I think it's worth splitting it up into multiple mini-bugs. This is now the mini-bug for integrating mofo-style!

Pomax:
Hmm, I don't see your name in the App Matrix for plan/351, can you ask @wex what he wants to have happen here?

alicoding:
@Pomax you meant @simonwex :P

toolness:
Just talked to him and it looks like I am responsible... @gesa want to bust this up?

gesa:
Here we go!

simonwex:
@Pomax it's because @alicoding was going to do it, but @toolness is now. Updated spreadsheet.

toolness:
Thanks @gesa!

Do you think you could maybe split this up into two separate PRs/issues, the way I did with #137 and #138, where one PR adds *optional* support for style guide checking and the other issue is for fixing all the existing style issues w/ our code? (Or if it's more convenient, multiple issues/PRs could be filed that each fix the style issues in a subset of files.)

PR #187, opened 2015-03-09
 Adds mofo-style to package and gulpfile

gesa:
for #142: Integrate mofo-style

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,6 +12,9 @@ var plumber = require('gulp-plumber');
 var merge = require('merge-stream');
 var sourcemaps = require('gulp-sourcemaps');
 var sitemap = require('gulp-sitemap');
+var beautify = require('gulp-jsbeautify');
+var jscs = require('gulp-jscs');
+var jshint = require('gulp-jshint');
 
 require('node-jsx').install();
 
@@ -21,6 +24,7 @@ var IndexFileStream = require('./lib/gulp-index-file-stream');
 var webpackConfig = require('./webpack.config');
 
 var BUILD_TASKS = [
+  'jscs',
   'copy-test-dirs',
   'copy-dirs',
   'less',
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -30,9 +30,14 @@
     "vinyl": "^0.4.6",
     "webpack": "^1.6.0"
   },
-  "devDependencies": {},
+  "devDependencies": {
+    "gulp-jsbeautify": "^0.1.1",
+    "gulp-jscs": "^1.4.0",
```
alicoding: shouldn't we move this to the `dependencies` list instead now? since it is required in order to run the app.

gesa: Good catch!

gulpfile.js
```diff
@@ -125,6 +129,27 @@ gulp.task('generate-index-files', function() {
...
+});
+
+
+gulp.task('jscs', function () {
+  // jscs doesn't play nice with *.jsx files so we're avoiding lib/*.jsx
+  return gulp.src(['*.js', 'test/*.js', 'test/**/*.js'])
+      .pipe(jscs({ configPath: 'node_modules/mofo-style/linters/.jscsrc' }));
+});
+
+gulp.task('test', ['jscs', 'jshint', 'beautify']);
```
toolness: Do you mind changing this to, say, `lint-test`, just so it's not confused with `npm test`?

gesa: I actually added it to the end of `npm test`

but come to think of it they're in the wrong order.

gulpfile.js
```diff
@@ -125,6 +129,27 @@ gulp.task('generate-index-files', function() {
...
+
+gulp.task('jshint', function() {
+  return gulp.src('./lib/*.js')
+      .pipe(jshint({ lookup: 'node_modules/mofo-style/linters/.jshintrc' }))
+      .pipe(jshint.reporter('default'));
+});
+
+
+gulp.task('jscs', function () {
+  // jscs doesn't play nice with *.jsx files so we're avoiding lib/*.jsx
```
toolness: Ah that's a bummer--we should file that as an issue once this lands, so we don't forget!

package.json
```diff
@@ -30,9 +32,12 @@
...
     "webpack": "^1.6.0"
   },
-  "devDependencies": {},
+  "devDependencies": {
+    "gulp-jscs": "^1.4.0",
+    "gulp-jshint": "^1.9.2"
+  },
   "scripts": {
-    "test": "gulp smoketest && mocha test/*.test.js && npm run browsertest",
+    "test": "gulp smoketest && mocha test/*.test.js && npm run browsertest && gulp lint-test",
```
toolness: Do you think you could remove this for now, solely so that our tests aren't broken while you're fixing the errors raised by `lint-test`? Then once you've fixed them, we can re-add it.

toolness:
Aside from my one comment about removing `gulp lint-test` from `npm test` until we've got the existing lint bugs fixed, I think this looks great. If you do that and Travis is happy, I say merge!

PR #189, opened 2015-03-09
Tell user to re-run 'npm start' if crucial files change.

toolness:
After running "git pull" or changing files like `gulpfile.js`, `package.json`, and `webpack.config.js`, it might not occur to novice (or even expert) developers that they need to re-run `npm start`/`gulp watch`. This simple modification to `gulp watch` just watches those three files: if they ever change, it immediately tells the user to re-run `npm start` and exits.

(Even better would be to auto-restart the process ourselves, but I'm afraid of weird edge cases or bizarre infinite-restart loops and such.)


gesa:
clever.

toolness:
Hehe thanks!

PR #193, opened 2015-03-10
Fixes Issue#186 - Add DOM elements to the Clubs page

### LINKED ISSUE #234: Define <main> element
Currently we have `<footer>` on our pages, I wonder if we should define `<main>` as well?  Looks like https://github.com/mozilla/teach.webmaker.org/blob/develop/lib/page.jsx#L16 is the place to change if we decide to have `<main>`.

/cc @toolness @alicoding @gesa

mmmavis:
- Added DOM elements and layout to the Clubs page
- Added `fixme-page.jsx` so broken links have a place to go and don't fail our test

Fixes #234 
(edited by @gesa to make sure this PR automatically closes the issue when landed)


lib/homepage.jsx
```diff
@@ -38,11 +39,10 @@ var CaseStudies = React.createClass({
   render: function() {
     return (
       <div className="case-studies">
-        <blockquote className="primary-quote">
-          <img src="/img/maurya-nyc@2x.png" alt="Maurya NYC Quote" width="148" />
-          Web literacy is about more than coding - it is about how you can be a better web citizen.
-          <small>Maurya C. New York, United States</small>
-        </blockquote>
+        <Blockquote className="primary-quote"
```
toolness: Cool, I like the idea of componentizing this into a thing we can easily reuse across all our pages!

lib/blockquote.jsx
```diff
@@ -0,0 +1,15 @@
+var React = require('react');
+
+var Blockquote = React.createClass({
```
toolness: Since we're planning on reusing this across multiple pages, I think it would be useful to add some [prop validation](http://facebook.github.io/react/docs/reusable-components.html#prop-validation)... that helps self-document the code and will also make our smoke test raise errors if props don't validate at build time!

lib/clubs.jsx
```diff
@@ -0,0 +1,116 @@
...
+      <div>
+        <HeroUnit image="/img/hero-clubs.jpg">
+          <h1>Mozilla Learning Clubs</h1>
+          <div><Link to={ctaLink} className="btn btn-awsm">Add Your Club</Link></div>
+          <div><p>or <Link to="/fixme">find out more</Link> about us</p></div>
+        </HeroUnit>
+        <WebLitMap/>
+        <HowClubWorks/>
+        <Blockquote className="primary-quote"
+                    imgSrc="/img/mikko-finland@2x.png" imgAlt="Mikko Finland Quote"
```
toolness: So, this is a weird question, but ... are we actually using the non-`@2x` images yet? Do web browsers somehow handle this when they see the `@2x`, or do we need to file a bug to add some JS logic or CSS rules for that? I'm completely unfamiliar with how to adjust to this post-retina world we live in.

toolness:
Looking great so far!

lib/clubs.jsx
```diff
@@ -0,0 +1,116 @@
...
+      imgAlt: "icon tips",
+      head: "Helpful Tips",
+      subhead: "Tips for running your Club"
+    }
+  ],
+  render: function() {
+    return(
+      <div className="row">
+        <div className="col-sm-4 col-md-4 col-lg-4">
+          <IconLink info={this.ICON_LINKS[0]} />
```
alicoding: I think it might make sense to loop through `ICON_LINKS` instead of doing indexing like this.
Maybe `forEach` instead.

mmmavis: Oh yea, I think in this case we can loop it through since each of the icon link has the same width.  Cool, I'll change what's on the `homepage` as well.

lib/clubs.jsx
mmmavis: Oh yea, sorry was gonna raise this question as I'm not sure either.  I know for `background-image` we can simply use CSS to detect and use the right version of file.  Not sure if there's something for `` as well.  Maybe @gesa would know?

gesa: I wrote a [teensy-tiny script](https://github.com/mozilla/webmaker-landing-pages/blob/master/site/resources/js/core.js#L10-L30) for this. Best practice, images should [include a noscript](https://github.com/mozilla/webmaker-landing-pages/blob/5d37adef6ce8272839226e8e99c45036189cc9cf/site/src/from/net-neutrality.html#L39-L42)

toolness:
Erm, so with the landing of the directory reorganization in #238, this PR is likely broken :( Sorry about that... If it's a lot of trouble for you to merge with the latest `develop` branch, let me know and I can help!

less/styles.less
```diff
@@ -51,8 +62,43 @@ body {
   .justify-content(center);
 }
 
-.hero-unit h1 {
```
alicoding: We are going to apply the same h1 style to all pages?

alicoding:
Code looking good @mmmavis but there are few things which I think we'll need design reviews since it doesn't exactly match what's in the redpen, but I'm not going to fail your ticket on that since it's not entirely up to me I guess :P

So, I would just r+ and get you to rebase and squash once again then merge once travis give you green light :+1:

less/styles.less
mmmavis: I assume so but I could be wrong.  Personally hierarchy-wise I like having one single `<h1>` elements on a page and to me all the `<h1>` hero heading should share the same styling.  We don't have to worry about this here as @gesa will be refactoring CSS/LESS later.  (We can do a design review after).

alicoding:
Your test is failing @mmmavis

mmmavis:
@alicoding yea the unnamed links.  I guess I need to find a way to fix those `<Link>` that don't have a destination yet.  

A patch will be committed.

toolness:
@mmmavis hmm good question... you could either make them `<a href="">` instead of `<Link to="">` for the time being and they won't explode the tests... Either that or you could just link them to `home` for now.

mmmavis:
@toolness so `<Link to="">` and `<Link>` (without the `to` attr) both don't work.  Can I suggest creating a `fixme-page.jsx` so we can use it as `<Link to="/fixme">`?  I feel like that way it's easier for us to search for all "broken links" later.

toolness:
Ah good idea! Do you think you can add `fixme-page.jsx` in your PR?

mmmavis:
@toolness yup on it :) :+1:

mmmavis:
@alicoding @toolness @gesa 
Let's land this first and file mini tickets later for small bugs/tweaks.  Can someone review it one last time and merge this for me?  :hamster:

components/blockquote.jsx
```diff
@@ -0,0 +1,24 @@
...
+    quote: React.PropTypes.string.isRequired,
+    author: React.PropTypes.string.isRequired
+  },
+  render: function() {
+    return (
+      <blockquote className={this.props.className}>
+        <figure>
+          <img src={this.props.imgSrc} alt={this.props.imgAlt} width="148" />
+        </figure>
+        {this.props.quote}
```
toolness: May I recommend actually changing this to `{this.props.children}`? That way the quote can actually contain rich markup like italics, bold and such, and would also read nicely when used, e.g.:

```html
<Blockquote imgSrc="/blah.jpg" author="Mmmmmmmmavis">
  React is <em>cooooooooooooool</em>!
</Blockquote>
```

Anyways is no big deal, we can file a bug for it and land it later if you like the idea.

components/clubs.jsx
```diff
@@ -0,0 +1,164 @@
+var React = require('react');
```
toolness: Is this file meant to be part of the PR? It seems like a duplicate of `clubs-page.jsx`...

components/fixme-page.jsx
```diff
@@ -0,0 +1,13 @@
+var React = require('react');
+
+var PlaceholderPage = require('./placeholder-page.jsx');
+
+var FixmePage = React.createClass({
```
toolness: This is a great idea! In the future maybe we can add some kind of "pre-production-test" to ensure that we have no links pointing here before we push a new release. :)

components/clubs.jsx
mmmavis: Duh, this is back again?  I feel like I've deleted this file (along with other ones) a million times.  Anyways I'll do it again.

components/page-end-cta.jsx
```diff
@@ -0,0 +1,17 @@
+var React = require('react');
+var Router = require('react-router');
+var Link = Router.Link;
+
+var PageEndCTA = React.createClass({
+  render: function() {
+    return (
+      <div className="page-end-cta">
+        <p>{this.props.description}</p>
```
toolness: Similar to my [suggestion about `Blockquote`](https://github.com/mozilla/teach.webmaker.org/pull/193/files#r26265692) above, I think that using `this.props.children` instead of `this.props.description` here might be nice, especially if the need for markup in the description ever arises.

less/styles.less
```diff
@@ -36,6 +39,14 @@ body {
 
   .hero-unit {
     margin-right: -(@grid-gutter-width / 2);
+    .learn-more {
+      font-size: 14px;
```
toolness: Should we be using `px` or `rem` here?

less/styles.less
toolness: Yeah, I agree w/ having only one `<h1>` per page.

less/styles.less
```diff
@@ -51,8 +62,43 @@ body {
...
+    list-style-type: none;
+    padding-left: 0;
+  }
+}
+
+section {
+  padding: 25px;
+}
+
+.hero-unit h1 {
```
toolness: Hmm should all these `px` measurements be converted to `rem` too? I think @gesa is the expert on this one so I will defer to her.

components/page-end-cta.jsx
mmmavis: Oops, didn't realize Aki's fix has been landed already.  I see
```
html, body {
  font-size: 62.5%;
}
```

I'll go through this file and fix units.

less/styles.less
mmmavis: Oops, didn't realize Aki's fix has been landed already.  I see
```
html, body {
  font-size: 62.5%;
}
```

I'll go through this file and fix units.

toolness:
Ok, cool--once the duplicate file is deleted, I think this is good to merge, and we can file tickets for the tweaks.

components/blockquote.jsx
mmmavis: cool, ticket filed here: https://github.com/mozilla/teach.webmaker.org/issues/249

toolness:
Woooooooooooooot

:+1:

less/styles.less
mmmavis: Ticket filed here: https://github.com/mozilla/teach.webmaker.org/issues/248

mmmavis:
Heh, finally! Thanks for reviewing and providing feedback! :dancers:

PR #194, opened 2015-03-10
Use webpack's DefinePlugin to transfer whitelisted process.env vars to the browser.

### LINKED ISSUE #191: Figure out a better way of storing configuration data
For things like mapbox IDs (see #185), google analytics tracking IDs (see #180), and other publicly-facing configuration data (e.g. REST API servers in the near future), it might be nice to allow them to be overridde-able by OS environment variables so that it's easy to tweak configurations without needing to commit changes to the actual repository.

toolness:
This fixes #191 by making it possible for the Mapbox access token and map ID to be overridden via environment variables at build time.

If/when we move to using a dynamic server instead of a static site generator, this mechanism should transfer to the new environment without any extra effort.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,12 @@ string), the boolean is true; otherwise, it's false.
   in situations where absolute URLs are required, such
   as generating a `sitemap.xml` file.
 
+* `MAPBOX_ACCESS_TOKEN` is the [Mapbox][] access token to use for
+  embedded maps in the website. Optional.
+
+* `MAPBOX_MAP_ID` is the Mapbox map ID to use for embedded maps in
+  the website. Optional.
+
 ## References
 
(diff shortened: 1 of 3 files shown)


toolness:

README.md
```diff
@@ -140,6 +140,12 @@ string), the boolean is true; otherwise, it's false.
   in situations where absolute URLs are required, such
   as generating a `sitemap.xml` file.
 
+* `MAPBOX_ACCESS_TOKEN` is the [Mapbox][] access token to use for
```
alicoding: TIL! [Mapbox][] that ref to line:168 O_o

alicoding:
r+! looking great!

toolness:
Woot thanks Ali!

PR #196, opened 2015-03-10
Fix #195 - Rename SingletonMapComponent to Map

### LINKED ISSUE #195: Rename SingletonMapComponent to Map
Ok, since we're not using an `id` anymore in the component, this component no longer needs to be a singleton, so I propose removing the `Singleton` from the class name.

Also, `Component` is redundant, so I recommend getting rid of that too, and just calling this thing `Map`.

What do you think @alicoding?

alicoding:
Closes #195

### DIFF
--- a/lib/map.jsx
+++ b/lib/map.jsx
@@ -1,6 +1,6 @@
 var React = require('react');
 
-var SingletonMapComponent = React.createClass({
+var Map = React.createClass({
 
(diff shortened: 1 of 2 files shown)


alicoding:
@toolness R?

toolness:
Woot the code looks great, as long as Travis is happy, I am happy too!

PR #197, opened 2015-03-10
Fix #190 - Fix all mentions of toolness/mozilla-learning-html in package...

### LINKED ISSUE #190: Fix all mentions of toolness/mozilla-learning-html in package.json
I just realized a bunch of things in `package.json` are referring to the old repository at http://github.com/toolness/mozilla-learning-html. We should change that!

alicoding:
closes #190

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "mozilla-learning-html",
+  "name": "teach.webmaker.org",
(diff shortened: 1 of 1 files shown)


alicoding:
@toolness

toolness:
Awsm thanks!

PR #235, opened 2015-03-10
 LESS/CSS - make sure we use rem

### LINKED ISSUE #228: LESS/CSS - make sure we use rem
Currently every is in `px`, we should make sure we switch over to `rem` instead.

gesa:
Fixes #228 
I did a bunch of visual regression tests and this should have no impact on the current styles.

### DIFF
--- a/less/styles.less
+++ b/less/styles.less
@@ -12,8 +12,13 @@
 @redSidebarColor: #BB3C33;
 @footerBackgroundColor: #f1f1f1;
 
+html, body {
+  font-size: 62.5%;
+}
+
 body {
   font-family: 'Open Sans', sans-serif;
+  font-size: 1.4rem;
   font-weight: 300;
 }
 


less/styles.less
```diff
@@ -12,8 +12,13 @@
 @redSidebarColor: #BB3C33;
 @footerBackgroundColor: #f1f1f1;
 
+html, body {
+  font-size: 62.5%;
```
Pomax: 62.5% of what? at the very least `html` needs a master value set

gesa: the whole point of using 62.5% is because we, as developers, have no control on what a user has set their browser default pixel size to—and we respect that. generally, browser default font size is 16px, so that's what we base our calculations off of. if a user has changed their default font size, then the whole page will respond relative to that user's settings.

toolness:
Looks good to me, thanks aki!

PR #237, opened 2015-03-11
 Detect regular/retina display and render proper images

### LINKED ISSUE #236: Detect regular/retina display and render proper images
As we chatted on PR https://github.com/mozilla/teach.webmaker.org/pull/193 we need to come up with a way to detect and render images for regular/retina display automatically.

@toolness 
> So, this is a weird question, but ... are we actually using the non-@2x images yet? Do web browsers somehow handle this when they see the @2x, or do we need to file a bug to add some JS logic or CSS rules for that? I'm completely unfamiliar with how to adjust to this post-retina world we live in.

@gesa 
> I wrote a [teensy-tiny script](https://github.com/mozilla/webmaker-landing-pages/blob/master/site/resources/js/core.js#L10-L30) for this. Best practice, images should [include a noscript](https://github.com/mozilla/webmaker-landing-pages/blob/5d37adef6ce8272839226e8e99c45036189cc9cf/site/src/from/net-neutrality.html#L39-L42)

gesa:
Fixes #236

### DIFF
--- a/test/browser/imagetag.test.jsx
+++ b/test/browser/imagetag.test.jsx
@@ -1,0 +1,41 @@
+var should = require('should');
+var stubRouterContext = require('./stub-router-context.jsx');
+
+var ImageTag = require('../../components/imagetag.jsx');
+
+describe('ImageTag', function () {
+  var
+      imagetag;
+
+  imagetag = stubRouterContext.render(ImageTag, {
+    src1x: 'foo1x.jpg',
+    src2x: 'foo2x.jpg'
+  });
+
+  it('should set pixel density to 1 by default', function () {
+    imagetag.state.pixelDensity.should.equal(1);
+  });
+
+  it('should set primary src to the value of src1x by default', function () {
+    imagetag.getDOMNode().getAttribute('src').should.equal('foo1x.jpg');
+  });
+
+/*
+  need to figure out how to set window.devicePixelRatio before render.
+
+  describe('on hi-resolution (or "retina") displays', function () {
+    var window = {};
+    before(function () {
+      window.devicePixelRatio = 2.1234;
+    });
+
+    it('pixel density should be set to 2', function () {
+      imagetag.state.pixelDensity.should.equal(2);
+    });
+
+    it('primary src should be set to the value of src2x', function () {
+      imagetag.getDOMNode().getAttribute('src').should.equal('foo2x.jpg');
+    });
+  });
+*/
+});
--- a/components/blockquote.jsx
+++ b/components/blockquote.jsx
@@ -1,10 +1,13 @@
 var React = require('react');
+var ImageTag = require('./imagetag.jsx');
 
 var Blockquote = React.createClass({
   propTypes: {
     className: React.PropTypes.string.isRequired,
     imgSrc: React.PropTypes.string.isRequired,
+    imgSrc2x: React.PropTypes.string,
     imgAlt: React.PropTypes.string.isRequired,
+    imgWidth: React.PropTypes.string,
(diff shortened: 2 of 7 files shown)


toolness:
This looks good! Do you think you could write tests for it?

Off the top of my head, I'm thinking that `window.devicePixelRatio` could be stubbed to return whatever we want, and then we could test to verify that the proper source is being set.

Also, I'm wondering if `Image` is a better name than `ImageTag` -- the `Tag` seems a bit redundant?

Also, what about this [`srcset`](http://www.webkit.org/demos/srcset/) attribute that Webkit supports? Is that being standardized? If so, should we just use that?

toolness:
Erm, so with the landing of the directory reorganization in #238, this PR is likely broken :( Sorry about that... If it's a lot of trouble for you to merge with the latest `develop` branch, let me know and I can help!

gesa:
So the reason it's `ImageTag` and not `Image` is that `Image` is already defined as the javascript image constructor, so I thought redefining it would be a no-no. (the rest of your comments are WIP)

gesa:
I did, however, optimistically add `srcset` ^_^

toolness:
Nice! Also, a fine rationale for ImageTag.

test/browser/imagetag.test.jsx
```diff
@@ -0,0 +1,29 @@
...
+    src1x: 'foo1x.jpg',
+    src2x: 'foo2x.jpg'
+  });
+
+
+  it('should set pixel density to 1 when it is less than 1.5', function () {
+    var window = {
+      devicePixelRatio: 0.9
+    };
+    imagetag.state.pixelDensity.should.be['1'];
```
toolness: Oh, I think the reason this test appears to be passing is because the end of the line should be a function call, not an array reference, i.e.:

```
imagetag.state.pixelDensity.should.be(1);
```

Otherwise, as an array reference it's just returning `undefined` with no side effects. I suspect that once you change it and line 27 to function calls, at least one of them should fail, because I don't think that your component has any way of accessing the local `window` object you're creating at lines 17 and 24.

If my prediction is correct, then one way to proceed is to use [Mocha's hooks](http://mochajs.org/#hooks) to save and restore the actual value of `window.devicePixelRatio` before and after test runs, and set `window.devicePixelRatio` within the tests.

If that doesn't work, e.g. because a host browser doesn't like you overwriting its stuff, then an alternative might be to pass a fake window object in as an additional prop, but use [default props](http://facebook.github.io/react/docs/reusable-components.html#default-prop-values) to have the global window be its default value, so that production code doesn't need to explicitly pass it in.

toolness:
Cool! I think the actual component is good, but there's some [issues](https://github.com/mozilla/teach.webmaker.org/pull/237#discussion_r26266929) with your tests... If you like, we could commit the code without the tests and then bust a separate PR for the tests, or you could keep working on the tests in this PR.

test/browser/imagetag.test.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+    it('pixel density should be set to 2', function () {
+      imagetag.state.pixelDensity.should.equal(2);
+    });
+
+    it('primary src should be set to the value of src2x', function () {
+      imagetag.getDOMNode().getAttribute('src').should.equal('foo2x.jpg');
+    });
+  });
+*/
```
toolness: After I merge this, can you file an issue to uncomment + fix this one?

toolness:
Ultra cool, congrats on your first React component!

test/browser/imagetag.test.jsx
gesa: Done. #259

Issue #233, opened 2015-03-10
Reorganize code


toolness:
Today @alicoding, @gesa, @mmmavis, @vazquez and I discussed a folder structure to better represent the structure of the site, and make it easier for new developers to contribute to it.

It roughly looks like this:

```
components/
  Input.jsx
  Form.jsx
  Homepage.jsx
less/
  common.less
  components/
    Input.less
    Form.less
    Homepage.less
mixins/
  funky-react-mixin.js
lib/
  gulp-index-file-stream.js
  index-static.jsx
  routes.jsx
```

Things left to do:
  - [x] Create the folders and move the JS *but not the less* (#238)
  - [ ] Refactor the LESS (#226)
  - [X] ~~Get rid of the `vendor` directory and delegate to npm for that~~ Filed as #246, we'll get to it later

gesa:
@toolness if I remember correctly, you were going to handle the js rearranging, and once that landed I was going to do the Less. Is that correct? (related—is there an issue for moving the js yet? I'd like to reference it in #226)

toolness:
@gesa - yep, just added the WIP PR as #238!

toolness:
Ok, so I've gotten rid of the `pages` directory now for reasons mentioned in the commit log of e931c09aeb29dcbf703d5b11e9b3554ae65f444d:

> * I suspect that the distinction between 'pages' and 'components' will
>  become blurry as the site evolves and becomes more dynamic. In fact,
>  in some sense the distinction already *is* blurry. If things do
>  become too complicated in the future, though, we can always go back
>  to using the 'pages' directory.
>
> * Pages use lots of components, and having to start every component's
>  `require()` with `../components/` was getting quite tedious.

toolness:
Ok, so as per a discussion in https://github.com/mozilla/teach.webmaker.org/pull/238#issuecomment-78269006, we've decided to **not** put components in their own directories with `index.jsx` files and LESS code in them, because we ran into technical difficulties.

Instead, we'll just have `.jsx` files in the `components` directory, and put LESS for the components in a parallel structure at `less/components/`. I've changed the layout in this issue's description to reflect this.

Does that sound reasonable @gesa?

gesa:
sounds great to me @toolness

hannahkane:
do we need to move this to the March 27 heartbeat?

toolness:
actually i think it can be marked as complete now...

PR #238, opened 2015-03-11
Reorganize directories for JS

toolness:
This helps with #233 by making a `components` directory where we can put all our JSX components.

It also creates a separate JSX file for each page, which makes it easier for us to build out each page.

Things left to do:

- [X] ~~Put each component into its own individual directory with an `index.jsx` file.~~ We're not doing this, see discussion below for rationale.


toolness:
Ok, so I'm now finding that having an `index.jsx` inside a directory doesn't work in as slick a way as `index.js` does. For instance, if you have `foo/index.jsx`, then `require('./foo')` won't work. After a bit of experimentation, I could find these solutions:

1. Explicitly refer to `index.jsx` when loading components (e.g. `require('./foo/index.jsx')`)
2. Name our index files `index.js` and **process all `.js` files as JSX**.  This has a number of drawbacks, though; for instance, syntax highlighters will get confused by the JSX in .js files, and we'll no longer have an easy way to tell our linter to ignore JSX.
3. Do further research into discovering whether there's a way to make `require('./foo')` properly load `index.jsx` files. (There might be some kind of third-party module or StackOverflow question out there that has solved it, which we simply haven't found yet.)
4. Don't put components in their own directories; keep them as their own files. We could then store their LESS in a *parallel* directory structure under the `less` directory. Thus, for instance, the LESS for the `foo` component might be stored in `less/components/foo.less`.

Any thoughts, @alicoding @gesa @mmmavis?

toolness:
Any thoughts on my above comment, @alicoding @gesa @mmmavis?

alicoding:
for `2` I'm using [Babel Sublime](https://github.com/babel/babel-sublime) which can deal with both `.jsx` and `.js` nicely for React, and we can update the build process similar to Kate's [React-Web-App](http://github.com/k88hudson/react-web-app) where webpack will process `.js` as `.jsx` as specified.

toolness:
So on IRC @alicoding just pointed out that [mofo-example-app](https://github.com/MozillaFoundation/mofo-example-app) uses the `.jsx` extension explicitly. Given that precedent, we should probably just go with my option (4) above.

We should probably bring this issue up with mofo engineering at large, though, since it'd be nice for all projects to use the same convention, and there *are* some upsides to simply allowing JSX in all `.js` files.

alicoding:
This is good to merge as long as Travis is happy and green :dancer:

PR #250, opened 2015-03-12
Fixes #248 - convert px to rem

### LINKED ISSUE #248: Go through style.less and convert `px` to `rem`
Now that Aki's patch has landed, we should make sure we are using `rem` instead of `px` (in places other than `media query`).

mmmavis:


toolness:
Thanks mavis!

PR #252, opened 2015-03-12
Add mapbox markers

alicoding:


alicoding:
@toolness R?

components/map.jsx
```diff
@@ -2,25 +2,69 @@ var React = require('react');
...
+    accessToken: React.PropTypes.string.isRequired,
+    mapId: React.PropTypes.string.isRequired,
+    mapParentID: React.PropTypes.string.isRequired,
+    mapClassName: React.PropTypes.string.isRequired
+  },
+  componentDidMount: function() {
+    var that = this;
+    require('mapbox.js'); // this will automatically attach to Window object.
+    // We have to set this globally in order to make .featureLayer() to work
+    L.mapbox.accessToken = this.props.accessToken;
```
toolness: Actually, it looks like it wasn't working earlier because `accessToken` needs to be passed as an option to `featureLayer` constructor too: https://www.mapbox.com/mapbox.js/api/v2.1.5/l-mapbox-featurelayer/

Could you change it to use that?

components/map.jsx
```diff
@@ -2,25 +2,69 @@ var React = require('react');
...
+    // Create a map in the div mapboxDiv element
+    var map = L.mapbox.map(that.props.mapParentID, this.props.mapId)
```
toolness: Is there a particular reason we're using IDs again? Why can't we use an element reference like we did before?

components/map.jsx
```diff
@@ -2,25 +2,67 @@ var React = require('react');
...
-      return (
-        <div ref="mapboxDiv"></div>
-      )
-    }
-  });
+  propTypes: {
+    accessToken: React.PropTypes.string.isRequired,
+    mapId: React.PropTypes.string.isRequired,
+    mapParentID: React.PropTypes.string.isRequired,
+    mapClassName: React.PropTypes.string.isRequired
```
toolness: I think we should just call this `className`, it seems to be the convention for React components.

components/map.jsx
```diff
@@ -2,25 +2,67 @@ var React = require('react');
...
-    render: function () {
-      return (
-        <div ref="mapboxDiv"></div>
-      )
-    }
-  });
+  propTypes: {
+    accessToken: React.PropTypes.string.isRequired,
+    mapId: React.PropTypes.string.isRequired,
+    mapParentID: React.PropTypes.string.isRequired,
```
toolness: We should probably get rid of this since you're not using it anymore?

components/clubs-page.jsx
```diff
@@ -131,6 +130,10 @@ var ClubsPage = React.createClass({
         </HeroUnit>
         <section>
           <WebLitMap/>
+          <div className="mapDiv" id="mapDivID">
+            <Map accessToken={process.env.MAPBOX_ACCESS_TOKEN || 'pk.eyJ1IjoiYWxpY29kaW5nIiwiYSI6Il90WlNFdE0ifQ.QGGdXGA_2QH-6ujyZE2oSg'}
+           mapId={process.env.MAPBOX_MAP_ID || 'alicoding.ldmhe4f3'} mapParentID={'mapDivID'} mapClassName={'mapDivChild'}/>
```
toolness: We should probably get rid of `mapParentID` and the `id` on line 133 since you're not using them anymore?

components/map.jsx
```diff
@@ -2,25 +2,69 @@ var React = require('react');
...
+    }]).addTo(map);
+  },
+  componentWillUnmount: function() {
+    this.map.remove();
+  },
+  // Called on initialization and after each change to the components
+  // props or state
+  render: function() {
+    return (
+      <div className={this.props.mapClassName} ref="mapboxDiv"></div>
```
toolness: oh, we can actually get rid of the `mapboxDiv` ref--calling `this.getDOMNode()` does the same thing, since `mapboxDiv` is the top-level element of the component!

alicoding:
all done @toolness :D

components/map.jsx
```diff
@@ -2,25 +2,67 @@ var React = require('react');
...
-      )
-    }
-  });
+  propTypes: {
+    accessToken: React.PropTypes.string.isRequired,
+    mapId: React.PropTypes.string.isRequired,
+    className: React.PropTypes.string.isRequired
+  },
+  componentDidMount: function() {
+    var that = this;
```
toolness: Is this actually needed? Not a big deal, but...

toolness:
Woot looks good!! You can remove the [`that`](https://github.com/mozilla/teach.webmaker.org/pull/252#discussion_r26330016) if you want, and then feel free to merge once travis is happy.

components/map.jsx
alicoding: ah it was needed when I was doing `.ready()` but we have changed that. I will remove.

PR #269, opened 2015-03-13
Fixes #244 - Build modal dialogs

### LINKED ISSUE #244: Build modal dialogs
According to the [mockup](https://redpen.io/wc356809d269ed0827) of the /clubs page we need modal dialogs to display some forms.  I'm opening this ticket separately so @alicoding doesn't have to wait longer for my [big Clubs page PR](https://github.com/mozilla/teach.webmaker.org/pull/193) to land.

@toolness 
> Ugh good question. I've been writing my own modals for previous projects, because I'm slightly annoyed by the "boil the ocean" approach of [react-bootstrap](http://react-bootstrap.github.io/), but it might ultimately be more prudent to just go with the latter... Ugh. For now, if you can try reusing the [`modal.jsx`](https://github.com/mozilla/webmaker-addons/blob/gh-pages/src/modal.jsx) I wrote for the webmaker addon, that would be preferable since its implementation is so simple--then we could migrate to something else if we decide it's too limiting.

Is what you said on https://github.com/mozilla/teach.webmaker.org/issues/186#issuecomment-78112843 still the approach we are going to take?

mmmavis:
**Done**
-  'Add Your Club' modal dialog done
- 'Learn More About Mozilla Learning Clubs' modal dialog


mmmavis:
@toolness @alicoding @gesa I *think* this PR is now ready for review :grimacing:

alicoding:
Do we want form validation in this ticket now? IMO it might make sense to do it now?

components/clubs-page.jsx
```diff
@@ -105,9 +106,12 @@ var BottomCTA = React.createClass({
     return(
       <div className="row">
         <div className="col-sm-offset-2 col-sm-8 col-md-offset-2 col-md-8 col-lg-offset-2 col-lg-8">
-            <img className="divider" src="/img/clubs-line-divider.svg" alt="line divider" />
-          <PageEndCTA linkTo={this.props.ctaLink} ctaBtnText="Add your club to the map">
-            <p>Do you meet regularly with a group of learners to increase web literacy skills?</p>
+          <PageEndCTA linkTo={this.props.ctaLink}>
+            <div>
+              <img className="divider" src="/img/clubs-line-divider.svg" alt="line divider" />
+              <p>Do you meet regularly with a group of learners to increaset web literacy skills?</p>
```
alicoding: s/increaset/increase

alicoding:
I'm wondering about the modal itself. Should we also add onClose when the user click outside the modal and or click the `ESC` key? I think this might be needed to make it easy for a11y users?

less/components/index.less
```diff
@@ -5,6 +5,8 @@
 @import "hero-unit";
 @import "icon-link";
 @import "imagetag";
+@import "modal";
+@import "map";
```
alicoding: Good catch!!

alicoding:
Very good start @mmmavis! I'd like some discussion regarding my questions about the validation and that small typo and this is almost good to go!

toolness:
> I'm wondering about the modal itself. Should we also add onClose when the user click outside the modal and or click the ESC key? I think this might be needed to make it easy for a11y users?

Yeah, good idea--but I think we should first merge this PR and file that as a separate ticket. Since this is our first modal, there's a lot of "new stuff" riding on it and I think we can move faster together on it if we commit something with basic functionality for now, and then focus on improving it after it lands!

Same goes with the validation you brought up... I'd actually like to keep some of the validation logic shared w/ the [teach-api](https://github.com/mozilla/teach-api) codebase so that when its back-end validation changes, our front-end validation logic doesn't mismatch.

components/clubs-page.jsx
```diff
@@ -116,18 +120,95 @@ var BottomCTA = React.createClass({
...
   render: function() {
-    var theCtaLink = "fixme"; // CTA link should be the same for hero CTA and page bottom CTA
     return (
       <div>
         <HeroUnit image="/img/hero-clubs.jpg">
           <h1>Mozilla Learning Clubs</h1>
-          <div><Link to={theCtaLink} className="btn btn-awsm">Add Your Club</Link></div>
-          <p className="learn-more">or <Link to="fixme">find out more</Link> about us</p>
+          <div><a className="btn btn-awsm" onClick={this.showModal} data-modal-name="ModalAddYourClub">Add Your Club</a></div>
+          <div><p className="learn-more">or <a onClick={this.showModal} data-modal-name="ModalLearnMore">find out more</a> about us</p></div>
```
toolness: I wonder if you should make `showModal` take the modal name as its argument and use `showModal.bind`, similar to how you did with [`closeModal`](https://github.com/mozilla/teach.webmaker.org/pull/269/files#diff-043f02d87f17791d954cfbed0983455fR233)? It makes the style more consistent, and I think not relying on the DOM could also make it easier to unit test.

toolness:
Cool, save for my one [comment](https://github.com/mozilla/teach.webmaker.org/pull/269#discussion_r26503521) I think this looks good!

Bugs we should file after this lands:

- [ ] Allow ESC to close a modal.
- [ ] Allow clicking outside a modal to close it.
- [ ] Ensure that modals have proper accessibility markup.
- [ ] Hook the modal's submit button up to the Teach API so it actually submits data.
- [ ] The modal's form fields should be validated.
- [ ] Add unit tests for the modal.

components/clubs-page.jsx
```diff
@@ -116,18 +120,95 @@ var BottomCTA = React.createClass({
...
+      }
+    };
+  },
+  closeModal: function(targetModal) {
+    var modalShownState = this.state.modalShown;
+    modalShownState[targetModal] = false;
+    this.setState(modalShownState);
+  },
+  showModal: function(e) {
+    var targetModal = e.target.getAttribute("data-modal-name");
```
toolness: One fear I have about this is that if we ever decide to make the link more complex, e.g. by putting an image inside the `<a>`, then it could be possible for `e.target` to be the image instead of the link, in which case `data-modal-name` would be falsy!

Since we have the benefit of React here I suggest we use [bind](https://github.com/mozilla/teach.webmaker.org/pull/269/files#r26503521) instead.

components/clubs-page.jsx
mmmavis: heh good eye :D

mmmavis:
+1 on what @toolness said - I think we should keep this PR basic and small and evolve our modal dialogs later.  
Let me work on the `showModal` tweak now :D

alicoding:
Agreed. Lets land this once you fixed all the stuff then file bugs to get everything else done :+1:

mmmavis:
Cool, fixes are in.  Let me know and I'll rebase & merge.

components/clubs-page.jsx
```diff
@@ -105,9 +106,12 @@ var BottomCTA = React.createClass({
...
       <div className="row">
         <div className="col-sm-offset-2 col-sm-8 col-md-offset-2 col-md-8 col-lg-offset-2 col-lg-8">
-            <img className="divider" src="/img/clubs-line-divider.svg" alt="line divider" />
-          <PageEndCTA linkTo={this.props.ctaLink} ctaBtnText="Add your club to the map">
-            <p>Do you meet regularly with a group of learners to increase web literacy skills?</p>
+          <PageEndCTA linkTo={this.props.ctaLink}>
+            <div>
+              <img className="divider" src="/img/clubs-line-divider.svg" alt="line divider" />
+              <p>Do you meet regularly with a group of learners to increase web literacy skills?</p>
+              <a className="btn btn-awsm" onClick={this.props.onClick} data-modal-name="ModalAddYourClub">Add your club to the map</a>
```
toolness: Since we're not using `data-modal-name` anymore can you get rid of this?

Other than that, this looks great, feel free to merge when Travis is happy!

components/modal.jsx
```diff
@@ -0,0 +1,31 @@
...
+      this.props.onClose();
+    }
+  },
+  render: function() {
+    return (
+      <div className="modal show">
+        <div className="modal-backdrop"></div>
+        <div className="modal-dialog" onClick={this.handleClick}>
+          <div className="modal-header">
+            <button type="button" className="close" data-dismiss="modal" aria-hidden="true">x</button>
```
toolness: Doh, this is a problem w/ my code since I gave you this file, but I just realized that `data-dismiss` shouldn't be in here either, since it's specific to bootstrap's jQuery-based JavaScript stuffs. If it's easy for you to remove then that would be cool, but otherwise I will do it since it's my fault!

mmmavis: No worries, @toolness .  I don't think that attribute is ever used in my code.  I should've excluded it!  Thanks for spotting this

mmmavis:
Thanks for merging it.  Travis took so long to build this time :zzz: :sleeping:

PR #281, opened 2015-03-13
Add mapbox marker clusters

alicoding:


alicoding:
@toolness R?

alicoding:
@toolness once you have time if you can review this patch :)

components/map.jsx
```diff
@@ -9,60 +9,44 @@ var Map = React.createClass({
...
-      },
-      properties: {
-        title: 'First Point',
-        description: 'Party description?',
-        "icon": {
-          "iconUrl": "/img/map-marker.svg",
-          "iconSize": [33, 33], // size of the icon
-          "iconAnchor": [15, 15] // point of the icon which will correspond to marker's location
-        }
+    L.mapbox.accessToken = this.props.accessToken;
```
toolness: o snap it's the return of `L.mapbox.accessToken`! Does the cluster plugin not allow an access token to be passed into it?

If we're using a global access token, then I think it shouldn't be passed-in to a `Map` component, since we don't want different `Map` components potentially stepping over each other (or conversely, requiring devs to constantly pepper the `accessToken` throughout multiple maps)... I recommend just defining a module-level variable in `map.jsx` called `accessToken`, rather than passing it in through a prop. What do you think?

components/map.jsx
```diff
@@ -9,60 +9,44 @@ var Map = React.createClass({
...
+    // TODO: The ID in the featureLayer should be change to our internal
+    // GeoJSON once we have the server side database setup.
+    var mb = L.mapbox.featureLayer('examples.map-h61e8o8e');
```
toolness: Er, what is this ID referring to? Also, does our server need to "cluster" things itself? Right now the data model on the server-side is just a list of [`Club` objects](https://teach-api.herokuapp.com/api/clubs/) with latitude and longitude...

alicoding: This is just an example geoJson object to demonstrate that cluster does work. This should be replace to our own internal one once we have the server-side, modal and this hooked up.

components/map.jsx
```diff
@@ -1,70 +1,54 @@
...
+      });
+      that.map.addLayer(markers);
+
     });
 
-    // Add features to the map.
-    MapLayer.setGeoJSON(geoJson);
   },
   componentWillUnmount: function() {
+    console.log(this.map)
```
toolness: oops can you remove this?

alicoding: Oopsy!

components/map.jsx
```diff
@@ -1,70 +1,54 @@
 var React = require('react');
-
+var mapboxId = process.env.MAPBOX_MAP_ID || 'alicoding.ldmhe4f3';
```
toolness: err it was the access token i thought should be moved here, since we're setting it globally on `L.mapbox.accessToken`--not the map id... Though *defaulting* the `mapId` property to `alicoding.ldmhe4f3` probably isn't a bad idea, as it makes it super easy to plop a map anywhere!

alicoding: So I did for both `accessToken` and `mapId` now :P

components/map.jsx
toolness: Ah ok, so long as the server doesn't need to do any fancy clustering logic itself, I am cool with this... if the server *does* need to be modified to cluster stuff itself, then I have concerns about clustering being a P1 for launch!

toolness:
O wow so many Webmaker Clubs

screen shot 2015-03-17 at 12 34 17 pm

I'm noticing that when I click on a non-cluster marker, there's no popup like there is in the current `develop` branch of the site:

screen shot 2015-03-17 at 12 35 52 pm

Could you bring that back, since we have to show information there?

alicoding:
@toolness I have fixed that problem and also updated the PR to use teachAPI :)

alicoding:
@toolness I'm stuck on one failing test, but I have to leave for now and will get back and fix this test later today. Unless you know what is wrong and can help point out to me quickly :)

toolness:
Ah yeah, I think it's because you're not setting `this.map` until the teach API request is finished... and because it hasn't finished by the time the test suite tears down the component, it's crashing.

Another thing to consider is what happens if the teach API request completes *after* the component has been unmounted from the DOM--which is certainly what will happen in the test suite, but it might also happen in the live site too, e.g. if the network is really slow and the user navigates to a different page on he site before the teach API request finishes.

So here's a diff that fixes both those problems:

```diff
diff --git a/components/map.jsx b/components/map.jsx
index e7bdd89..210ab04 100644
--- a/components/map.jsx
+++ b/components/map.jsx
@@ -30,6 +30,11 @@ var Map = React.createClass({
     require('leaflet.markercluster');
     L.mapbox.accessToken = accessToken;
     teachAPI.getAllData(function(err, data) {
+      if (!that.isMounted()) {
+        // We were unmounted before the API request completed,
+        // so do nothing.
+        return;
+      }
       if(err) {
         console.log(err)
         return;
@@ -72,7 +77,9 @@ var Map = React.createClass({
     });
   },
   componentWillUnmount: function() {
-    this.map.remove();
+    if (this.map) {
+      this.map.remove();
+    }
   },
   // Called on initialization and after each change to the components
   // props or state
```

That said, we should ideally be mocking out that Teach API call for the tests, but whatevs, we can deal with it in a different PR :)

components/map.jsx
```diff
@@ -1,68 +1,75 @@
...
-        "icon": {
-          "iconUrl": "/img/map-marker.svg",
-          "iconSize": [33, 33], // size of the icon
-          "iconAnchor": [15, 15] // point of the icon which will correspond to marker's location
-        }
+    require('leaflet.markercluster');
+    L.mapbox.accessToken = accessToken;
+    teachAPI.getAllData(function(err, data) {
+      if(err) {
+        console.log(err)
```
toolness: Hmm we should probably figure out something more robust to do here, but can leave that for a later PR...

components/map.jsx
```diff
@@ -1,68 +1,75 @@
...
-      properties: {
-        title: 'Second Point',
-        description: 'Address I guess?',
-        "icon": {
-          "iconUrl": "/img/map-marker.svg",
-          "iconSize": [33, 33],
-          "iconAnchor": [15, 15]
+      var geoJSON = geoJSONit(data);
+      that.map = L.mapbox.map(that.getDOMNode())
+        .setView([40.73, -50.011], 3)
```
toolness: For the sake of not being biased to any one particular region of the world, we might want to just set this to "the whole world" or something... Another alternative is to use [`map.fitBounds()`](http://leafletjs.com/reference.html#map-fitbounds) and pass it the [`.getBounds()`](http://leafletjs.com/reference.html#featuregroup-getbounds) of whatever feature layer thingy you're using, but odds are that will eventually just be the whole world.

Anyhow, we can leave it for another PR too, don't worry about fixing it for this one.

lib/teach-api.js
```diff
@@ -73,6 +73,20 @@ _.extend(TeachAPI.prototype, {
     }
 
     return req;
+  },
+  getAllData: function(callback) {
```
toolness: Do you mind renaming this to `getAllClubsData`, since it's very likely that we'll be storing more than just clubs in the Teach API?

lib/teach-api.js
```diff
@@ -73,6 +73,20 @@ _.extend(TeachAPI.prototype, {
...
+    request('get', process.env.TEACH_API_URL || this.baseURL + "/api/clubs")
```
toolness: Oh, you should just use `this.baseURL` here, as its value already takes `process.env.TEACH_API_URL` into account.

lib/teach-api.js
```diff
@@ -73,6 +73,20 @@ _.extend(TeachAPI.prototype, {
...
+      .accept('json')
+      .end(function(err, res) {
+        if(err) {
+          return callback(err);
+        }
+        try {
+          callback(null, JSON.parse(res.text));
```
toolness: I believe you can actually just return `res.body` here, which contains the already-parsed response content.

lib/teach-api.js
```diff
@@ -73,6 +73,20 @@ _.extend(TeachAPI.prototype, {
...
+        } catch(e) {
+          return callback(e);
```
toolness: No need for this try/catch since `res.body` will already be the parsed content, so there's no `JSON.parse` to throw an exception!

alicoding:
All done @toolness :D

toolness:
You goose! There are [linting errors](https://travis-ci.org/mozilla/teach.webmaker.org/builds/54791441#L443-L469) now, but they're pretty minor, so I'm going to merge this and fix the linting issues myself. :P

PR #287, opened 2015-03-16
Add login support via Teach API

toolness:
This adds login support via the [Teach API](https://github.com/mozilla/teach-api).

Notes:
* I'm using [superagent](https://github.com/visionmedia/superagent) here because it works on both in node and the browser. It's quite likely that we'll eventually want a dynamic server, or at least have the static site generator pull data from the Teach API for its generated pages, so it'll be nice if our code works in both environments out-of-the-box.
* Ideally the teach API object should be passed into any component that needs it, but it's cumbersome and arguably what [React Contexts](https://www.tildedave.com/2014/11/15/introduction-to-contexts-in-react-js.html) were made for. However, Contexts aren't documented yet, which makes me reluctant to use them; furthermore, because `react-router` uses them, it's actually impossible for us to pass them through our view heirarchy *without* using them. So instead of that, I'm just making the teach API a global singleton for now, and we can migrate it to use Contexts once they're properly added to React. (Well, either that or we just decide to use Contexts anyways.)
* Right now we're using `window.alert()` for error conditions; this isn't awesome, but given that we'll be migrating away from persona-based login ASAP anyways, I'm hoping it's ok as a stopgap for launch.
* We're now including sinon via a script tag instead of a require; see the commit log of fb8c7bc74dbaab5e5377a001631e0f091d30f51b for more details.
* Previously, our browser tests weren't cleaning up components they created, which resulted in lots of spurious (but still helpful!) memory leak warnings. So as of 7a9749a and 65d7313, they all clean up their stuff.
* The Teach API singleton has a `request()` method that is just like superagent's eponymous function, but it also adds the proper authorization header if the user is logged in. So we can use this whenever we need to make authenticated requests to the Teach API (I probably should've included this in a separate PR, apologies).

TODOs:
- [x] Add tests for `lib/teach-api`.
- [x] Add tests for the `Login` JSX component.
- [x] Move the `Login` component to its own file.
- [X] ~~Have the top-level application manage the `TeachAPI` instance and pass it into the `Login` component.~~ See above note.
- [X] ~~Don't use `window.alert()` for error conditions.~~ See above note.
- [x] Add an environment variable allowing developers to override the Teach API server URL.


toolness:
Hey @alicoding, can you take a look at this? Sorry it's kinda big--a lot of it is tests, though...

components/login.jsx
```diff
@@ -0,0 +1,112 @@
...
+var Router = require('react-router');
+var Link = Router.Link;
+
+var teachAPI = require('../lib/teach-api');
+
+var Login = React.createClass({
+  getDefaultProps: function() {
+    return {
+      teachAPI: teachAPI,
+      alert: defaultAlert
```
alicoding: We don't need to declare 'defaultAlert' anywhere first?

components/login.jsx
```diff
@@ -0,0 +1,112 @@
...
+    };
+  },
+  componentDidMount: function() {
+    var teachAPI = this.props.teachAPI;
+
+    teachAPI.on('login:error', this.handleApiLoginError);
+    teachAPI.on('login:cancel', this.handleApiLoginCancel);
+    teachAPI.on('login:success', this.handleApiLoginSuccess);
+    teachAPI.on('logout', this.handleApiLogout);
+    this.setState({username: teachAPI.getUsername()});
```
alicoding: In case of an error this will be **null**?

components/login.jsx
toolness: I tend to put function declarations below variable declarations, so it's defined near the end of this file, but we can either move the function declaration up, or define the function as a variable above the component if you want.

components/login.jsx
```diff
@@ -0,0 +1,112 @@
...
+  },
+  componentWillUnmount: function() {
+    var teachAPI = this.props.teachAPI;
+
+    teachAPI.removeListener('login:error', this.handleApiLoginError);
+    teachAPI.removeListener('login:cancel', this.handleApiLoginCancel);
+    teachAPI.removeListener('login:success', this.handleApiLoginSuccess);
+    teachAPI.removeListener('logout', this.handleApiLogout);
```
alicoding: Do we ever need to clear **username**? Sorry I'm reviewing this on my phone right now if you did somewhere within this patch ignore me

components/login.jsx
alicoding: Oh it's fine! I just happen to do this on my phone. Sorry about that (my train is delayed lol)

components/login.jsx
```diff
@@ -0,0 +1,112 @@
...
+  },
+  getInitialState: function() {
+    return {
+      username: null,
```
alicoding: Now I see. This will always reset it to null right?

lib/teach-api.js
```diff
@@ -0,0 +1,80 @@
...
+  getUsername: function() {
+    var info = this.getLoginInfo();
+    return info && info.username;
+  },
+  startLogin: function() {
+    if (!(process.browser && window.navigator.id)) {
+      return this.emit('login:error',
+                       new Error('navigator.id does not exist'));
+    }
+    window.navigator.id.get(function(assertion) {
```
alicoding: This is browserID thing right?

alicoding:
This is looking good @toolness I would like to test is once I get to the office but the codes are good to me

alicoding:
tested and working as expected! all tests are passing, travis is happy and I'm happy \o/

toolness:
Woot thanks Ali!

PR #292, opened 2015-03-16
Get rid of the vendor directory. Bye bye bootstrap in our repo!

### LINKED ISSUE #246: Get rid of the vendor directory
As mentioned in #233, it'd be nice if we could get rid of the `vendor` directory and delegate to npm or something for that stuff.

This means removing the following files/directories and obtaining them through some other means:
- [x] `vendor/prefixer.less`
- [ ] `vendor/bootstrap`

This isn't a super high priority so it shouldn't be added to v1 or anything.

gesa:
Fixes #246


toolness:
niiiiiiiiiiiiice!! thanks aki!!

PR #297, opened 2015-03-17
Add div.modal-backdrop to body when modal created/destroyed. And other modal things.

### LINKED ISSUE #304: Modal Dialog - Add unit tests for the modal
(Per discussion on https://github.com/mozilla/teach.webmaker.org/pull/269#issuecomment-81800754)

toolness:
The existing code puts a `.modal-backdrop` *inside* the modal, which prevents users from being able to type anything into the modal fields. This PR changes that behavior so that a `<div class="modal-backdrop">` is added to the `body` element when the modal is created, and removed when it's destroyed, similar to how [`bootstrap/modal.js`](https://github.com/twbs/bootstrap/blob/master/js/modal.js#L193-L194) does things.

It also adds tests to verify that the new functionality works, as well as existing functionality, which fixes #304.

@mmmavis can you review this?

### DIFF
--- a/test/browser/modal.test.jsx
+++ b/test/browser/modal.test.jsx
@@ -26,6 +26,11 @@ describe("modal", function() {
 
   afterEach(removeModal);
 
+  it('closes modal when area outside modal is clicked', function() {
+    TestUtils.Simulate.click(modal.getDOMNode());
+    onClose.callCount.should.equal(1);
+  });
+
   it('closes modal when close button is clicked', function() {
     var closeButton = TestUtils.findRenderedDOMComponentWithClass(
       modal,
--- a/test/browser/modal.test.jsx
+++ b/test/browser/modal.test.jsx
@@ -1,0 +1,39 @@
+var should = require('should');
+var sinon = window.sinon;
+var React =require('react/addons');
+var TestUtils = React.addons.TestUtils;
+
+var Modal = require('../../components/modal.jsx');
+
+describe("modal", function() {
+  var modal, onClose;
+
+  function removeModal() {
+    if (modal) {
+      React.unmountComponentAtNode(modal.getDOMNode().parentNode);
+      modal = null;
+    }
+  }
+
+  beforeEach(function() {
+    onClose = sinon.spy();
+    modal = TestUtils.renderIntoDocument(
+      <Modal onClose={onClose}>
+        i am modal content
+      </Modal>
+    );
+  });
+
+  afterEach(removeModal);
+
+  it('creates body > .modal-backdrop when mounted', function() {
+    document.querySelectorAll('body > .modal-backdrop').length
+      .should.equal(1);
+  });
+
+  it('removes body > .modal-backdrop when mounted', function() {
+    removeModal();
+    document.querySelectorAll('body > .modal-backdrop').length
+      .should.equal(0);
+  });
+});
--- a/test/browser/modal.test.jsx
+++ b/test/browser/modal.test.jsx
@@ -31,7 +31,7 @@ describe("modal", function() {
       .should.equal(1);
   });
 
-  it('removes body > .modal-backdrop when mounted', function() {
+  it('removes body > .modal-backdrop when unmounted', function() {
     removeModal();
     document.querySelectorAll('body > .modal-backdrop').length
       .should.equal(0);
--- a/test/browser/modal.test.jsx
+++ b/test/browser/modal.test.jsx
@@ -26,6 +26,24 @@ describe("modal", function() {
 
   afterEach(removeModal);
 
+  it('closes modal when close button is clicked', function() {
+    var closeButton = TestUtils.findRenderedDOMComponentWithClass(
+      modal,
+      'close'
+    );
+    TestUtils.Simulate.click(closeButton);
+    onClose.callCount.should.equal(1);
+  });
+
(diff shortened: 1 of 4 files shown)


test/browser/modal.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+    if (modal) {
+      React.unmountComponentAtNode(modal.getDOMNode().parentNode);
+      modal = null;
+    }
+  }
+
+  beforeEach(function() {
+    onClose = sinon.spy();
+    modal = TestUtils.renderIntoDocument(
+      <Modal onClose={onClose}>
```
toolness: We don't do anything with `onClose` yet, but we will in future tests, so I figured I might as well bust it up now.

toolness:
Eh, I went ahead and added tests for the close box too, that way all the JS in the component is tested now.

mmmavis:
Ah, @toolness , now I remember why I altered your code in [webmaker-addon](https://github.com/mozilla/webmaker-addons/blob/gh-pages/src/modal.jsx#L6-L14) (though I introduced another bug heh). If we created `.modal-backdrop` in `componentDidMount` the height of `.modal-backdrop` doesn't get calculated right.  It doesn't extend full page and instead is always positioned at page top and the height of it is always the height of your browser's single viewport.

If you scroll down and click on the bottom CTA on the Clubs page, you will see what I mean.

I wonder if override CSS rules would fix this problem?

test/browser/modal.test.jsx
```diff
@@ -0,0 +1,57 @@
...
+    document.querySelectorAll('body > .modal-backdrop').length
+      .should.equal(1);
+  });
+
+  it('removes body > .modal-backdrop when unmounted', function() {
+    removeModal();
+    document.querySelectorAll('body > .modal-backdrop').length
+      .should.equal(0);
+  });
+});
```
mmmavis: I'm not very familiar with test writing since it's been a while since I last touched it (shhhhh :wink: ) But this looks great @toolness !  

We might want to edit the `.modal-backdrop` test part when we figure out where to insert that DOM.

toolness:
Ohh, interesting! Huh... I wonder if bootstrap's JS code gets around that issue somehow. Hmmmmmm.

I will investigate further and get back to you.

toolness:
Actually @mmmavis, I don't think I'm experiencing the bug you're talking about... this is what scrolling down to the bottom and clicking the CTA button gives me:

<a href="https://toolness-media.herokuapp.com/v/teach-site-modal-backdrop/play/loop">Play 'Teach site modal backdrop' video</a>

mmmavis:
ummm, how do I record video @toolness . This is not what I'm seeing :hushed:

toolness:
Oh weird! I was wondering if it might have something to do with the landing of #292, since perhaps the bootstrap we're now using via npm fixed CSS bugs in the one we were using from our vendor directory, but if you're still experiencing the bug on my branch, that is definitely funky. What browser/version/OS are you using? (My video is showing Firefox 36, and Chrome 41 works similarly.)

I took my video using a personal copy of [Snagit](http://www.techsmith.com/snagit.html), but I bet mozilla has licenses for some more robust screencasting software!

mmmavis:
Firefox 36.0.1
Chrome 41.0.2272.89
on Mac OSX 10.9.5

iamjessklein:
It looks the same on my Firefox as the movie @toolness posted.

toolness:
@mmmavis does it exhibit the same erroneous behavior for you at http://mozilla-learning-html.s3-website-us-east-1.amazonaws.com/clubs/ (I just posted a build of my branch there temporarily)?

mmmavis:
@toolness it works fine on http://mozilla-learning-html.s3-website-us-east-1.amazonaws.com/clubs/

toolness:
Um.... weird. Now I am on my PC desktop and the issue you are talking about **is** happening.

toolness:
Ahh this is *so weird*!

The `bootstrap.css` in my desktop's `node_modules/bootstrap/dist/css` looks like this:

```css
.modal-backdrop {
  position: absolute;
  top: 0;
  right: 0;
  left: 0;
  background-color: #000;
}
```

While the exact same file on my laptop, which I made the earlier movie on, looks like this:

```css
.modal-backdrop {
  position: fixed;
  top: 0;
  right: 0;
  bottom: 0;
  left: 0;
  z-index: 1040;
  background-color: #000;
}
```

Looking at the CSS, it makes sense that the latter one works properly, while the former one doesn't... But I'm confused as to why my two systems have totally different CSS files on them.

Ah, looking at the `node_modules/bootstrap/package.json`, it appears that the former is actually bootstrap 3.3.2, while the latter is 3.3.4. Re-running `npm install` isn't doing anything because our project's `package.json` just wants anything at bootstrap 3.3.2 or later!

I'll fix our `package.json` and add it to this PR!

toolness:
Ok @mmavis, if you update this PR and run `npm install`, hopefully modals should work for you!

toolness:
Ok @mmmavis, if you update this PR and run npm install, hopefully modals should work for you!

toolness:
Hmm, I think @mmmavis is on vacation now, but I am pretty sure my latest commit fixes her issues with this PR, so I am going to merge it.

@mmmavis if you see this when you come back and you're unsatisfied, please file an issue and we'll get back to it. Thanks!

PR #309, opened 2015-03-17
Like mozilla

ScottDowne:


toolness:
@gesa can you review this?

ScottDowne:
I added the start of the teach-like-mozilla page as a test, but I can remove that bit.

ScottDowne:
Steps to test:

1. go to http://localhost:8008/teach-like-mozilla/

Expected: if retina, should see a @2x image in the source if regular should see a regular image.

Actually, I don't think it uses the regular if no 2x is supplied. We probably want that.

gesa:
One problem I see with this approach is that the browser is going to download both images no matter what. We want to set this up so that the browser will *only* download the 1x if the DPR is low, and *only* download the 2x if the DPR is high.

components/hero-unit.jsx
```diff
@@ -7,11 +7,16 @@ var HeroUnit = React.createClass({
   render: function() {
     return (
```
gesa: what about a function here to calculate the device DPR and set the appropriate source to a variable that the `hero-unit` class can use?

components/hero-unit.jsx
```diff
@@ -7,11 +7,16 @@ var HeroUnit = React.createClass({
...
       <div className="row">
-        <div className="col-md-12 hero-unit" style={{
+        <div className="col-md-12 hero-unit hero-unit-1x" style={{
           backgroundImage: 'url(' + this.props.image + ')'
         }}>
           {this.props.children}
         </div>
+        <div className="col-md-12 hero-unit hero-unit-2x" style={{
+          backgroundImage: 'url(' + this.props.image2x || this.props.image + ')'
```
gesa: this is a fantastic way to deal with the fallback if someone doesn't include the 2x in the `HeroUnit` tag .

components/hero-unit.jsx
ScottDowne: I'll need to figure out how that's done but sounds like a plan.

gesa: Idk if this will help, but I had a similar problem recently and kind of went about it in an overly complicated way: https://github.com/mozilla/teach.webmaker.org/blob/develop/components/imagetag.jsx#L4-L16 (mostly due to the fact that CSS couldn't solve my problem)

ScottDowne: it did :D

gesa:
perrrrrrfect. R+!

PR #312, opened 2015-03-18
Use contexts to manage modal display/hiding in an accessible way.

### LINKED ISSUE #301: Modal Dialog - Ensure that modals have proper accessibility markup
(Per discussion on https://github.com/mozilla/teach.webmaker.org/pull/269#issuecomment-81800754)

toolness:
This fixes a number of issues with the modals introduced in #269:

* It positions the modal as a sibling element to the page content, which allows us to give the page content an `aria-hidden="true"` attribute while modals are showing, which is a best practice for accessibility (see [The Incredible Accessible Modal Window](http://accessibility.oit.ncsu.edu/training/aria/modal-window/version-3/) feature 3).
* It ensures that while a modal is shown, other interactive elements on the page can't be focused with the keyboard (see *The Incredible Accessible Modal Window* feature 4).
* It uses [React Contexts](https://www.tildedave.com/2014/11/15/introduction-to-contexts-in-react-js.html) to manage the showing/hiding of modals, which significantly simplifies logic for pages that need to show modals. Now they just use the `ModalManagerMixin` and call `this.showModal(SomeModalClass, {someProp: "blah"})` whenever they want to show a modal.

This PR fixes #301.

TODOs:

- [x] Move the modal functionality into a mixin, so components that want to display modals don't need to define `contextTypes`.

### DIFF
--- a/components/clubs-page.jsx
+++ b/components/clubs-page.jsx
@@ -123,9 +123,7 @@ var BottomCTA = React.createClass({
 var ModalAddYourClub = React.createClass({
   render: function() {
     return(
-      <Modal
-        modalTitle="Add Your Clubs To The Map"
-        onClose={this.props.onClose}>
+      <Modal modalTitle="Add Your Clubs To The Map">
         <form>
           <fieldset>
             <label>What is the name of your Club?</label>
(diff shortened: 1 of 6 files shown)


toolness:
@alicoding can you review this? You will need to skim through [React Contexts](https://www.tildedave.com/2014/11/15/introduction-to-contexts-in-react-js.html) to figure out the implementation, unfortunately, but I couldn't think of any other way of implementing this >.<

alicoding:
Code looks good and no confusion when I reviewed them. I have tested with the sense of using keyboard only to see if it is accessible and try to break it if possible and nothing seems to be a problem at all. I'm R+ing this patch as long as travis is happy I'm happy :)

toolness:
Wooooooooot thanks ali!

PR #314, opened 2015-03-18
Testing out some cross-browser testing with sauce

k88hudson:
Just playing around with generating cross-browser screenshots and stuff with sauce

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,12 @@ node_js:
 - '0.12'
 env:
   global:
-    secure: G3YBwV29SBIqtWfP1EIyaABZ65jvzFT15vUMRSf8IZgINRkRl8p7Js38ioaYahlCIVVom02iulY5N/n/Q5qdsgckZUGYe0FYy4+Qz1Qr69rATpFCpREPh02lfkm5TojjefGdDqlm9CUJvJGhtDXIOUlAK66yd85O2gyZb9DK0CQ=
+    - secure: G3YBwV29SBIqtWfP1EIyaABZ65jvzFT15vUMRSf8IZgINRkRl8p7Js38ioaYahlCIVVom02iulY5N/n/Q5qdsgckZUGYe0FYy4+Qz1Qr69rATpFCpREPh02lfkm5TojjefGdDqlm9CUJvJGhtDXIOUlAK66yd85O2gyZb9DK0CQ=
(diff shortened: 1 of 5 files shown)


toolness:
Sweet!! :+1:

k88hudson:
Yeah I'm just sort of playing around. If you want to check out an example of a test:

https://saucelabs.com/tests/ee8e3f3500c141639daeb5150a356079

toolness:
Ooo nice!

Another thing vaguely kinda related to this is http://teach.mofostaging.net/test/manual/. It is kind of like a [contact sheet](http://en.wikipedia.org/wiki/Contact_print) of all the screens on the site, not sure if it'd be useful for the saucelabs tests or not.

k88hudson:
I love the manual tests page, so awesome! I wonder if saucelabs should target that page heh, it would certainly be quicker and you'd get more viewport information. The only problem is I think you wouldn't be able to scroll easily and get useful screenshots... i'll look into it though

k88hudson:
@toolness sorry this was on hold for a bit. I've added a check for react warnings on each page.

Any ideas as to how to automatically add routes here?

toolness:
Hey. I think your last comment was cut off? I just see it abruptly ending at "router/handler paths into"

But I think I get your meaning--we actually do make a list of router paths for the purposes of generating the static site, and it's how we generate the routes on the manual test page too... It's just the `URLS` export of [`lib/routes.jsx`](https://github.com/mozilla/teach.webmaker.org/blob/develop/lib/routes.jsx#L35-L40)... But I have a feeling you might already know that...

As you can see from that snippet, though, we're not extracting the paths out in a particularly robust way, so factoring that out into its own module (or even pushing it upstream into `react-router`) could be rad.

toolness:
Are the sauce tests also running the automated test suite at `/test/`? That would be awesome. Right now tests are failing on IE9 (see #455) and because it's quite a hassle even for us windows users to test on IE9 (I need to power up a VM b/c I can't have multiple versions of IE on my computer at once) it would be great if we could get early warnings on this.

k88hudson:
@toolness not currently but that's a great idea

k88hudson:
@toolness So I used the url list you export -- in the future maybe we could switch to using https://github.com/k88hudson/react-router-viewports/blob/master/src/build-route-tree/build-route-tree.js

I also added the tests page so we can pick up failed tests.

A few remaining issues but i'd rather add follow up tickets to this:

* I can't access logs with the IE webdriver (not implemented), so we'll have to figure out how to detect failed browser tests some other way (e.g. inspecting dom elements etc)
* Firefox on sauce is pretty borked right now -- need to look into why

toolness:
Hey @k88hudson, um, I didn't realize this was an actual legit PR intended for merging because the title is "Testing out some cross-browser testing with sauce". Is this something you think we should resolve the merge conflicts on and merge?

PR #322, opened 2015-03-18
Sidebar Less cleanup

gesa:
Puts us in a better position to understand just how deeply our nesting goes (so for future iterations we can maybe loosen up on the specificity and still accomplish the styles we're looking for)


toolness:
Cool, this looks good to me--thanks aki!

PR #325, opened 2015-03-18
Add a Teach API Client mixin

toolness:
This PR does a number of things:

* Components now access the Teach API via a mixin rather than a global singleton. This is ultimately done via [React Contexts](https://www.tildedave.com/2014/11/15/introduction-to-contexts-in-react-js.html); since we're already using them ourselves as of #312, I figured we might as well use them here too. Among other things, it allows for less error-prone event binding.
* The `stub-router-context` module is now called the `stub-context`, as it stubs out the Teach API too.
* A new `stub-teach-api` module can be used to stub out the Teach API in nifty ways for testing. `login.test.jsx` has been modified to use it.
* The `routes.test.jsx` module is gone now; it was a very fragile test that was a pain to constantly fix, and didn't actually test anything particularly useful. It'll be easier to just monkeypatch the `Link` tag at runtime to log a warning if any Links are ever created that are referred to by path instead of route name.

### DIFF
--- a/components/login.jsx
+++ b/components/login.jsx
@@ -2,31 +2,25 @@ var React = require('react');
 var Router = require('react-router');
 var Link = Router.Link;
 
-var teachAPI = require('../lib/teach-api');
+var TeachAPIClientMixin = require('../mixins/teach-api-client');
 
 var Login = React.createClass({
+  mixins: [TeachAPIClientMixin],
+  statics: {
+    teachAPIEvents: {
+      'login:error': 'handleApiLoginError',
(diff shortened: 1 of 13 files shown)


toolness:
@alicoding can you take a look at this? It looks like a big PR but a lot of it is just renaming stuff...

alicoding:
I really like how we reuse things we already have that suppose to simplify things and using mixin is awesome 👍

R+ as long as Travis is happy!

toolness:
Hooray, thanks Ali!

PR #330, opened 2015-03-19
Implement rudimentary club creation and deletion

toolness:
This implements rudimentary club creation and deletion.

By "rudimentary" I mean that a number of dialogs have been stubbed out with `window.alert()` but communication with the Teach API is done properly and only the owner (creator) of a club has the ability to delete it.

Other notes:

* In general I am trying to make as few components directly use the teach API as possible. so that components are passed in information via `props` and use callbacks etc to delegate state change to their parents. I *think* this is a [best practice](http://facebook.github.io/react/docs/two-way-binding-helpers.html) in React and makes the operation of our app more understandable.
* As such, the `Map` component no longer directly accesses the Teach API, and is instead given a list of clubs by its parent.
* Because Leaflet renders marker popups as static HTML, we can't make those React components. So we need to do some annoying things to integrate UI in the popups.

TODOs:
- [ ] I should really move the `autobind` function out of `teach-api.js` and into its own module, and add tests for it.
- [ ] Even though the UI is in a lot of flux, I should probably at least add a few tests for the Clubs page to ensure that the modals work.
- [ ] Obviously most/all of the `window.alert`s need to be converted to more user-friendly interactions, many of which have already been mocked up by our designers. (But I think there may be some edge cases that haven't been mocked up!)

### DIFF
--- a/components/map.jsx
+++ b/components/map.jsx
@@ -11,6 +11,8 @@ function geoJSONit(data) {
         "type": "Point"
       },
       "properties": {
+        "url": i.url,
+        "owner": i.owner,
         "description": i.description,
         "website": i.website,
         "location": i.location,
@@ -23,11 +25,33 @@ function geoJSONit(data) {
 
 // Note that this class will always be rendered to static markup, so
 // it can't have any dynamic functionality.
+//
+// Furthermore, because we aren't a "live" React element, events
+// dispatched from the map will need to be processed when they bubble
+// up to the parent map component.
 var MarkerPopup = React.createClass({
   getWebsiteDomain: function() {
     return urlParse(this.props.website).hostname;
   },
   render: function() {
+    var actions = null;
+
+    if (this.props.isOwned) {
+      actions = (
+        <div>
+          <button className="btn btn-default btn-xs"
+           data-club-action="edit" data-club-url={this.props.url}>
+            <span className="glyphicon glyphicon-pencil"></span> Edit
+          </button>
+          &nbsp;
+          <button className="btn btn-default btn-xs"
+           data-club-action="delete" data-club-url={this.props.url}>
+            <span className="glyphicon glyphicon-trash"></span> Remove
+          </button>
+        </div>
+      );
+    }
+
     return (
       <div>
         <b>{this.props.title}<br/></b>
@@ -38,6 +62,7 @@ var MarkerPopup = React.createClass({
         <p><a href={this.props.website} target="_blank">
           {this.getWebsiteDomain()}
         </a></p>
+        {actions}
       </div>
     );
   }
@@ -73,6 +98,8 @@ var Map = React.createClass({
       // we have to check if this is a feature or marker-cluster.
       if (feature) {
         html = React.renderToStaticMarkup(React.createElement(MarkerPopup, {
+          isOwned: feature.properties.owner == this.props.username,
+          url: feature.properties.url,
           title: feature.properties.title,
           description: feature.properties.description,
           website: feature.properties.website,
@@ -86,7 +113,7 @@ var Map = React.createClass({
         }));
         marker.bindPopup(html);
       }
-    });
+    }.bind(this));
     this.map.addLayer(this.markers);
     this.updateMap();
   },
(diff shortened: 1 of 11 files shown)


toolness:
Hey @alicoding, can you take a look at this? It's getting to be a bit of a large PR and I don't want it to become too overwhelming. Perhaps you can review it now and I can file the remaining TODOs as issues?

alicoding:
I'm going to review this soon, but lets not add more into this one.

toolness:
Ok sounds good! For the time being, folks can see this deployed on my personal s3 bucket:

http://mozilla-learning-html.s3-website-us-east-1.amazonaws.com/

toolness:
Hey, I'm just adding tests now, no new functionality!

components/map.jsx
```diff
@@ -21,76 +23,143 @@ function geoJSONit(data) {
...
-          });
-        }
-      });
+    });
 
-      // setView of the map based on markers on the map
-      that.map.fitBounds(L.mapbox.featureLayer()
-        .setGeoJSON(geoJSON)
-        .getBounds());
+    this.map.fitBounds([[65, 0], [-65, 0]]);
```
alicoding: Hmm why we are hard coding the coordinate here?

test/browser/teach-api.test.js
```diff
@@ -120,7 +156,82 @@ describe('TeachAPI', function() {
...
+      api.deleteClub('http://myserver/clubs/1');
+      requests.length.should.equal(1);
+      requests[0].method.should.eql('delete');
+      requests[0].url.should.eql('http://myserver/clubs/1');
+    });
+
+    it('returns no error on success', function(done) {
+      api.updateClubs = sinon.spy();
+      api.deleteClub('http://foo', function(err) {
+        should(err).equal(null);
```
alicoding: Heh that url

alicoding:
I have tested and its working as expected except for the location I think we need to work on that a bit since it allows random string and it added to the map but in a weird coordinate. 

I left a comment about the bound position of the map too.

components/map.jsx
toolness: Oh, sorry, I did that because we're no longer actually rendering the map with all the markers known at instantiation time...  Originally the map will be given a `clubs` prop that is an empty array, because we're still asking the Teach API for a list of markers, but once we have some entries, the component will update. That means we have to have some kind of initial viewport, and since there will be clubs around the world, I figured it'd be best to just show the whole map.

Another alternative, I think, is to just set the initial viewport via the mapbox UI, on the `alicoding.ldmhe4f3` map... I *think* that whatever its initial viewport is, will be the initial viewport that the map widget uses if we don't override it like we're doing here.

Um, I think we're also overriding it on line 81, actually--so we should probably get rid of line 92 regardless!

toolness:
> I have tested and its working as expected except for the location I think we need to work on that a bit since it allows random string and it added to the map but in a weird coordinate. 

Ah yeah, I think we'll need to file a separate issue about that and fix it later... We'll probably want to use one of those fancypants as-you-type autocompletion widgets/libraries like they did for the events site here, and directly pass the lat/long from that into the teach API. Right now we're basically just giving the teach API the location string and it's doing the geocoding for us.

That said, I'm not sure the autocomplete is a P1 for our initial launch--since we're giving users the ability to edit their pin, if they notice it's in a wrong part of the world, they'll be able to edit it and tweak the location string a bit, at least. I guess @hannahkane should make the call on that though!

hannahkane:
That's a tough call. Will the user see a warning or error if they enter something that can't be geocoded? That seems necessary for basic usability. Agree with @toolness's assessment that auto-complete can probably come later, though.

alicoding:
Let's file a bug and add that later but this will definitely make it hard to get the marker in the correct coordinate so I would definitely strongly suggest we do this as part of V1 here

toolness:
Hmm, unfortunately I think that autocomplete might actually be the only reasonable solution here--geocoders seem to find *something* for nearly any kind of input, so even if someone types random text or just "springfield", a lat/long *will* be found--it just likely won't be the place that the user is thinking of. The great thing about autocomplete is that it shows the state/country that the location is in, which provides instant feedback that the system is thinking of the same (or a different) place as the user.

That's fine though, it does seem like it's quite important for usability so we can just make autocomplete a P1 for V1.

Anyhow, will definitely file a bug for it!

hannahkane:
Thanks to both of you. I trust your instincts. Autocomplete it is!

PR #331, opened 2015-03-19
Manually set heights for each device and improve display

k88hudson:
I love the manual tests page! It's pretty awesome.

Just a few things here:

* changed the device widths to match up with more common devices
* move the devices object up to the top so it's easier to maintain/add to etc.
* manually set height since it will be more useful for testing actual device sizes, above the fold, etc.
* do actual sizes + scroll x instead of scaling -- i think especially for checking visual details, it's preferable to check the actual size and just scroll horizontally
* show mobile first in the queue since it's the most different/most likely to be overlooked 
* move the labels to the top of the manual test iframes so they're easier to see/read
* add a bit more emphasis to the route links to distinguish the sections

What do you think @toolness @gvn?

### DIFF
--- a/test/browser/static/manual/index.html
+++ b/test/browser/static/manual/index.html
@@ -7,13 +7,35 @@ body {
   padding-top: 70px;
 }
 
+.container-fluid {
+    padding: 0 30px;
+}
+
 .route {
   font-family: monospace;
   font-size: 3rem;
+  margin-top: 6rem;
+  margin-bottom: 2rem;
+  padding: 1rem 0;
 }
 
 .route a {
   color: inherit;
+  background: #FFCD36;
+}
+
+.examples {
+  white-space: nowrap;
+}
+
+.route-thumbnail {
+  display: inline-block;
+  vertical-align: top;
+  margin-right: 30px;
+}
+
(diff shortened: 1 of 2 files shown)


toolness:
Sweeeet! I'm going to take a closer look at this soon but I definitely like the approach!

toolness:
Nice, this looks great @k88hudson!! Thanks for improving the manual tests :)

gvn:
Yeah, I really like this. Very clean and useful!

test/browser/static/manual/index.html
```diff
@@ -7,13 +7,35 @@
...
+}
+
+.route-thumbnail {
+  display: inline-block;
+  vertical-align: top;
+  margin-right: 30px;
+}
+
+iframe {
+    border: 1px solid #CCC;
```
gvn: 2 spaces

test/browser/static/manual/index.html
```diff
@@ -7,13 +7,35 @@
   padding-top: 70px;
 }
 
+.container-fluid {
+    padding: 0 30px;
```
gvn: 2 spaces

Issue #296, opened 2015-03-17
Update footer to reflect latest design


hannahkane:
Design work is here: https://projects.invisionapp.com/share/7X2FVXRN9#/screens

vazquez:
I can take this.

ScottDowne:
This was merged!

PR #336, opened 2015-03-19
updates to footer

ScottDowne:


ScottDowne:
Fixes #296

gesa:
R+

components/footer.jsx
```diff
@@ -98,16 +98,14 @@ var Footer = React.createClass({
...
             </div>
             <div className="col-xs-6">
               <ul className="list-unstyled">
-                <li><a href="">Contact</a></li>
-                <li><a href="">Twitter</a></li>
-                <li><a href="">Facebook</a></li>
-                <li><a href="">Partners</a></li>
+                <li><a href="">Legal</a></li>
+                <li><a href="">Privacy</a></li>
+                <li><a href="/about">About</a></li>
```
toolness: Actually can you make this using react-router's [Link](https://github.com/rackt/react-router/blob/master/docs/api/components/Link.md) component? That's the one that actually triggers in-page transitions, and if you use the name of the route instead of an absolute path, it will also trigger compile-time errors if the route name is invalid (this isn't done w/ absolute paths, which means that it's easier for dead links to creep into our codebase).

gesa: Ahh! Sorry I missed that @toolness, still trying to break non-react habits.

toolness: Ah no worries, it's really not even a react thing, just a `react-router` thing :) I've actually been meaning to somehow document it somewhere, like as part of a quick intro to the project for devs--I guess the README?

ScottDowne: Cool! so that's what Link is doing.

ScottDowne:
Fixed the about link in footer to use Link so it has a transition.

components/footer.jsx
```diff
@@ -98,16 +100,14 @@ var Footer = React.createClass({
               <ul className="list-unstyled">
                 <li><a href="https://webmaker.org/">Webmaker</a></li>
                 <li><a href="https://sendto.mozilla.org/">Donate</a></li>
-                <li><a href="">Legal</a></li>
-                <li><a href="">Privacy</a></li>
+                <li><a href="">Twitter</a></li>
```
gesa: elsewhere the twitter account we're linking to `https://twitter.com/webmaker`

components/footer.jsx
```diff
@@ -98,16 +100,14 @@ var Footer = React.createClass({
...
               </ul>
             </div>
             <div className="col-xs-6">
               <ul className="list-unstyled">
-                <li><a href="">Contact</a></li>
-                <li><a href="">Twitter</a></li>
-                <li><a href="">Facebook</a></li>
-                <li><a href="">Partners</a></li>
+                <li><a href="">Legal</a></li>
```
gesa: we know these will be internal, so they should be Link components that direct to `fixme`

ScottDowne: Privacy and Legal are going to exist internally?

toolness: I believe so, but @hannahkane knows for sure whether privacy and legal link to pages on the teach site vs. links to external sites on other mozilla properties!

hannahkane: I'm actually waiting to hear from moco legal about that. pinged again yesterday. @vasquez did design those pages, so maybe we should just build them. i'll file a ticket to do that.

hannahkane: Actually, wait...it's easier just to link to the Mozilla Legal and Privacy pages, isn't it? Let's do that. They're even more comprehensive, so it's perhaps a better choice all around.

ScottDowne:
Updated the links based on comments.

gesa:
R+++

PR #347, opened 2015-03-19
make icon links a component that shows three links instead of a componen...

ScottDowne:
...t that displays one and needs to be wrapped three times whenever we use it


components/clubs-page.jsx
```diff
@@ -208,7 +166,29 @@ var ClubsPage = React.createClass({
...
+              subhead: "Connect with other Club Leaders"
+            },
+            {
+              linkTo: "fixme",
+              imgSrc: "/img/icon-tips.svg",
+              imgAlt: "icon tips",
+              head: "Helpful Tips",
+              subhead: "Tips for running your Club"
+            }
+          ]}/>
```
toolness: Hmm, I definitely like the direction you're going here, but this code seems to suggest yet another iteration that looks something like this:

```jsx
<IconLinks>
  <IconLink linkTo="fixme" imgSrc="/img/icon-connect.svg"
   imgAlt="icon connect" head="Connect" subhead="Connect with other Club Leaders"/>
  ...
</IconLinks>
```

And so on? It definitely *feels* more React-ish, and it would also allow us to define a more granular `propTypes` on `IconLink` that will let us catch errors and self-document our code a bit better.

What do you think?

toolness: Oh, another advantage of this approach is that I think it would obviate the need to put a `key` prop on every `IconLink`. Which makes sense, since those links are so static.

ScottDowne: I like, I'll update to do that now if nothing more pressing comes up.

ScottDowne: Would this mean that IconLinks would have a dependency on IconLink, and vice versa?

toolness: I don't think that it would be a hard dependency, based on your code, i.e. the code would still run and compile if `IconLink`s were used outside of `IconLinks`--they just might not style particularly well?

Then again, I may not have read your code closely enough yet. I was thinking `IconLinks` would essentially just be sugar for something like `<div className="col-blah myFunkyClass">{this.props.children}</div>`.

ScottDowne: Yeah, that's pretty much the jist of it.

I did make the changes, and I think it's an improvement.

toolness:
Nice, this is awsm, thanks!

Issue #353, opened 2015-03-20
Add autocomplete for the location field on the club form


toolness:
As per the discussion in https://github.com/mozilla/teach.webmaker.org/pull/330#issuecomment-83788403, we need to implement autocomplete for the location field in the club form, perhaps using the [Google Places API](https://developers.google.com/maps/documentation/javascript/places) or something similar.

toolness:
Just wanted to note that on the [server-side](https://github.com/mozilla/teach-api/blob/883716b6e0f7c301abcf9538ccf6e8bff6de1bae/clubs/models.py#L56-L62) we're currently using OpenStreetMap's Nominatim geocoder, which could be cause of the particularly inaccurate results we're getting... The Python geocoding library the server is using supports other geocoders, so we could try different ones too, but I think that using the Google Places autocomplete on the client-side kills multiple birds with one stone.

toolness:
Hmm, just noticed that the [Logo Requirements](https://developers.google.com/maps/documentation/javascript/places#LogoRequirements) for the Google Places Library include:

> If your application displays Places API data on a map, that map must be provided by Google.

*Technically* we do display the data that's ultimately being given to us on a map, but we use Mapbox (which uses OpenStreetMap), not Google. Fortunately, there appears to be a way we can do the autocomplete via Mapbox, which I'm looking into.

toolness:
I think I will try using the ol' [react-select](http://jedwatson.github.io/react-select/) for this. Was going to use [react-autocomplete](https://github.com/rackt/react-autocomplete) but that project seems to be abandoned.

toolness:
Ok, this issue is subsumed by #370 at this point, so I'm closing it.

PR #354, opened 2015-03-20
Implement stepwise/wizard-based add/edit club flow.

toolness:
In #330 we added "rudimentary" club creation, meaning that we use `window.alert` for lots of things. This gets rid of all of those and makes the modal flow much more like the "wizard" style UI from @vazquez's original mockups.

This is the first time I'm attempting to implement a UI using [TDD](http://en.wikipedia.org/wiki/Test-driven_development)--which may or may not be a great idea, but React actually makes it relatively easy to try out. So far I've got the logic of the wizard tested and implemented; now I need to style things properly and add some interactions that enhance usability.

TODO:

- [x] Style buttons properly. Right now they're just `.btn-primary` and look nasty and generic-bootstrappy.
- [x] Add @vazquez's nifty globe-with-marker to the success modal.
- [x] The "waiting for network" step and "an error occurred" step should still show the form. The form should be disabled while waiting for the network, and it should be re-enabled when errors occur, so that the user can try to re-submit it without losing their data.
- [x] The "take me to my club" button calls `this.props.onSuccess`, but the clubs page's implementation of that function doesn't actually do anything yet; it needs to actually show the club on the map.


toolness:
Hey @vazquez, if you want to help style this, here are some quick hacks you can make to `clubs-page.jsx` to make it always load the part of the modal you want, without having to constantly go through the modal flow whenever you make a code change:

```diff
diff --git a/components/clubs-page.jsx b/components/clubs-page.jsx
index c86e220..c3d56b3 100644
--- a/components/clubs-page.jsx
+++ b/components/clubs-page.jsx
@@ -105,7 +105,13 @@ var ModalAddYourClub = React.createClass({
       website: '',
       description: '',
       location: '',
-      step: this.getStepForAuthState(!!this.getTeachAPI().getUsername()),
+
+      // TODO UNCOMMENT THIS BEFORE COMMITTING
+      // (This will auto-set the step to whichever one you want to style)
+      step: this.STEP_SHOW_RESULT,
+      // (This was the original code, restore it before committing)
+      //step: this.getStepForAuthState(!!this.getTeachAPI().getUsername()),
+
       resultURL: null,
       networkError: false
     };
@@ -270,6 +276,10 @@ var ClubsPage = React.createClass({
   },
   componentDidMount: function() {
     this.getTeachAPI().updateClubs();
+
+    // TODO UNCOMMENT THIS BEFORE COMMITTING
+    // (This will auto-open the 'show your club' modal on page load)
+    this.showAddYourClubModal();
   },
   showAddYourClubModal: function() {
     this.showModal(ModalAddYourClub, {
```

toolness:
So I added the "take me to my club" functionality, but because we're using the clustering plugin, there's no easy definitive way to "find" the marker on the map, because those markers don't actually exist until the clustering plugin decides the current zoom level is reasonable enough to distinctly show the marker.

Also, I'm now realizing that if people *only* enter the city/country of their club, we're potentially going to have lots of pins stacked on top of each other--for example, if there are two webmaker clubs in brooklyn, then their pins will both likely be located in the center of brooklyn, directly on top of each other. Which means we need to either have a single marker for multiple webmaker clubs (showing all clubs in the location in the marker's popup), or "fake" things by artifically spreading the clubs a few hundred feet away from each other, or something else.

toolness:
Ok @alicoding, I think this is ready for review!

alicoding:
can you rebase your PR too @toolness ?

toolness:
Done!

I am rather surprised that it didn't result in merge conflicts, but everything seems to be ok.

alicoding:
> Also, I'm now realizing that if people only enter the city/country of their club, we're potentially going to have lots of pins stacked on top of each other--for example, if there are two webmaker clubs in brooklyn, then their pins will both likely be located in the center of brooklyn, directly on top of each other. Which means we need to either have a single marker for multiple webmaker clubs (showing all clubs in the location in the marker's popup), or "fake" things by artifically spreading the clubs a few hundred feet away from each other, or something else.

I've thought about this as well and was wondering maybe we should use Google location autocomplete for pinning markers on the map to prevent this kind of problem. It is actually a one input field for that which doesn't add any addition step to add the club and I think the UX is really obvious for adding the location here. The advantage for this will also help us get the exact coordinate for each markers for each clubs we have.

alicoding:
When editing my club... Wouldn't it make sense to open the popup dialog once the **changed club** information? right now it is set bound of the map to the area of the club, but if there are multiple clubs in the area we're going to see the clusters which it means the user have to click or zoom few times to see their changed club's information.

components/modal.jsx
```diff
@@ -1,10 +1,9 @@
-var React = require('react');
-var ReactAddons = require('react/addons');
+var React = require('react/addons');
```
alicoding: Is this mean we will have access to the normal `React` object plus `addons` mixins at the same time?

pages/clubs.jsx
```diff
@@ -82,63 +84,176 @@ var BottomCTA = React.createClass({
...
+
+    e.preventDefault();
+    this.setState({
+      step: this.STEP_WAIT_FOR_NETWORK,
+      networkError: false
+    });
+    if (this.props.club) {
+      clubState = _.extend({}, this.props.club, clubState);
+
+      // For now, let's always force re-geocoding.
```
alicoding: Lets file an issue to use Google location autocomplete and we always deal with this client-side rather than relying on the API to get the `lat` and `long` for us. IMO it doesn't seem to be reliable at all because we can get random location sometimes.

pages/clubs.jsx
```diff
@@ -82,63 +84,176 @@ var BottomCTA = React.createClass({
...
+            <fieldset>
+              <label>What is the name of your Club?</label>
+              <input type="text" placeholder="We love creative Club names"
+               disabled={isFormDisabled}
+               required
+               valueLink={this.linkState('name')} />
+            </fieldset>
+            <fieldset>
+              <label>Where does it take place?</label>
+              <input type="text" placeholder="Type in a city or a country"
```
alicoding: can we add validation to these field to not only make it `required` but also making sure is not a `space` character? I just enter ` ` and it worked, but this seems to be weird to accept such input though is not an error.

alicoding:
This looking good, functional and tests are passing, but I have bunch of questions that might need some more discussion here. If you want to merge this and file bunch of issues to fix them later I'm okay with that otherwise lets discuss to see if you are agree about what I've said here.

toolness:
Sounds good! Yeah, I'll file issues and we can bust them up individually.

components/modal.jsx
toolness: Yep, it's the equivalent of including their JS bundle that includes all the mixins. The original code that was committed was actually in error--React doesn't complain when a mixin object is `undefined`, which it was, and I'm just fixing things here.

pages/clubs.jsx
toolness: Agreed--this has already been filed as #353. :)

toolness:
Addressing some of your concerns here...

> I've thought about this as well and was wondering maybe we should use Google location autocomplete for pinning markers on the map to prevent this kind of problem. It is actually a one input field for that which doesn't add any addition step to add the club and I think the UX is really obvious for adding the location here. The advantage for this will also help us get the exact coordinate for each markers for each clubs we have.

Can you post those thoughts in #360 and we can discuss there?

> When editing my club... Wouldn't it make sense to open the popup dialog once the **changed club** information? right now it is set bound of the map to the area of the club, but if there are multiple clubs in the area we're going to see the clusters which it means the user have to click or zoom few times to see their changed club's information.

Absolutely! I wanted to do that in this PR, but it's actually kind of nontrivial to do with the clustering plugin, as I mention here: https://github.com/mozilla/teach.webmaker.org/pull/354#issuecomment-84063644

If we *really* want the popup to pop-up when the add/edit modals are closed, I recommend filing a separate issue for it and doing some research to figure out how hard it will be. We should also keep in mind, though, that we'll eventually be adding the "my clubs" sidebar on the left side of the page, which will also make it easier for folks to find their clubs.

PR #367, opened 2015-03-23
Docs for local environment variables

adamlofting:
Just some extra docs about environment variables for local dev.


toolness:
This looks good--sorry `GA_ACCOUNT` wasn't already documented, thanks for adding it.

It needs a section for how to do the env vars thing in Windows but I can add that. Merging now!

PR #370, opened 2015-03-23
Add autocomplete to location field in add/edit clubs dialogs.

### LINKED ISSUE #353: Add autocomplete for the location field on the club form

toolness:
This fixes #353.

TODOs:

- [x] Add tests.
- [x] The widget is technically disabled when the form is being submitted, but it isn't styled to *look* disabled, so we should change this.
- [ ] We're using `react-select`'s default CSS, which is subtly different from our CSS. Might want to get @vazquez to optimize it (we can file this as a separate issue if needed).
- [ ] There isn't an easy way to set `required` on the input field inside the `Select` component, so right now we just do a `window.alert` telling the user to fill out the location field. This is suboptimal, but we can file this as a separate issue if needed.

### DIFF
--- a/components/map.jsx
+++ b/components/map.jsx
@@ -73,9 +73,16 @@ var MarkerPopup = React.createClass({
 
 var Map = React.createClass({
   propTypes: {
-    className: React.PropTypes.string.isRequired
+    className: React.PropTypes.string.isRequired,
+    clubs: React.PropTypes.array.isRequired,
+    username: React.PropTypes.string,
+    onDelete: React.PropTypes.func.isRequired,
+    onEdit: React.PropTypes.func.isRequired
   },
   statics: {
+    setAccessToken: function(value) {
+      accessToken = value;
+    },
     getAutocompleteOptions: function(input, callback) {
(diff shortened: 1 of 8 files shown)


toolness:
@alicoding do you think you have time to review this?

alicoding:
@toolness I will not be able to review them in few hours, but if you can wait I will do them otherwise maybe @ScottDowne ?

toolness:
I'm fine either way! @ScottDowne, if you're interested in reviewing this then you are welcome to assign the ticket to yourself and add the "In Progress" flag to this ticket. If you don't do that, then Ali, whenever you're free, feel free to do the same for yourself... thanks!

ScottDowne:
I'll get to this later tonight :)

ScottDowne:
Works well, tests pass and understanding the code was easy enough :)

PR #373, opened 2015-03-23
Additional GA Event and Modal View Tracking

### LINKED ISSUE #362: Additional Analytics Instrumentation
Now that the site build has progressed, there are further user interactions we want to count / measure in Google Analytics. This ticket is to review the site, list those interactions, and add the additional GA tracking code we require.

adamlofting:
Fixes #362 by adding further analytics tracking events to key points in the UX.

### DIFF
--- a/lib/googleanalytics.js
+++ b/lib/googleanalytics.js
@@ -1,9 +1,12 @@
+function warn (s) {
+  console.warn('[googleanalytics.js', s);
+}
+
 module.exports = {
   initialize: function() {
     if (!process.env.GA_ACCOUNT) {
       return;
     }
-
     // https://developers.google.com/analytics/devguides/collection/analyticsjs/
     (function(i, s, o, g, r, a, m) {
       i['GoogleAnalyticsObject'] = r;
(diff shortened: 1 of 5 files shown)


lib/googleanalytics.js
```diff
@@ -21,15 +24,70 @@ module.exports = {
...
     }
+  },
+
+  /**
+   * modalview:
+   * a proxy to basic GA pageview tracking
+   * @param  {String} modalName e.g. 'add-or-edit-club'
+   */
+  modalview: function (modalName) {
+    path = '/modal/' + modalName;
```
toolness: If we decide to make the modals visible at their own URLs so that e.g. the "add your club" modal is visible at "/mozilla-web-clubs/add/", is this solution compatible with that?

adamlofting: Just to clarify, would that mean there are two ways to the display the form? (ie as a modal, and as a standalone page)?

toolness: Nope, it'd just be the case that `/mozilla-web-clubs/add` would show you the "darkened" page in the background w/ the modal focused in front of it, and anything that currently opens the modal via JS would instead just be a clickable link to `/mozilla-web-clubs/add`. The fancypants router would then trigger the transition to the modal.

If that is really weird and nobody else does it then we don't have to... I just like every "thing" on a page to have a URL I guess. It's also a bit annoying to implement in the current architecture so it wouldn't be implemented in v1.

toolness: Oh, aside from that, it would theoretically allow us to support people actually using those modals on browsers with JS disabled. Additional work beyond just making the modals have their own URLs would need to be done, though. Again though, not sure whether the modals are considered "progressive" or "required" functionality--I'd think at least the "learn more" would be required, but I suppose we could always have that link be a `mailto:` link w/ the progressive enhancement being the modal...

Anyways, that's off-topic for now I guess!

adamlofting: That's interesting, and I've not seen it done before. 

If we make that switch, I think we would have pageview tracking via the router, and we could remove the extra tracking in the modal module.

toolness: Ok, cool--I just wanted to make sure it wouldn't screw up metrics that we track over time, e.g. because what used to be the '/modal/addclub` "route" in ga suddenly changes to `/mozilla-web-clubs/add`.

Thanks!

toolness:
This looks rad! Thanks @adamlofting.

PR #376, opened 2015-03-23
Allow modals to be scrolled through

### LINKED ISSUE #374: Modals can't be scrolled through if they go off-page
Here's an example:

The scroll bar on the side in that screenshot actually scrolls the *background* content, not the modal. There does not appear to be a way for users to scroll the modal itself.

toolness:
This fixes #374.

### DIFF
--- a/components/page.jsx
+++ b/components/page.jsx
@@ -25,6 +25,13 @@ var Page = React.createClass({
   hideModal: function() {
     this.setState({modalClass: null, modalProps: null});
   },
+  componentDidUpdate: function(prevProps, prevState) {
+    if (this.state.modalClass && !prevState.modalClass) {
+      document.body.classList.add('modal-open');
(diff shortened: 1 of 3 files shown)


toolness:
@gesa can you review this?

gesa:
@toolness Absolutely! Could you give me a quick list of steps to test in-browser? (i.e. how to trigger modal)

toolness:
Hooray!

You can just go to http://localhost:8008/mozilla-web-clubs/ and then click on either the "Add Your Club" or "Learn More" links/buttons in the hero unit area. Both of those open dialogs. Then if you shrink your browser window so it's really short (vertically speaking) you should be able to scroll through the modal.

For reference, doing the same thing on the staging site at http://teach.mofostaging.net/ should *not* work--it should only be the background that scrolls, and the modal stays fixed.

gesa:
LGTM! R+

PR #386, opened 2015-03-24
Enable minification for production

### LINKED ISSUE #140: Enable minification for production
In e0ad0ec08fa09e043f9a58efac748ef232bc9b91 I changed things so we *never* minify code instead of minifying it all the time. This solution is suboptimal too, just not at this stage of development, when we're more concerned about developer ergonomics than production efficiencies.

Eventually we do need to build minified code for production, though, which is why I'm filing this issue.

gesa:
Fixes #140 

**The good news:** when i test this locally (basically by adding these tasks back in to the `gulp watch` task) this works great. I'm about 99% certain this will be fabulous on S3—though the only *real* way to test that is to upload it to S3.

**The bad news:** the sourcemaps this creates will be *relatively* worthless. So much so that I seriously considered not even including them.


toolness:
thanks  @gesa!

I was originally thinking we'd just use webpack to minify the JS--its sourcemaps will be as useful as they are now, but with the added bonus of minification. It would require a different strategy than you're employing though--I was thinking that rather than separate gulp tasks for minification, we'd have to pass a [command-line argument into gulp](https://github.com/gulpjs/gulp/blob/master/docs/recipes/pass-arguments-from-cli.md), like `--production`, and then have our webpack config be altered based on that. What do you think?

toolness:
Oh, also, I'm not very familiar with the LESS middleware, but I'm wondering if there's a similar way to tell it to minify as well? That way we could preserve the sourcemap goodness on our CSS as well!

gesa:
Just leaving a note here that on `develop` currently css sourcemaps do not work. So, that just became scope for this PR.

gesa:
Okay! I solved the sourcemaps problem! They now work on local development servers and minified on production! @toolness ready for review.

gesa:
i have NO IDEA why the build is failing. can't reproduce locally =/

toolness:
Oh I've actually run into this before! I believe it's due to https://github.com/plus3network/gulp-less/issues/140...

toolness:
O yea, upgrading to a newer version of `gulp-less` seems to fix things on my end.

```diff
diff --git a/package.json b/package.json
index 30687a7..515b5fe 100644
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "gulp-jscs": "^1.4.0",
     "gulp-jshint": "^1.9.2",
     "gulp-jsmin": "^0.1.4",
-    "gulp-less": "^2.0.3",
+    "gulp-less": "^3.0.2",
     "gulp-minify-css": "^1.0.0",
     "gulp-plumber": "^0.6.6",
     "gulp-prettify": "^0.3.0",
```

toolness:
That said, I am getting this weird "Potentially unhandled rejection" warning when running `npm test`:

gesa:
Yeah I had to downgrade the version of gulp-less in order to get functional sourcemaps…

toolness:
Ok, I've salvaged a good bit of this PR into e6c5ad9b2b85a1e03ecbc92c51fd6dbc3baf15d0 and 71ee7e4cb7237bca0d0250adbe2c74550f3682aa, so I'm closing it.

Issue #375, opened 2015-03-23
Scroll focus moves from page to map as user scrolls down page


toolness:
On desktop/laptop systems, my mousewheel scrolling starts out scrolling the page, but as soon as the cursor is above the map, my mousewheel starts zooming-out the map instead of scrolling the page down.

Here it is in action:

<a href="https://toolness-media.herokuapp.com/v/teach-site-map-scroll-focus/play/loop"><img src="https://toolness-media.herokuapp.com/v/teach-site-map-scroll-focus.poster.play.jpg" alt="Play 'teach-site-map-scroll-focus' video"></a>

There are some UX solutions we can try out to make this less frustrating, but I wanted to document it here.

hannahkane:
The more I use this page, the more of a blocker this seems, so I'm labeling this v1 and putting it into the current milestone. 

Per @iamjessklein, one possible solution is to make the map zooming require a click first.

iamjessklein:
+1 it's presently challenging - although not impossible - to get to the core content.

iamjessklein:
This issue is now being worked on in #387 and we should comment on the fix there.

PR #387, opened 2015-03-24
Fix scrolling issues w/ map

toolness:
This is an attempt to fix the scrolling issues described in #375.

### DIFF
--- a/components/map.jsx
+++ b/components/map.jsx
@@ -162,6 +162,7 @@ var Map = React.createClass({
     }.bind(this));
     this.map.addLayer(this.markers);
     this.updateMap();
+    this.map.scrollWheelZoom.disable();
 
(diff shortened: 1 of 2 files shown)


toolness:
Ok, this PR adds the following behaviors to the map:

* By default, the mousewheel doesn't scroll the map, which means that users can easily scroll through all the page content.
* If the map becomes **focused**, i.e. if the user tabs to it or clicks/pans on it, it will acquire a blue outline. At this point, the mousewheel can be used to zoom in and out.
* If the map loses focus, i.e. if the user tabs to a different widget on the page or clicks outside the map, the map loses its blue outline and the mousewheel cannot be used to zoom in and out.

iamjessklein:
This is a much more friendly solution.

ScottDowne:
Yeah, very nice. R+

PR #389, opened 2015-03-24
Never send auth token to third-party origins.

### LINKED ISSUE #366: TeachAPI should never send auth tokens to third-party origins
We should add a check to the Teach API's `request` method to ensure that it only adds an `Authorization` header if the request is going to the teach API server. Because we're passing potentially-absolute URLs into the request method, if there ends up being a vector for malicious parties to set them, we really don't want the user's auth token accidentally getting leaked.

Actually, a better alternative is simply to disallow the `request` method from sending *any* data to third-party origins, since this should never actually happen in practice.

toolness:
This fixes #366.

### DIFF
--- a/lib/teach-api.js
+++ b/lib/teach-api.js
@@ -91,9 +91,9 @@ _.extend(TeachAPI.prototype, {
       if (url.indexOf(this.baseURL + '/') == 0) {
         req.set('Authorization', 'Token ' + info.token);
       } else {
-        console.warn("Teach API base URL is " + this.baseURL +
-                     " which is at a different origin from " +
-                     url + ". Not sending auth token.");
+        console.warn('Teach API base URL is ' + this.baseURL +
+                     ' which is at a different origin from ' +
+                     url + '. Not sending auth token.');
       }
     }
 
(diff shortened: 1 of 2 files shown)


lib/teach-api.js
```diff
@@ -84,10 +84,17 @@ _.extend(TeachAPI.prototype, {
   },
   request: function(method, path) {
     var info = this.getLoginInfo();
-    var req = request(method, urlResolve(this.baseURL, path));
+    var url = urlResolve(this.baseURL, path);
+    var req = request(method, url);
 
     if (info && info.token) {
-      req.set('Authorization', 'Token ' + info.token);
+      if (url.indexOf(this.baseURL + '/') == 0) {
```
ScottDowne: Another ==

ScottDowne: Could this be changed to `!url.indexOf(this.baseURL + '/')` ?

ScottDowne: Does it need the `+ '/'`?

ScottDowne: Eh, no, the === 0 would be better. string.indexOf logic always makes me do a double look.

ScottDowne:
Just update the == and r+

Issue #337, opened 2015-03-19
CSS fixes


hannahkane:
Reserving time for some clean-up!

- [x] Navigation Bar
- [x] Clubs Curriculum page
- [x] About page
- [ ] Footer
- [x] Web Literacy
- [ ] Map Popups

hannahkane:
@vazquez - wondering if we can close this ticket, or is there more to do that's not covered in other tickets?

vazquez:
We can close :)

iamjessklein:
woot :+1: :postal_horn:

PR #391, opened 2015-03-24
Show multiple clubs in the same popup.

### LINKED ISSUE #360: Only one club in a given location is visible
As mentioned in https://github.com/mozilla/teach.webmaker.org/pull/354#issuecomment-84063644:

> I'm now realizing that if people *only* enter the city/country of their club, we're potentially going to have lots of pins stacked on top of each other--for example, if there are two webmaker clubs in brooklyn, then their pins will both likely be located in the center of brooklyn, directly on top of each other. Which means we need to either have a single marker for multiple webmaker clubs (showing all clubs in the location in the marker's popup), or "fake" things by artifically spreading the clubs a few hundred feet away from each other, or something else.

I tried adding two clubs with the same location name, "Mumbai", in the site and this is indeed the case. The clustering plugin does "cluster" them for as long as it can, which at least shows people that there are 2 clubs there, but once the user is maximally zoomed-in to the map, it only shows one of the markers.

toolness:
This attempts to fix #360 by allowing clubs with the same location/latlong to appear in the same popup.

I'm not happy with the CSS right now, but maybe that can be :boom: **MASSAGED** :boom: during #337?

### DIFF
--- a/components/map.jsx
+++ b/components/map.jsx
@@ -80,6 +80,7 @@ var Map = React.createClass({
     onEdit: React.PropTypes.func.isRequired
   },
   statics: {
+    clubsToGeoJSON: geoJSONit,
     setAccessToken: function(value) {
       accessToken = value;
     },
--- a/components/map.jsx
+++ b/components/map.jsx
@@ -13,25 +13,46 @@ function geoJSONit(data) {
         "type": "Point"
       },
       "properties": {
-        "url": i.url,
-        "owner": i.owner,
-        "description": i.description,
-        "website": i.website,
-        "location": i.location,
-        "title": i.name
+        "clubs": [{
+          "url": i.url,
+          "owner": i.owner,
+          "description": i.description,
+          "website": i.website,
+          "location": i.location,
+          "title": i.name
+        }]
       },
       "type": "Feature"
     }
   });
 }
 
-// Note that this class will always be rendered to static markup, so
-// it can't have any dynamic functionality.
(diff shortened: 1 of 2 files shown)


iamjessklein:
word choice! :-1:  
Agreed with the CSS concerns, particularly if there are like 10 orgs in one location. However, fantastic first step.

toolness:
I recommend we wait until there actually *are* 10 orgs in one place :)

iamjessklein:
Right, I am guessing that this probably isn't a concern for this release. cc/ @hannahkane

hannahkane:
Agree that it's not a priority for initial release. Will be a good problem to have, when it arises. :)

toolness:
Oh, I should also note that there is some repeated information here (the `Mumbai, Maharashtra, India` in the screenshot in the description). We can certainly remove repeated references of that but some design work would help.

iamjessklein:
:+1:  agreed. I'll take a look at it today.

ScottDowne:
Looks good to me, easy to walk through and tests pass.

Can we get a ticket filed for the design fixes for some time in the future?

iamjessklein:
@ScottDowne design issue filed in #420

jgmac1106:
When a club writes a long narrative description it messes up the map navigation. At least for me when I try to read the text box the map pushes down and the text box fades away.

hannahkane:
Filed a separate issue for the bug. Thanks, @jgmac1106!

Issue #379, opened 2015-03-23
import react-a11y in browser


toolness:
at the very least we should import it for unit tests, but doing it for dev builds too would be good. otherwise the react that only gets rendered client-side does not show us accessibility warnings.

hannahkane:
@toolness - is this needed for v1?

toolness:
@hannahkane I wouldn't say so--it's more of a nice-to-have; the `react-a11y` library is experimental anyways and I am running into some technical issues in #398 anyhow, so I'd say we should leave it off v1.

toolness:
Since we're not using `react-a11y` anymore, I'm closing this issue.

PR #398, opened 2015-03-25
[WIP] Import react-a11y during browser-side tests.

toolness:
This is working towards #379. However, it appears the `<Select>` component, provided by `react-select`, is making `react-a11y` very upset, and there doesn't seem to be any easy way to calm it down, so I'm not sure what to do.

### DIFF
--- a/components/modal.jsx
+++ b/components/modal.jsx
@@ -25,6 +25,7 @@ var Modal = React.createClass({
       <div className="modal show"
        role="dialog"
        aria-labelledby="modal-label"
+       tabIndex="-1"
        onClick={this.handleOutsideOfModalClick}>
         <div className="modal-dialog">
           <div className="modal-header">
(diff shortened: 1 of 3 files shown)


ScottDowne:
We now get some messages in the tests that don't look to be fails now.

```
You have a click handler on a non-interactive element but no `role` DOM property. It will be unclear what this element is supposed to do to a screen-reader user. http://www.w3.org/TR/wai-aria/roles#role_definitions
You have a click handler on a non-interactive element but no `tabIndex` DOM property. The element will not be navigable or interactive by keyboard users. http://www.w3.org/TR/wai-aria-practices/#focus_tabindex
You have a click handler on a non-interactive element but no `role` DOM property. It will be unclear what this element is supposed to do to a screen-reader user. http://www.w3.org/TR/wai-aria/roles#role_definitions
You have a click handler on a non-interactive element but no `tabIndex` DOM property. The element will not be navigable or interactive by keyboard users. http://www.w3.org/TR/wai-aria-practices/#focus_tabindex

```

If this is a concern, let's file it and land this.

toolness:
Hmm, I think I'm going to hold off on merging this actually--those warnings are coming from the `<Select>` element, which we can't change (unless we fork or push upstream), and I'd rather have no warnings at all than have a bunch of warnings that we can't do anything about and learn to habitually ignore. I guess I will just close this PR for now and we can revisit it after v1 launch.

PR #403, opened 2015-03-25
Use Router.RefreshLocation instead of Router.HashLocation.

### LINKED ISSUE #232: Figure out how to do web-1.0 style URL fragments with React-router
Since react-router seems to use the URL hash for routing on browsers that don't support pushState, it's unclear how we use the URL hash for its standard purpose, i.e. linking to named anchors/ids in static content. Since our project is an actual website and not a web app, it's quite likely that we'll want to create such links in the near future.

toolness:
This fixes #232. I also know that we could just pass `Router.HistoryLocation` to `Router.run()` and let it gracefully degrade to `Router.RefreshLocation` on its own, but that would make it harder for us to manually test.

### DIFF
--- a/lib/main.jsx
+++ b/lib/main.jsx
@@ -11,9 +11,7 @@ function startRunningSite() {
   if (config.ENABLE_PUSHSTATE) {
     routes.run(Router.HistoryLocation, pageHolder);
   } else {
-    if (!window.location.hash.slice(1))
-      window.location.hash = '#' + url;
-    routes.run(Router.HashLocation, pageHolder);
+    routes.run(Router.RefreshLocation, pageHolder);
   }
 }
 


toolness:
@adamlofting I want you to know this is happening--even though it shouldn't affect the tracking code, since we're using `react-router`'s post-processed URL in our analytics, I think you should know that this means URLs will *actually* be unique now, i.e. the same page will no longer be reachable via both `http://teach/clubs/` and `http://teach/#/clubs/`. This is so that we can use the location hash for its old-fashioned purpose, i.e. navigating to named links/ids on the page.

That said, I guess it *does* mean that we won't be doing any kind of analytics on the location hash--so if we want to do that, we should add code that observes `hashchange` and reports it, taking the current `location.pathname` into account too... Do you think we should do that? I can file a separate issue once this lands, if so.

toolness:
Hmm, it appears that in order for `react-a11y` to be happy, we'll have to add [`role="link"`](http://www.w3.org/TR/wai-aria/roles#link) to any web-1.0 style intra-document links. This is a bit redundant since we're actually just using the `<a>` tag for what it was originally designed for, but it looks like `react-a11y` assumes that location hashes in links are always used to trigger weird JS logic, rather than being used for its original purpose.

With all the workarounds we're doing to make `react-a11y` happy--also see #398 and our [map component](https://github.com/mozilla/teach.webmaker.org/blob/104d25a0ea3de55eff47ebfa798e0ae6ba70c786/components/map.jsx#L166-L172) for other examples--I think we may need to either start pushing upstream changes to `react-a11y` or abandon it entirely in the near future.

adamlofting:
Thanks for the heads up. 

GA by default ignores the # anchors (I guess to simplify the total count of view for a given page), so we don't need to do anything here yet. It could be interesting work on this one day though.

PR #410, opened 2015-03-25
Don't require website field.

### LINKED ISSUE #352: url should not be required field on Clubs form
Some clubs likely won't have a website, so let's make this field optional. 

Also, can we provide a more useful error message for field validation. e.g. "Please enter a web address that starts with http:// or https://"

toolness:
This fixes #352.

**Note:** This is going to have merge conflicts with #391. We should probably review and commit that first, then rebase this PR. That said, the basic code for this feature will remain the same, though we should add tests to ensure that the marker popup works properly when a website for a club isn't present.


ScottDowne:
Works pretty well.

However one nit, an invalid URL gives me a "Please enter a URL" error which makes it sound required. Should we consider changing the error?

ScottDowne:
I'll r+ this because I don't think the error message is critical or a blocker, but would be nice to fix.

toolness:
Ah yeah, that is a separate thing that hannah found too--explanation is in https://github.com/mozilla/teach.webmaker.org/issues/352#issuecomment-84214121. I will file a issue for it!

toolness:
Hmm, I am confused as to why GitHub thinks this can be automatically merged.

toolness:
Ok @ScottDowne, I filed #424 to address your concern about the error message!

PR #414, opened 2015-03-26
Implementing revised version of logo. Fixes #412

### LINKED ISSUE #412: Implement revised version of mozilla wordmark on sidebar.
Aki did as we were told in #399 but we quickly realized it was a [bad idea](https://github.com/mozilla/teach.webmaker.org/issues/401). We decided to go with a [solution](https://github.com/mozilla/teach.webmaker.org/issues/401#issuecomment-86050643) that @jess and @vazquez came up with:

I guess we can use the [white wordmark SVG](https://github.com/toolness/Mozilla-Learning/blob/mozilla-marquee/img/shared-iconography/mozilla_wordmark_white.svg) I hacked in #405 for this (it's the same one Aki committed but with a different `fill`).

Note that I've [backed out](https://github.com/mozilla/teach.webmaker.org/commit/54b7b7e15e87d0a072669e999859f60e7eec829a) #399 for now because it just makes me sad as I work on the site.

vazquez:


alicoding:
r+

Issue #255, opened 2015-03-12
Build "Tips for Running Your Clubs" page


hannahkane:
Design is here: http://invis.io/HR2IW8U98

@vazquez - any assets to add?

vazquez:
No assets to hand over (the line we have already).

The copy is available here: https://docs.google.com/document/d/1syswSeON-0o1svFz8F457uCncg8Kvx5IW8-2Xi3RBpw/edit?usp=sharing

hannahkane:
@ScottDowne - can I assign this one to you?

ScottDowne:
@vazquez What colour orange do we want for this? (I kinda like using off colours of the main pages for child pages)

ScottDowne:
@vazquez I also need some colours on the expander bit.

toolness:
By the way, we want to make sure the path for this page is at `/clubs-toolkit/`... or if it's not, we want to change our [email](https://github.com/mozilla/teach-api/blob/master/clubs/email.py) copy to mention the new URL, because it's the URL @iamjessklein specified in https://github.com/mozilla/teach-api/pull/8.

iamjessklein:
@toolness I just checked the [sitemap](http://invis.io/9G2DK7SR2) and it's actually at /toolkit/ 
I will go ahead and make that adjustment.

iamjessklein:
@toolness or @ScottDowne can you R? this PR https://github.com/mozilla/teach-api/pull/9

ScottDowne:
I thought (and currently have) the url at "teach.mozilla.orb/clubs/toolkit"

iamjessklein:
@scottdowne that's right - I originally had the wrong url in there.

vazquez:
Colors! 

![screenshot 2015-03-27 09 33 38](https://cloud.githubusercontent.com/assets/1119821/6868751/8cce0dc8-d464-11e4-801b-fb1bb282622d.png)

iamjessklein:
:rainbow:

ScottDowne:
Merged, closing.

PR #425, opened 2015-03-26
update bottom cta component and use it in a new stub page for clubs tool...

ScottDowne:
...kit

### DIFF
--- a/components/expander.jsx
+++ b/components/expander.jsx
@@ -33,7 +33,9 @@ var Expander = React.createClass({
             <span className="ion"></span>
           </div>
           <div className="expander-items-container">
-            {this.props.children}
+            <div className="items-margin">
+              {this.props.children}
+            </div>
           </div>
         </div>
       </div>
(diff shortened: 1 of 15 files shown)


ScottDowne:
This just starts the toolkit page and makes the bottom cta bit in clubs something I can reuse.

I just need to make the expander thingy from web-lit something I can reuse and then use it here too.

ScottDowne:
Scratch that, this is much closer to done.

What is left:
- [ ] Figure out color codes for expander menu in this page.
- [ ] Plug the "Add you map to the club" button into something. This might be tricky because it's not on the same page, so we need to wire it up somehow.

ScottDowne:
Ugh, wish travis was faster. I'll fix tomorrow.

vazquez:
:+1:

ScottDowne:
Fixes #255

vazquez:
I opened a PR on your branch with my changes. Looking good man :+1: 

https://github.com/ScottDowne/teach.webmaker.org/pull/3

PR #427, opened 2015-03-26
Scroll map into view when taking users to their club.

### LINKED ISSUE #382: creating a club never scrolled me to the map
A small UX thing, but probably good to fix at some point.

STR:

1. Go to the clubs page
2. Scroll to the bottom
3. Click "add you club to the map"

When done, I see a "take me to my club" button. Clicking it didn't scroll up to the map, thus I wasn't aware it took me to my club.

The "add your club" button at the top of the page likely doesn't have this issue.

toolness:
This fixes #382.

I originally tried using [`smooth-scroll`](https://github.com/cferdinandi/smooth-scroll) to do this, but ran into weird issues, and also got the heebie-jeebies about it conflicting with React (it seems to use `history.pushState`, which is weird to me), so I just resorted to `Element.scrollIntoView()` for now. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView), it's supported by all browsers as far back as IE 6, so we should be safe.

### DIFF
--- a/pages/clubs.jsx
+++ b/pages/clubs.jsx
@@ -431,6 +431,7 @@ var ClubsPage = React.createClass({
     this.showModal(ModalLearnMore);
   },
   handleAddClubSuccess: function(club) {
+    this.refs.map.getDOMNode().scrollIntoView();
     this.refs.map.focusOnClub(club);
   },
   handleClubDelete: function(url, clubName) {


ScottDowne:
Not sure why, but not able to get this to work yet.

toolness:
Huh, it just does nothing for you? What browser are you on? I wonder if `scrollIntoView()` does nothing on some browsers, if the browser thinks "o well a little bit of that element is visible on the page already so i just won't do anything".

ScottDowne:
I was Firefox stable, and I tested the bottom most button too, which shows none of the map.

ScottDowne:
Dug a bit more. I think it's my local setup for teach-api.

Let's not hold this back because of that. I say we move this to staging, because, that's what it's for :D

PR #432, opened 2015-03-27
Remove linkTo on PageEndCTA in clubs.jsx

toolness:
The PageEndCTA component doesn't seem to take a `linkTo` prop, nor does the `BottomCTA` seem to take a `ctaLink` prop. I didn't write the original code though, so I'm submitting this PR for review in case I'm mistaken.


ScottDowne:
I noticed this too, and took a closer look and came to the same conclusion that it wasn't needed.

I removed it in toolkit patch that was using cta so I didn't push it up yet.

PR #433, opened 2015-03-27
Events landing page

### LINKED ISSUE #41: Build Events landing page for site
### RACI
* Phase: Build
* Owner: @hannahkane 
* Decision Maker: @hannahkane 
* Design Lead: NA
* Dev Lead: 
* Quality: @davidascher 

Design is here: http://invis.io/TY2GT8GBX

toolness:
This will fix #41.

I'm regularly publishing my progress so far to http://mozteach.toolness.org/events/.

Notes:

* The 2x version of the hero unit image seems kind of big at 1.5mb. If iPhone 5's are downloading this, that seems kind of excessive if they're being retrieved over a 3G connection. I'm also not really sure if it's even going to make much of a difference from the 1x size on such small devices, but I dunno. I kind of wish there were a way to tell a browser "yo, if this image is going to be on a retina display with a good wifi connection *and* will be shown at a size larger than a postage stamp, then use this higher-res version." (Maybe it's already possible to do this and I just don't know about it?)
* I'm using the `values` class from the site landing page to do the "Join us for the global Maker Party" CTA area right now, because it's so visually similar. `values` doesn't seem very semantically relevant to what the content is, though, so I wonder if we should rename the class, or if we should make a mixin class in LESS, or something else entirely.
* I'm using the `PageEndCTA` component but it's not *technically* at the end of the page--really it's sort of the middle of the page... Dunno if that matters.
* I really dislike the way I styled the "sign up for Maker Party emails" form, but it works for now. Left a TODO in the code about refactoring it later.
* For some reason the email field in that form doesn't work on Firefox 36.0.4 and I can't figure out why--seems to have something to do with the CSS on it, as disabling that makes the field work okay. I think it's OK for now b/c that form doesn't actually do anything yet anyways.
* I tried using `react-slick` for the carousel but it was terrible at responsively scaling itself to its container width. I used this as an opportunity to try `react-bootstrap` and its carousel works much better.
* However, `react-bootstrap`'s carousel makes `react-a11y` unhappy, so I had to disable its warnings. Coupled with the issues I ran into in #398 and other [workarounds](https://github.com/mozilla/teach.webmaker.org/blob/39e66bfd6f49ec1191e886cce2257edbed50d13a/components/map.jsx#L212-L218) we've had to do to appease `react-a11y`, I think this is for the best. We should probably just wait for that library to mature (and/or help it mature) before we start using it again.

TODOs:

Note that any of these can be filed as separate issues after landing this PR, if we want to take a more iterative approach.

- [X] ~~I'm not sure what the actual "color" for this page is: the sidebar for this page in the `develop` branch is red, while Ricardo's [invision mockup](https://projects.invisionapp.com/share/TY2GT8GBX#/screens) has a blue sidebar and blue-shaded line art on the page. In this PR, the sidebar is currently red while the on-page line art is blue-shaded. Which one "wins"? (Changing the line art won't be hard, if we need to, I just need the new fill color and can edit the SVGs by hand.)~~ The sidebar is blue.
- [ ] I've changed the sidebar color to blue, but I didn't get rid of the `@redSidebarColor` variable that used to define the events page sidebar color, because `less/pages/about.less` seems to be using it for a `.list-with-illust` rule. However, I'm not sure if *that* page is supposed to be using that color or not... Might be worth looking into, albeit sort of out of the scope of this issue.
- [x] Style the email sign-up form.
- [ ] Make the email sign-up form POST to https://sendto.mozilla.org/page/s/maker-party-signup like the one at the bottom of https://party.webmaker.org/ does.
- [x] Use ~~[react-slick](https://github.com/akiran/react-slick)~~ [react-bootstrap](http://react-bootstrap.github.io/) to build the carousel with the appropriate [photos](https://github.com/mozilla/teach.webmaker.org/issues/41#issuecomment-86774658).


iamjessklein:
@toolness yes it's the blue as per the invision mockup - the red is off-brand and so 5 minutes ago. :blue_heart: :large_blue_circle: :large_blue_diamond:

iamjessklein:
@vazquez will need to provide the specific fill color info.

toolness:
Thanks! Based on my eyedropper tool on his mockup, the base sidebar colour seems to be `#2D4A8C`.

vazquez:
Correct @toolness! The colour is `#2D4A8C`

toolness:
Ok, I think this is probably good for review now--should probably just commit this and file issues for the remaining TODOs. @ScottDowne can you take a look?

ScottDowne:
R+ but we have merge conflicts, nothing critical.

toolness:
o snap

PR #444, opened 2015-03-27
flexbox for illustration and fix arrow position

ScottDowne:


vazquez:
I found the reason why we're experiencing weird gaps in Firefox 38. Let's tackle this in person.

vazquez:
Narrowed it down to one problem. It has to do with pixel density when the component is querying for image size.

ScottDowne:
I figured it out.

tdldr is we need to, and should be setting explicit sizes for all images.

The long story:
srcset attribute, which is deciding which image size to display is bugged in firefox dev and nightly. The bug is what causes the width to increase.

The bug goes away with explicit sizes set. This is also a good practice and a huge performance boost. There is no reason we shouldn't be doing it.

The reason it's only in dev and nightly: srcset was just implemented and currently doesn't exist on stable. Right now our stable just shows large images because that's the default behavior. Pretty terrible.

If we want to not be too large on devices that don't support srcset we should be using the 1x.

I didn't go through and update all images to use explict height and widths, I just did it to the ones that are problems, we can do the rest in another patch.

@vazquez review?

vazquez:
YESS!!!! :boar: :boom: :scream: :rabbit: 

A well-deserved R+. Looking great everywhere.

vazquez:
Merging.

Issue #324, opened 2015-03-18
Update icon links in both the club page and teach like mozilla page


ScottDowne:
Both the pages, club and teach like mozilla have icons and the bottom that are links. Right now that links are "fixme"

I'm not sure where we want these to go yet, but they are as follows:

- teach like mozilla
  - listen
  - attend
  - say hello
- clubs
  - curriculum
  - connect
  - helpful tips

hannahkane:
ah, good questions. :)

- teach like mozilla
 - listen: http://mzl.la/TTWpodcasts
 - attend: https://wiki.mozilla.org/MozillaWiki:Teach_the_Web_Talks
 - say hello: http://discourse.webmaker.org/category/meet
- Clubs
 - curriculum: the Clubs Curriculum page that @gesa is making in #42
 - connect: http://discourse.webmaker.org/category/meet
 - helpful tips: the Tips page that's currently in the design queue (#254)

gesa:
Thanks @hannahkane!

qq: which email address does this link to? ![](http://snaps.akibraun.com/rs9ki.png)

hannahkane:
Help@webmaker.org

ScottDowne:
Because some of these pages don't exist yet, we should just leave this open and do it all then.

ScottDowne:
Looks like the events page is also missing links in the bottom three link icons. @hannahkane

hannahkane:
Good catch.
 * Event Resources should go to the page being built in #388. Assuming the url will be /events/resources
 * Join the Conversation should go to http://discourse.webmaker.org/category/maker-party
 * Add Your Event should go to https://events.webmaker.org/

ScottDowne:
@hannahkane Looks like I need that links for bottom of the web lit page too (I may have missed it somewhere). Updating events page now.

hannahkane:
Web Lit page links:
 * Contribute: https://wiki.mozilla.org/Webmaker/WebLiteracyMap
 * Learn More: https://mozilla.github.io/webmaker-whitepaper/
 * Get Reading: https://www.mozilla.org/en-US/about/manifesto/

[Please note: there's an error in the Get Reading copy. That should say "Web Literacy is core to the Mozilla mission and values", rather than "Webmaker is core..."]

ScottDowne:
Should be live now, closing.

PR #450, opened 2015-03-29
Filling in some links, and making the IconLink component handle both Lin...

ScottDowne:
...k routes and simple hrefs

### DIFF
--- a/components/icon-link.jsx
+++ b/components/icon-link.jsx
@@ -1,11 +1,31 @@
 var React = require('react');
+var Router = require('react-router');
+var Link = Router.Link;
+
+var LinkSwap = React.createClass({
+  render: function() {
+    // Swap out Link or a simple anchor depending on the props we have.
+    if (this.props.linkTo) {
+      return (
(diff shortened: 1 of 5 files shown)


ScottDowne:
Fixes #324

components/icon-link.jsx
```diff
@@ -1,11 +1,31 @@
...
+        </Link>
+      )
+    }
+    return (
+      <a href={this.props.href}>
+        {this.props.children}
+      </a>
+    )
+  }
+});
```
vazquez: This is great.

vazquez:
R+ I really like the flexibility of this component.

The only comment I would have is regarding the component name. Would something like `LinkHandler` make more sense to describe this component? Anyway, just an opinion. Otherwise, awesome :star:

ScottDowne:
@vazquez I think a LinkHandler would be a nice addition, but it would be a new thing separated out of IconLink which does a bit more than just create a link. I say file a new issue to create a LinkHandler which then boils down IconLinks a bit.

PR #454, opened 2015-03-30
Move analytics to a module

adamlofting:
Hey @toolness - I moved this analytics code out to an npm module https://github.com/adamlofting/react-ga/ so we can use the same logic in other projects. 

The module itself could do with test coverage, but I'll follow up with you separately on that task.


lib/routes.jsx
```diff
@@ -50,7 +51,8 @@ exports.generateStatic = function(url, cb) {
 };
 
 exports.run = function(location, el) {
-  ga.initialize();
+  var options = { debug: false };
```
toolness: We can actually set debug to `process.env.NODE_ENV != 'production'` here, yay!

toolness:
This looks good to me! You can change the one suggestion I made or just push it, up to you.

lib/routes.jsx
toolness: Oh wait, that debug thing is quite verbose. Perhaps we should just add a `GA_DEBUG` env var, and if it is `on` then we set that flag to true.

PR #465, opened 2015-03-31
Bump reactga to fix bug, and add a debug flag

adamlofting:
Bumps the version of `reactga` to fix a bug, and adds and .env variable to enable devs to switch on debug info for reactga if they wish.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ string), the boolean is true; otherwise, it's false.
 `MAPBOX_MAP_ID` | is the Mapbox map ID to use for embedded maps in the website. Optional.
 `TEACH_API_URL` | is the origin of the [Teach API][] server. Defaults to `https://teach-api.herokuapp.com`.
 `GA_ACCOUNT` | is the property ID of the Google Analytics account. E.g. `UA-123...`. It defaults to the property ID for the Teach site. Set it to `DISABLED` to disable Google Analytics entirely.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -157,6 +157,7 @@ string), the boolean is true; otherwise, it's false.
 `MAPBOX_MAP_ID` | is the Mapbox map ID to use for embedded maps in the website. Optional.
 `TEACH_API_URL` | is the origin of the [Teach API][] server. Defaults to `https://teach-api.herokuapp.com`.
 `GA_ACCOUNT` | is the property ID of the Google Analytics account. E.g. `UA-123...`. It defaults to the property ID for the Teach site. Set it to `DISABLED` to disable Google Analytics entirely.
+`GA_DEBUG` | When set to true will output verbose info to the console about what data is being sent to Google Analytics.
```
toolness: Actually, booleans are notoriously confusing when defined in environment variables, since env vars are strings and don't easily convert to booleans. For instance, reading this docstring, one might think that setting the env var to "false" would ensure that GA is disabled, but that's actually not the case at all, looking at the actual code.

As such, I recommend just specifying an *exact* value to enable GA debugging. Since we're using `off` to disable `LESS_AUTOPREFIXER`, how about specifying `on` to force-enable Google Analytics debugging, with any other value disabling it?

adamlofting: Good call. I'll amend and rebase.

adamlofting:
@toolness I've changed the .env variable from boolean to string.

PR #466, opened 2015-03-31
force width and height on illustration components and re add pixel densi...

ScottDowne:
...ty check because it's still needed for retina and browsers without support for srcset

### DIFF
--- a/less/pages/web-literacy.less
+++ b/less/pages/web-literacy.less
@@ -1,7 +1,4 @@
 .web-literacy {
-  section {
-    padding: 0px;
-  }
   .web-map {
     text-align: center;
     .web-map-contents {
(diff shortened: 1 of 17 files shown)


less/pages/web-literacy.less
```diff
@@ -1,24 +1,6 @@
...
-        font-size: 3rem;
-      }
-    }
-    @media screen and (min-width: @screen-teensy) {
-      h2 {
-        font-size: 2.5rem;
-      }
-    }
+  section {
+    padding: 0px;
```
vazquez: Let's give section a padding of `2.5rem 0 0 2.5rem`

vazquez:
R+

ScottDowne:
Updated, I found another issue but filed it instead: https://github.com/mozilla/teach.webmaker.org/issues/467

Issue #423, opened 2015-03-26
Mobile menu cuts off sub header


ScottDowne:
![mobile](https://cloud.githubusercontent.com/assets/197334/6857553/dc63191e-d3dd-11e4-83e4-02d45a44605b.png)

You can see clubs-curriculum is cut off.

ScottDowne:
Merged, closing this.

Issue #510, opened 2015-04-01
MOBILE: In navigation the Clubs Toolkit is hidden


iamjessklein:
When Mozilla web clubs is selected, the child section Clubs Toolkit is dissected and does not fully show.

sabrinang:
![screenshot_navigation_clubs-toolkit](https://cloud.githubusercontent.com/assets/2492510/6946495/83415b62-d855-11e4-898a-9286ceca57fe.png)

jamieleung:
It appears adding `height: 100%` to `.sidebar` could fix this issue, shall I go ahead and apply this fix?
![screen shot 2015-04-01 at 19 31 36](https://cloud.githubusercontent.com/assets/285441/6948458/db5c2a3e-d8a5-11e4-866f-2c1d8e33073b.png)

iamjessklein:
Yeah that would be rad @jamieleung - we will check out the PR. :+1:  Thanks

toolness:
Er, isn't this a duplicate of #423?

iamjessklein:
indeed it is @toolness @ScottDowne

toolness:
Ok, I guess I will close it as a DUPE then!

PR #521, opened 2015-04-01
fix bug in sidebar display cutting off sub items by reducing complexity ...

ScottDowne:
...in sidebar

### DIFF
--- a/less/components/sidebar.less
+++ b/less/components/sidebar.less
@@ -51,7 +51,7 @@ html.no-js .sidebar {
   transition: background-color @sidebarTransitionTime;
   color: white;
   font-size: 1.4rem;
-
+  padding: 0 0 1.5rem 0;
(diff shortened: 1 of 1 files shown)


ScottDowne:
Fixes #423

iamjessklein:
In this scenario - that there are 2 PRs that deal with the same issue (#510) - what is the protocol?

toolness:
Looks good to me, thanks scott!

@iamjessklein, we basically figure out the resolution on a case-by-case basis. In this case, there was [consensus}(https://github.com/mozilla/teach.webmaker.org/pull/537#issuecomment-89320579) that the other PR increased code complexity to solve the problem (which makes a codebase harder to understand and maintain/change) while this one reduces complexity to solve the same problem, so we're going with this one!

PR #535, opened 2015-04-01
Fix #512

### LINKED ISSUE #512: MOBILE: Image of globe should be centered
center globe image

iamjessklein:
I changed aligned the globe so that it’s centered on mobile.

### DIFF
--- a/less/components/activities-page.less
+++ b/less/components/activities-page.less
@@ -67,7 +67,7 @@
   }
   .illustration {
     .image-container {
-      text-align: left;
+      text-align: center;
       @media (min-width: @screen-sm-min) {
         text-align: center;
       }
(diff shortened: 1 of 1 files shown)


iamjessklein:
@ScottDowne can you take a look?

ScottDowne:
This also centers all the activity kit thumbnails.

ScottDowne:
Good opportunity here to show a bit about how react components work.

The Illustration component already has center as the default text-align: https://github.com/mozilla/teach.webmaker.org/blob/develop/less/components/illustration.less#L13

So you could get the same thing in your patch by removing the `text-align: left;` vs changing it to `text-align: center;`

iamjessklein:
R? @ScottDowne

ScottDowne:
Moved to #542

PR #551, opened 2015-04-02
Sidebar content on mobile collapses when URL path changes.

### LINKED ISSUE #509: MOBILE: The navigation should collapse (it remains open)
After selecting a menu link the navigation should collapse (it remains open)

toolness:
This fixes #509. I didn't implement an `onClick` handler because that would also require implementing keydown handlers for accessibility and all that junk, which would additionally mean lots of unit tests.

I also figured that what we *really* want is to have the sidebar collapse on mobile whenever the current URL changes, not necessarily just when users click on a link in the sidebar. Since that was also easier to implement and test, I went with that approach.

### DIFF
--- a/components/sidebar.jsx
+++ b/components/sidebar.jsx
@@ -54,10 +54,20 @@ var Sidebar = React.createClass({
       ]
     }
   ],
+  propTypes: {
+    sidebar: React.PropTypes.string
+  },
   getInitialState: function() {
     return {
       showCollapsibleContent: false
     };
+  },
+  componentWillReceiveProps: function(nextProps) {
+    if (this.props.pathname !== nextProps.pathname) {
+      this.setState({
+        showCollapsibleContent: false
+      });
+    }
   },
   handleHamburgerClick: function() {
     this.setState({
--- a/components/page.jsx
+++ b/components/page.jsx
@@ -67,7 +67,7 @@ var Page = React.createClass({
          onFocus={this.state.modalClass && this.handleNonModalFocus}>
           {DevRibbon ? <DevRibbon/> : null}
           <div className="row">
-            <Sidebar/>
+            <Sidebar pathname={this.getPathname()} />
             <main className="content col-md-9">
               <RouteHandler/>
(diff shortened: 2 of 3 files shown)


components/sidebar.jsx
```diff
@@ -54,11 +54,21 @@ var Sidebar = React.createClass({
...
+  propTypes: {
+    sidebar: React.PropTypes.string
+  },
   getInitialState: function() {
     return {
       showCollapsibleContent: false
     };
   },
+  componentWillReceiveProps: function(nextProps) {
+    if (this.props.pathname !== nextProps.pathname) {
```
toolness: Check it out, I remembered a `!==` for once!

ScottDowne:
This makes it harder to navigate to sub pages.

I need to click on a page to see any of that page's sub pages.

If the menu is closed on clicking a new page, it means I need to open it again to see those sub pages.

This is actually more of a design question, but one that is exposed by your patch.

I am fine with landing this and filing a design issue, but if you have any ideas you want to try now that's fine too.

toolness:
> This makes it harder to navigate to sub pages.
>
> I need to click on a page to see any of that page's sub pages.
>
> If the menu is closed on clicking a new page, it means I need to open it again to see those sub pages.

Ahh great catch! @iamjessklein what do you think about this? IMO it is enough of a potential downside to the UX implications of this patch that I'd like her opinion before merging anything.

ScottDowne:
@toolness Cool. One option is we show all sub menus when on mobile.

iamjessklein:
I think that we should have the default on mobile be dropdown.

My main concern here is real estate, but let's go with this direction and iterate after we test it.

ScottDowne:
@iamjessklein Are you saying the sub menu items are expanded when on mobile? If so, I agree and we should do that.

toolness:
Hmm this PR isn't really going anywhere and it doesn't seem like anyone is super bothered with the way the navbar is currently working, so I'm gonna close this PR.

PR #553, opened 2015-04-02
Events Page - copy edits

iamjessklein:
*  Under Join The Global Movement, should "learning about remix using
paper and scissors" be "learning about * Under the Maker Party logo,
the sentence says "global" twice. Take out "the global" in the beginning


iamjessklein:
R? @ScottDowne :eyeglasses:

PR #567, opened 2015-04-02
Show add club modal on page load if ?modal=add is in url on clubs page.

### LINKED ISSUE #566: Make the "Add your Club" modal accessible via a URL
If the user acceses `?modal=add`, we should pop open the ol' "Add Your Club" modal at initial page load.

We need to do this in order to give the teach API a callback URL to go back to once the user has completed the login/signup flow for https://github.com/mozilla/id.webmaker.org/issues/51.

toolness:
This fixes #566.

Note that it doesn't remove `?modal=add` when the modal is dismissed. This is partly because it's a hassle to implement and test, but also partly because I'd like to implement a more robust solution for giving every modal its own URL in v2, which I allude to a bit in https://github.com/mozilla/teach.webmaker.org/pull/373#discussion_r26956659 (this will make it a lot easier to allow older/non-JS browsers to use the modals too, among other things). As such, this PR is really just a quick fix to unblock https://github.com/mozilla/id.webmaker.org/issues/51.

### DIFF
--- a/pages/clubs.jsx
+++ b/pages/clubs.jsx
@@ -462,7 +462,7 @@ var ModalLearnMore = React.createClass({
 
 
 var ClubsPage = React.createClass({
-  mixins: [ModalManagerMixin, TeachAPIClientMixin],
+  mixins: [ModalManagerMixin, TeachAPIClientMixin, Router.State],
   statics: {
     ClubList: ClubList,
     ModalAddOrChangeYourClub: ModalAddOrChangeYourClub,
@@ -476,6 +476,10 @@ var ClubsPage = React.createClass({
   },
   componentDidMount: function() {
     this.getTeachAPI().updateClubs();
+
+    if (this.getQuery().modal === 'add') {
+      this.showAddYourClubModal();
+    }
   },
   showAddYourClubModal: function() {
     this.showModal(ModalAddOrChangeYourClub, {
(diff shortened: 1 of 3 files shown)


ScottDowne:
R+ :white_check_mark: :white_check_mark: :white_check_mark:

We could use this for the "add club" link at the bottom of the club toolkit page.

I gave it a quick look but it looks like it needs to use the `Link` component and that's not a simple `linkTo += "?modal=add"`so I say punt it to a new issue.

toolness:
> I gave it a quick look but it looks like it needs to use the `Link` component and that's not a simple `linkTo += "?modal=add"`so I say punt it to a new issue.

Yeah, in v2 I'd actually like to change all modals have their own routes, e.g. `/clubs/add` for the clubs modal; I explain this a bit more in https://github.com/mozilla/teach.webmaker.org/pull/373#discussion_r26956659 but the gist is that it'll make react-router do more of the heavy lifting for us, it'll make the code more "react-esque" and easy to reason about, and it'll make it possible to make the modals accessible for folks without JS.

For now though I think it's best to keep `?modal=add`'s usage limited to a "hack" that allows our login flow to work--for instance, just turning the "add your club" into a `<Link>` won't actually work with this PR because that JS is only triggered in `componentDidMount`--meaning that clicking on the Link won't actually trigger the modal, because the component is already mounted! And I don't think that'd be the last of our problems, either... Giving every modal its own route in v2 will totally fix this, though.

PR #570, opened 2015-04-03
Geocoding is no longer overly specific and lame.

### LINKED ISSUE #526: CLUBS: Location too specific?
The list only lets me select from either M4H or M5C postal codes -- which are two very specific locations in Toronto. Should I be able to have a generic Toronto located club? 

screen shot 2015-04-01 at 15 25 58

Toronto M4H:

screen shot 2015-04-01 at 15 30 20

toolness:
This fixes #526.

Specifically, it resolves the following problematic situations:
* Showing the zip/postal code, which is overly specific and confusing. This is now hidden from view.
* Showing both "Nairobi, Nairobi, Kenya" (the city) and "Nairobi, Kenya" (the region/county). Only the city is shown now.
* Showing both "Toronto, Ontario, Canada" (the city) and "Toronto St, Toronto, Ontario, Canada" (the address). Only the city is shown now.

Here's an animated GIF of some of those queries in action with this PR:

geocoding

I should note that my fix relies a lot on the specific format of the GeoJSON that Mapbox's "public beta" geocoding API returns, but it adheres to the guidelines listed under the "Response format" section of their [geocoding API documentation](https://www.mapbox.com/developers/api/geocoding/). Specifically, we're relying on the formatting of feature IDs and feature `text`, `center`, and `context`, all of which are documented as being stable.

### DIFF
--- a/components/map.jsx
+++ b/components/map.jsx
@@ -133,6 +133,24 @@ var Map = React.createClass({
     setAccessToken: function(value) {
       accessToken = value;
     },
+    simplifyMapboxGeoJSON: function(features) {
+      return features.filter(function(feature) {
+        return /^place/.test(feature.id);
+      }).map(function(feature) {
+        var context = feature.context.filter(function(item) {
+          return /^(region|country)/.test(item.id);
+        }).map(function(item) {
+          return item.text;
+        });
+        var info = {
+          location: [feature.text].concat(context).join(', '),
+          latitude: feature.center[1],
+          longitude: feature.center[0]
+        };
+
+        return info;
+      });
+    },
     getAutocompleteOptions: function(input, callback) {
       var url = 'http://api.tiles.mapbox.com/v4/geocode/mapbox.places/' +
                 encodeURIComponent(input) +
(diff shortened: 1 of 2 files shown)


toolness:
Hmm, one funkiness I'm finding is that there are supposedly [at least 6 Springfields in Wisconsin](http://en.wikipedia.org/wiki/Springfield,_Wisconsin) but none of them are showing up in our geocoding results. This isn't really a bug in our front-end--[Mapbox's GeoJSON response for `springfield wi`](https://pastebin.mozilla.org/8828281) is actually only returning street addresses, not cities--and none of those street addresses are in cities called Springfield. So it's like Mapbox doesn't think there are *any* Springfields in Wisconsin, which is odd.

I guess I recommend we just go with this PR as-is, and file any future bugs that come up from QA parties and user feedback.

iamjessklein:
Nice! Also love the use of screencapture here.

hannahkane:
Nice. +1 on your plan for iterating based on specific edge cases identified by users.

iamjessklein:
:neckbeard: best pr name ever

ScottDowne:
R+

Issue #573, opened 2015-04-03
Show warning/upgrade message if major functionality is disabled for user's browser


toolness:
Currently we have a suboptimal list of minimum browser requirements for core functionality in the site; as mentioned in #572, we'll be disabling JS entirely for now if the user's browser doesn't meet them. Unfortunately, this means the loss of some rather core functionality for v1:

* Login won't work (and therefore adding/removing clubs won't work).
* The map on the clubs page won't work.
* The clubs list on the clubs page won't be shown (we can fix this for v1 if absolutely necessary, but it'll be a lot easier to fix for v2 due to [changes in architecture](https://github.com/mozilla/teach.webmaker.org/issues/585)).
* The carousel on the events landing page will likely be permanently stuck on the first image (hopefully not too big a deal).
* ~~It may be impossible to un-collapse the skills in the web literacy page and the sections of the clubs toolkit page (I'm unfamiliar with that code so I'm not sure; if needed, we could potentially have those sections always be un-collapsed when JS is disabled, the same way we currently do for the sidebar, but we'll need to file a separate issue for it).~~ Whoa I think a `:focus` rule is making it so those collapsible sections un-collapse when clicked/focused without requiring any JS, which is [brilliant](https://github.com/mozilla/teach.webmaker.org/commit/f0981c96352f2da5031bd865f38f70d50a17b8b7#commitcomment-10569879).

Some stuff will still work, but in a "gracefully degraded" way:

* Clicking on the tabzilla tab will take the user to mozilla.org instead of sliding down the tabzilla tab.
* Moving to different pages on the site will force full page refreshes and won't show cool transition effects.
* The sidebar will always be un-collapsed on mobile.

As the site evolves, more of the items in the first list ("doesn't work") will move to the second list ("gracefully degrades"). But in the meantime, we should show a warning to users letting them know that some parts of the site won't work for them, with links to information on upgrading their browser.

I don't know exactly how this warning should be worded, but we should probably keep in mind the fact that--at least in our experience with other Webmaker products--most folks on older browsers aren't actually in control of their browser choice, e.g. because they're at a library or workplace.

toolness:
Oh, I should add that we might want to simply hide some of the interactive elements on the page that require JS when JS is disabled. This can be done via CSS rules that start with `html.no-js`.

The warning can also be displayed via CSS rules that start with `html.no-js`.

hannahkane:
@toolness - do you need a UI for this?

PR #579, opened 2015-04-03
Add a 'safe mode' to the site.

### LINKED ISSUE #572: Fallback to non-JS version if feature testing indicates a minimum JS/HTML5 feature set isn't available
We should use something like [modernizr](http://modernizr.com/) to feature detect for the bare minimum feature set for the JS version of the site to work. For the time being, that means things like:

* Proper [CORS XMLHttpRequest](http://caniuse.com/#search=cors) support (sorry IE9!)
* ... Other stuff?

If any of these criteria aren't met, we should simply not load any of our JS bundles, nor should we remove the `no-js` class from the `<html>` element during page load. This should effectively be the same thing as disabling JS in the browser, which we have decent support for.

Note that our list of minimum feature set requirements isn't fixed--hopefully it might even *shrink* as time goes on rather than grow, as we start using more polyfills and other workarounds (e.g. [JSONP](http://en.wikipedia.org/wiki/JSONP) or [PPX](https://github.com/toolness/postmessage-proxied-xhr) as a substitute for CORS requests).

toolness:
This fixes #572 by simply disabling JS if the browser doesn't support a minimal feature set of supported JS APIs (at present, this means CORS). It also allows developers to force-enable this "safe mode" by adding `safemode=on` to the querystring of any page.

Note that I did look into [Modernizr](http://modernizr.com/) for this but because our detection is so simplistic, I found it easier to just borrow a snippet of code from them rather than create a custom build of Modernizr outright. That said, as I mention in the code comments, if our feature detection becomes more complex, we should probably consider migrating to Modernizr.

### DIFF
--- a/lib/index-static.jsx
+++ b/lib/index-static.jsx
@@ -2,6 +2,27 @@ var React = require('react');
 
 var routes = require('./routes.jsx');
 
+// This isn't actually called in node, it's stringified and plopped in
+// a script tag in the page header. It's basically an extremely simple
+// stand-in for Modernizr, but if it becomes more complex we should think
+// about actually migrating to that library.
+//
+// Modernizr code borrowed:
+//
+// * cors (needed to reach teach-api)
+function featureDetect() {
+  var safeMode = /[?&]safemode=on/i.test(window.location.search);
+  var cors = 'XMLHttpRequest' in window &&
+             'withCredentials' in new XMLHttpRequest();
+
+  if (!safeMode && cors) {
+    document.documentElement.setAttribute('class', '');
+    window.ENABLE_JS = true;
+  } else {
+    window.ENABLE_JS = false;
+  }
+}
+
 function generateWithPageHTML(url, options, pageHTML) {
   options = options || {};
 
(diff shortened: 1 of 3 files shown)


lib/index-static.jsx
```diff
@@ -2,6 +2,27 @@ var React = require('react');
...
+// stand-in for Modernizr, but if it becomes more complex we should think
+// about actually migrating to that library.
+//
+// Modernizr code borrowed:
+//
+// * cors (needed to reach teach-api)
+function featureDetect() {
+  var safeMode = /[?&]safemode=on/i.test(window.location.search);
+  var cors = 'XMLHttpRequest' in window &&
+             'withCredentials' in new XMLHttpRequest();
```
toolness: Note that it's fine if this code throws an exception--that essentially means that we *definitely* want to disable JS, and because `ENABLE_JS` won't be set to `true` at the end of the function, we'll be in safe mode.

ScottDowne:
R+

Something I noticed, the list of clubs doesn't show with js disabled. A bug?

toolness:
Thanks!

> Something I noticed, the list of clubs doesn't show with js disabled. A bug?

Yeah, I've documented this in #573 (see the third bullet point). Because that content is dynamic and changes w/ user content submissions, it's very hard to make it part of a static site that doesn't use JS; we could for instance trigger a regular cron job that regenerates the static site based on the latest clubs data, which might be a feasible solution if the clubs were the only user-generated content on the site planned in the near future.

However, we're going to be adding lots more UGC in the future, e.g. mentor profiles, so I think it's a better solution to just implement a lightweight dynamic server for v2 that renders views on-demand and aggressively caches them. Another benefit of this is that it will also make it easier for us to support standard, non-ajaxy form submission if we want, which would be a nice accessible solution for no-js folks too. (In other words, I think we should evolve this site so it "gracefully degrades" into a standard web-1.0 style website, but transforms into a badass hovercapable 1962 Chevrolet Corvette if the user's browser is cool enough.)

PR #580, opened 2015-04-03
Add css transitions for hamburger toggle.

### LINKED ISSUE #575: Use CSS transitions for sidebar/hamburger toggle on mobile
Right now it just appears/disappears without any transition, which is a bit jarring. Shouldn't be hard to add some nifty transition effects via [`ReactCSSTransitionGroup`](https://facebook.github.io/react/docs/animation.html).

toolness:
This fixes #575.

I set the transition to 0.5 seconds instead of 1 because when collapsing the burger, there is a [somewhat annoying](http://stackoverflow.com/questions/3508605/css-transition-height-0-to-height-auto#comment10354379_8331169) delay due to the fact that our `max-height` is transitioning from a very high value to 0 over that period.

This is also part of why I've added an opacity transition in addition to the max-height transition: the opacity change is immediately visible, which makes the hamburger feel more responsive when the content is being collapsed.

However, there's another reason for the opacity change too. Without it, some of the content inside the sidebar still displays when the hamburger is collapsed:

I'm not exactly sure why this is; maybe because we're using `position: relative` for some things inside the sidebar? In any case, I tried `overflow: hidden` on the sidebar content and that also results in bad things, so it was easier to just transition opacity to 0 when collapsing to ensure that the stuff inside the sidebar can't be seen.

I've tested this PR on Mobile Safari, latest Firefox, Chrome, IE11, and on a browser w/o JS enabled and all seem to work okay.

### DIFF
--- a/less/components/sidebar.less
+++ b/less/components/sidebar.less
@@ -8,7 +8,8 @@ html.no-js .sidebar {
   .collapsible-content {
     /* The links in the sidebar are crucial to navigating the site.
      * If JS is disabled, we want the links visible at all costs. */
-    display: block !important;
+    opacity: 1 !important;
+    max-height: none !important;
   }
 }
 
@@ -54,6 +55,19 @@ html.no-js .sidebar {
 
   a {
     color: inherit;
+  }
+
+  .collapsible-content {
+    transition: all 0.5s ease;
+    opacity: 1;
+    max-height: 1400px;
+  }
+
+  @media screen and (max-width: @screen-md) {
+    .collapsible-content.collapsed {
+      opacity: 0;
+      max-height: 0;
+    }
   }
 }
 
--- a/components/sidebar.jsx
+++ b/components/sidebar.jsx
@@ -76,7 +76,7 @@ var Sidebar = React.createClass({
         </div>
         <div className={this.state.showCollapsibleContent
                         ? "collapsible-content"
-                        : "hidden-xs hidden-sm collapsible-content"}>
+                        : "collapsed collapsible-content"}>
           <Login/>
           <ul className="sidebar-menu list-unstyled">
             {this.MENU_ENTRIES.map(function(entry, i) {
--- a/test/browser/sidebar.test.jsx
+++ b/test/browser/sidebar.test.jsx
@@ -29,7 +29,7 @@ describe("sidebar", function() {
   });
 
   it('should hide collapsible content', function() {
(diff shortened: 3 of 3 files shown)


less/components/sidebar.less
```diff
@@ -55,6 +56,19 @@ html.no-js .sidebar {
   a {
     color: inherit;
   }
+
+  .collapsible-content {
+    transition: all 0.5s ease;
+    opacity: 1;
+    max-height: 1400px;
```
ScottDowne: Given that the height of the menu is fairly static, we can probably get away with 50-60rem/500-600px

This would reduce the transition wait.

Another trick to helping with the transition is to toggle between a .5 and 1 second transition.

The delay is because when collapsing it starts at the bottom and needs to go through the "extra" pixels before you notice the transition, but when opening it starts at the top and the transition is immediate.

So when collapsed, set the transition to faster, and when expanded set the transition to longer.

You can see an example of this in the expander which has to deal with high and low sized contents.

Here it is expanded/focused with a slow transition: https://github.com/mozilla/teach.webmaker.org/blob/develop/less/components/expander.less#L29

When closed, I really speed it up: https://github.com/mozilla/teach.webmaker.org/blob/develop/less/components/expander.less#L37

Might help.

ScottDowne:
The opacity/overflow is super weird. I cannot figure out why it's doing it yet. Curious.

ScottDowne:
Hm, it seems adding `overflow: hidden` to `.sidebar` and not the content seems to hide the weirdness without adding other issues that I can see.

That being said, and argument can be made that the opacity transition actually looks pretty cool, so maybe we keep it anyway?

less/components/sidebar.less
toolness: Ah cool, thanks, I will bust that up!

toolness:
> That being said, and argument can be made that the opacity transition actually looks pretty cool, so maybe we keep it anyway?

Hehe, sounds good to me :)

less/components/sidebar.less
toolness: Actually, I am trying out the animation now and I think it's fine as-is?

Here it is online (for a very short time) if you or @iamjessklein or @vazquez want to try it out: http://mozteach.toolness.org/

ScottDowne: Yeah, it isn't a huge deal.

ScottDowne: That being said, This can be an r+ now.

Issue #388, opened 2015-03-24
Build Events Resources page


hannahkane:
This is a child page of /Events

Design is here: http://invis.io/ZU2HJU6JR

@vazquez to provide any needed assets.

vazquez:
Before you take on this ticket, let's talk about the different interactions that exist on this page.

toolness:
Rad!

Hmm it looks like this will require figuring out #232.

iamjessklein:
Hey - I haven't had a chance to review the ux here - mind if I take a stab at that today?

toolness:
Hmm, I'm curious what the tab bar should work like on browsers that don't have JS... 

![2015-03-25_10-40-52](https://cloud.githubusercontent.com/assets/124687/6827013/76e68a06-d2db-11e4-97cb-cf5e1463a6b9.jpg)

We could potentially have different routes for each tab, e.g. `/events/resources/before-event/` for the "before events" tab, or we could just require users have JS enabled to see the other tabs.

Alternatively, we could design the CSS in such a way that the tabs actually just look like a big long expanded list when JS isn't enabled. This has the added benefit of being useful as a `@media print` stylesheet, if users want to print out the page as a handy reference.

That said, making this accessible looks like it might actually involve a lot of work for a Friday deadline, so if there's any way for us to do *only* the accessible version (i.e. the version w/o any tab bar interactivity) for v1, that would be rad.

vazquez:
@iamjessklein Yes, if you find issues I think it's best to file them as separate tickets so that this can go through.

iamjessklein:
thanks, @vazquez will do.

vazquez:
Interesting points @toolness!

I am deferring to @hannahkane for that decision. Hannah, we are talking of the tab bar looking like below. The thing with this is that the page would become super super long again. With that being said though, I think it is an excellent approach for users that do not have JS enabled.

![screenshot 2015-03-25 10 57 49](https://cloud.githubusercontent.com/assets/1119821/6827443/ce165c96-d2dd-11e4-8b8c-1fa0cf712ac5.png)

hannahkane:
I'm +1 on the non-JS approach. Happy to have it implemented first. And excited to implement the JS version very soon after (though it doesn't have to be a blocker for the first release).

toolness:
Yaaaaaaay! Thanks for the extremely quick mockup ricardo and thanks for the decision hannah!

vazquez:
[...]

Issue #94, opened 2015-02-25
Create stylesheet to print activities and resources


LauraHilliger:
Conversation was started here: https://github.com/mozilla/webmaker-curriculum/issues/99

LauraHilliger:
Before I spend time making this basic stylesheet for club curricula prettier so our stuff prints nicely, do we have a webmaker print stylesheet somewhere?

cc/ @k88hudson @toolness @flukeout @cassiemc 

Tips on print styles are appreciated :pray: 

https://github.com/mozilla/webmaker-curriculum/pull/167

toolness:
Cool!

I recently found this [PDFKit](http://pdfkit.org/docs/vector.html) thing, which makes it possible to render a PDF in JavaScript. It's not HTML-based, though, so I get the impression that it *could* be used to create some slick PDFs, but there would be some sort of conversion process involved in taking your original HTML/markdown and turning it into a PDF. It's probably way too much effort but I thought I'd mention it here in case it inspires any other funky ideas.

k88hudson:
This might be a good starting point

https://github.com/twbs/bootstrap/blob/master/less/print.less

LauraHilliger:
well it looks ok. Not brilliant, but ok: https://github.com/mozilla/webmaker-curriculum/pull/170

If we ever decide to move back to markdown, should remember: https://gitprint.com/

toolness:
<a href="https://github.com/mozilla/learning-networks/issues/45"><img src="https://raw.githubusercontent.com/toolness/learning-issue-migrator/master/img/migrated.png"></a>
This issue has been moved to https://github.com/mozilla/learning-networks/issues/45.
Please don't comment on it here.

PR #582, opened 2015-04-04
Resources

ScottDowne:


ScottDowne:
Work in progress, but I wanted to get something out there.

ScottDowne:
I think this is ready for review.

Two things missing, the top nav links are un styled. Ricardo said this still needs to be designed.

I also have a bunch of links to create in the content which I simply have not yet. I think we should try for follow ups for those and otherwise land what I have already.

toolness:
Code reviewing this now!

iamjessklein:
yay :book:

toolness:
For `img/event-resources-page/resource-thumbnails-01.png` and its 2x version--these are PNGs of vector art that is already at [`img/events-page/maker-party-logo.svg`](https://github.com/mozilla/teach.webmaker.org/blob/develop/img/events-page/maker-party-logo.svg). The vector art version is only 8k, while the 2x version is 23k... so we might want to use the vector one instead? The same could potentially go for other art assets in this PR, but we could alternatively just file this as a separate issue.

toolness:
The "event support" section at the bottom seems a bit non-responsive to different page widths... For instance, it looks good when my window is an average size, but when my window is really wide, there's extra space left over on the right side:



and when my window is really narrow, the second column drops down below the first, leaving extra space over on the right again:



We can file this as a separate bug, just wanted to document it for now.

toolness:
When I hover over a resource and see its list of downloadables, it seems like there ought to be an extra `&nbsp;` between the icon and the text:

pages/event-resources.jsx
```diff
@@ -0,0 +1,407 @@
+var React = require('react');
+var ImageTag = require('../components/imagetag.jsx');
+
+var LinksContainer = React.createClass({
+  render: function() {
+    return (
+      <div className="row">{this.props.children}</div>
+    );
```
toolness: This seems like a really trivial React class, do you think it should just be inlined into whatever's using it, or do you think it will eventually become more complex/useful?

pages/event-resources.jsx
```diff
@@ -0,0 +1,407 @@
...
+  }
+});
+var EventItem = React.createClass({
+  render: function() {
+    return (
+      <div className="event-item">
+        <div className="event-icon">
+          <div className="participants-number">{this.props.participants}</div>
+          <div className="participants-label">PARTICIPANTS</div>
```
toolness: I prefer to use non-uppercase text (e.g. "Participants") and just use CSS to turn it into all-caps, but it's up to you.

pages/event-resources.jsx
```diff
@@ -0,0 +1,407 @@
...
+    );
+  }
+});
+var LogoAsset = React.createClass({
+  render: function() {
+    return (
+      <div className="logo-asset-container col-sm-4 col-md-4 col-lg-3">
+        <div className="logo-asset-header">{this.props.head}</div>
+        <ImageTag className="logo-asset-img"
+        width={200} height={200}
```
toolness: Should this be specified in CSS? I know sometimes there are situations where we need to specify it in the attributes, just not sure what this situation calls for.

pages/event-resources.jsx
```diff
@@ -0,0 +1,407 @@
...
+          <div className="logo-asset-center">{this.props.children}</div>
+        </div>
+      </div>
+    );
+  }
+});
+var LogoAssetLink = React.createClass({
+  render: function() {
+    return (
+      <a className="logo-asset-link" href={this.props.href}><span className="ion ion-ios-copy"></span>{this.props.children}</a>
```
toolness: What's the `ion` class? I don't see it in the less with this PR, and it's kind of hard to grep for.

pages/event-resources.jsx
```diff
@@ -0,0 +1,407 @@
...
+  }
+});
+var RemixLink = React.createClass({
+  render: function() {
+    return (
+      <a className="remix-link" href={this.props.href}>Remix</a>
+    );
+  }
+});
+var Tabulator = React.createClass({
```
toolness: Now that we're using react-bootstrap, we could potentially use its tabs for this, but it's up to you... That said, though, as per https://github.com/mozilla/teach.webmaker.org/issues/388#issuecomment-86063715 I thought for v1 that we were going to actually go with a non-JS approach, so that it's both accessible and printable?

pages/event-resources.jsx
```diff
@@ -0,0 +1,407 @@
...
+              <div onClick={onClick} className={className} key={key}>{section.head}</div>
+            );
+          })}
+        </div>
+        {this.props.tabs.map(function (section, key) {
+          var className = "tabulator-content-container";
+          className += " tab-" + key;
+          return (
+            <div key={key}>
+              <div className="tabulator-head-no-js">{section.head}</div>
```
toolness: Ooo `no-js` support! NICE

pages/event-resources.jsx
```diff
@@ -0,0 +1,407 @@
...
+      <div className={className}>
+        <div>
+          {this.props.tabs.map(function (section, key) {
+            function onClick(e) {
+              tabulator.showTab("tab-" + key);
+            }
+            var className = "tabulator-head col-sm-3 col-md-3 col-lg-3";
+            className += " tab-" + key;
+            return (
+              <div onClick={onClick} className={className} key={key}>{section.head}</div>
```
toolness: Should this also get a `keydown` handlere and an `aria-role` attribute for accessibility? Also, it seems like the non-selected tab might need the [`aria-hidden`](http://www.w3.org/TR/wai-aria/states_and_properties#aria-hidden) attribute but I'm not knowledgeable enough about a11y to be certain. Tests for this would also be ultra cool but can file as separate bug.

pages/event-resources.jsx
```diff
@@ -0,0 +1,407 @@
...
+
+var EventsResources = React.createClass({
+  statics: {
+    pageClassName: 'event-resources'
+  },
+  render: function() {
+    return (
+      <div className="inner-container">
+        <h1>Event Resources</h1>
+        <p>That you for helping us celebrate webmaking around the world! These resources will help you plan a unique event tailored especially for your audience.</p>
```
toolness: Typo, "That" -> "Thank"

pages/event-resources.jsx
```diff
@@ -0,0 +1,407 @@
...
+        </div>
+
+        <h2 id="event-details">Event Details</h2>
+        <p>There are many details to consider when planning your event. Here's a list of what you need to know:</p>
+
+        <Tabulator tabs={[
+          {
+            "head": "the basics",
+            "content": (
+              <div>
```
toolness: This looks kind of un-reacty but we can think of ways to make it more reacty later :)

pages/event-resources.jsx
toolness: Ohh I just tried it out with no JS and it looks A W S M !  Nice job. :+1: 

That said, I noticed that the printability of the page also depends on whether the browser has JS enabled or not--if the browser has JS, then only the content of the current tab is printed, while all tabs are printed if the browser doesn't have JS. Shouldn't be hard to fix though--and the rest of the page has issues w/ printing too, but I suspect lots of our pages do :)  So we should file that as a separate issue.

toolness:
Ok nice work! I'm marking it as R+, and it's up to you whether to file some/all of my concerns as issues vs. working on them in this PR.

pages/event-resources.jsx
ScottDowne: I tried. It had a weird parent/child markup relation.

I wanted the links and meta data to live in the main render bit, I needed it to have children, and I wanted the head to be attached to the content while inside the render bit.

However, once I got into tabulator, I needed all the head bits to live at the top, and the content to be attached to them via js.

I went through a few iterations to find one that held these values first. Kept hitting blockers but this one worked out.

I wanted to spend more time on this but came to a point where I was happy enough, for now.

So yes, later, but I just wanted to shed some light on where and is where it is :)

ScottDowne: It's an interesting problem. Maybe a better way to explain it is with examples :D Here is how I wanted the data to exist, when passed to Tabulator:

```
<Tabulator>
  <TabItem head="first">
    <p>content</p>
  </TabItem>
  <TabItem head="second">
    <p>content</p>
  </TabItem>
  <TabItem head="third">
    <p>content</p>
  </TabItem>
  <TabItem head="fourth">
    <p>content</p>
  </TabItem>
</Tabulator>
```

Given a data and markup style like this, how can I generate the tabulator?

toolness: Yeah, totally! I like that way too, but I also have no idea how to generate the tabulator from it, since `this.props.children` is so doggone opaque. React does have a [`React.Children`](https://facebook.github.io/react/docs/top-level-api.html#react.children) that might be useful for such a thing though.

pages/event-resources.jsx
ScottDowne: I think I need to make this a button for a11y.

pages/event-resources.jsx
ScottDowne: I'm not familiar with the term "printing" it's new to me, but happy to fix with a little help :)

pages/event-resources.jsx
ScottDowne: Yeah, you're probably right.

Fixing

pages/event-resources.jsx
ScottDowne: Agreed, done.

pages/event-resources.jsx
ScottDowne: Normally, yes, but with the way react handles dynamic content, it doesn't work well when it is in css.

pages/event-resources.jsx
ScottDowne: Right now it's form https://github.com/mozilla/teach.webmaker.org/blob/develop/package.json#L48

ScottDowne:
@toolness Updated.

pages/event-resources.jsx
toolness: Lol. I mentioned printing because I know it's an issue for other parts of the site, like teaching activities (see #94), and I figured that since this page has so much content, some people might want to print it out for handy reference too, but @LauraHilliger is a better judge than me.

From a technical standpoint this should be pretty easy though, just do a [print media query](http://www.joshuawinn.com/css-print-media-query/) and give it the same styling as your no-js rule.

pages/event-resources.jsx
toolness: Hmm ok. I kinda wonder if there is a way to just have screen-readers and such see the no-JS version of this tabset, so that it looks to them like a big ole page. That might actually be *more* accessible than making all the interactions work in an ARIA-compliant way, but I am no expert.

Anyhow sounds good!

pages/event-resources.jsx
ScottDowne: Was over thinking the print bit. I tried to setup a view for print but it didn't seem to work.

I say we punt this as well while I figure out wtf is up with the print media query.

toolness: Ok that's fine with me. I filed #638 to address it.

PR #586, opened 2015-04-06
Only toggle hamburger on enter/space.

### LINKED ISSUE #581: Only enter keypresses should toggle hamburger
While working on #580 I noticed that *any* keypress triggers the hamburger toggle when the hamburger is focused. Aside from being kind of annoying, it also makes it impossible (or at least very frustrating) for keyboard-only users to navigate to the links within the sidebar, because as soon as the user presses tab to navigate to them, the hamburger immediately un-collapses!

toolness:
This fixes #581.

### DIFF
--- a/components/sidebar.jsx
+++ b/components/sidebar.jsx
@@ -64,6 +64,15 @@ var Sidebar = React.createClass({
       showCollapsibleContent: !this.state.showCollapsibleContent
     });
   },
+  handleHamburgerKeyDown: function(e) {
+    // We're simulating a button, so we should simulate a click when
+    // the user presses enter or space:
+    //
+    // http://www.w3.org/TR/wai-aria-practices/#button
+    if (e.which === 13 || e.which == 32) {
+      this.handleHamburgerClick();
+    }
+  },
   render: function() {
     return (
       <div className="sidebar col-md-3">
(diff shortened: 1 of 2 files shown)


components/sidebar.jsx
```diff
@@ -71,7 +80,7 @@ var Sidebar = React.createClass({
           <Link to="home">
             <img src="/img/nav/mozilla-wordmark-white.svg" alt="Webmaker logo" className="moz-logo"/>
           </Link>
-          <span aria-label="toggle" role="button" onKeyUp={this.handleHamburgerClick} onKeyDown={this.handleHamburgerClick} className="glyphicon glyphicon-menu-hamburger hidden-lg hidden-md"
+          <span aria-label="toggle" role="button" onKeyDown={this.handleHamburgerKeyDown} className="glyphicon glyphicon-menu-hamburger hidden-lg hidden-md"
```
ScottDowne: Larger question, why is this using onKeyDown and not onClick?

toolness: It is using `onClick` in the line below this one... we have to have *both* an onClick handler and an onKeyDown handler to support the ol' a11y.

toolness: Was this the only issue you had with the PR? I wasn't sure, so I re-marked this one as "PR needs review"...

toolness: Oh, I am realizing now that the title of this PR might be a bit misleading. It's called "only toggle hamburger on enter/space" because at present, pressing *any* key toggles the hamburger. It doesn't mean "only enter/space should toggle the hamburger, ignore all other input devices" :)

ScottDowne: Can we make this span a button? Then the onClick should do exactly what we need without needing to filter input and we can then also remove role="button" I think.

toolness: Good idea--I've just filed #623 for that, and will close this PR.

PR #593, opened 2015-04-06
Improve form validation.

### LINKED ISSUE #424: URL validation on 'website' field in club modal should be friendlier
As mentioned in https://github.com/mozilla/teach.webmaker.org/issues/352#issuecomment-84214121, we are currently relying on `type="url"` to validate the website field in Club modals. This ultimately means that the validation is at the discretion of the host browser's built-in HTML5 form validation; on some browsers it says "Please enter a URL" while on others it may say "Please enter a web address that starts with http:// or https://" while others may say something different.

This isn't super ideal. Also, if the user enters "example.org" we could always prepend "http://" before sending the data to the server (the server uses a [`URLField`](https://docs.djangoproject.com/en/1.7/ref/models/fields/#urlfield) and thus expects any non-blank value to be a http or https URL).

### LINKED ISSUE #540: MOBILE: Didn't receive proper error message when I left required field blank on Clubs form
When using Safari on an iPhone (iOS 8), I didn't receive a proper error message when I left the "What is the name of your Club" or the "What do you focus your efforts on?" field blank. 

I got, "Unfortunately, an error occurred when trying to change your club. Please try again later." 

Should have received a message similar to "Please provide a location for your Club" (which is what I see if I leave the location field blank."

Suggested copy:

"Please provide a name for your Club"
"Please provide a description for your Club"

toolness:
This shows form validation errors for the add/change club modal at the top of the form.

Much of this is a fallback for browsers that don't support HTML5 form validation, which fixes #540.

However, it also validates the "location" field, which currently has very bad (and untested) validation that raises a `window.alert` if the user hasn't filled it out.

One thing that this does *not* currently do is validate the website field (if it's filled out). That's because there's an open question on how to exactly to resolve #424, and I'd like to resolve that in a separate PR.

### DIFF
--- a/pages/clubs.jsx
+++ b/pages/clubs.jsx
@@ -218,6 +218,22 @@ var ModalRemoveYourClub = React.createClass({
   }
 });
 
+var validateClub = function(clubState) {
+  var errors = [];
+
+  if (!clubState.name) {
+    errors.push("You must provide a name for your club.");
+  }
+  if (!clubState.description) {
+    errors.push("You must provide a description for your club.");
+  }
+  if (!clubState.location) {
+    errors.push("You must provide a location for your club.");
+  }
+
+  return errors;
+};
+
 var ModalAddOrChangeYourClub = React.createClass({
   mixins: [React.addons.LinkedStateMixin, ModalManagerMixin,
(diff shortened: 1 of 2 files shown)


ScottDowne:
R+

PR #596, opened 2015-04-06
homepage - edited button copy

### LINKED ISSUE #545: LANDING: "Join" button copy and CTA should be rethought
Right now the call to action is a bit ambiguous - the main CTA is "Unlock opportunities for all citizens of the Web," and then directs users to click the "Join us" button. It's unclear to me what I am joining and what the value is for me to actually Join. As far as I can tell the only action that I can perform exclusively with login is adding your club to the map. I am thinking that for the time being we might want to promote that instead of a hippy dippy "join us" - we are movers and shakers. <insert snark>  Then, in the future, if we roll out more functionality, we should promote it more prominently in this space. 

screen shot 2015-04-02 at 10 10 50 am

/cc @kevzawacki @ldecoursy

iamjessklein:
Instead of “learn more” it now says “start a Mozilla club”

Fixes #545

### DIFF
--- a/pages/home.jsx
+++ b/pages/home.jsx
@@ -25,7 +25,7 @@ var HomePage = React.createClass({
       <div>
         <HeroUnit image="/img/hero-unit.jpg">
           <h1>Unlock opportunities for all citizens of the Web.</h1>
-          <div><Link to="join" className="btn btn-awsm">Join Us</Link></div>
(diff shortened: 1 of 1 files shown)


iamjessklein:
@ScottDowne please review when you have a chance :four_leaf_clover:

alicoding:
r+

PR #600, opened 2015-04-06
Enable the softest of launches.

### LINKED ISSUE #589: Remove references to Login for the softest of launches
As per discussion with @toolness and @simonwex, we'll plan to deploy to production without authentication in place, so that we can deal with any deployment hiccups now, rather than later.

The site will be on production, but we won't link to it or promote it until the login project is finished and incorporated into the Teach site.

In order to make it not weird for anyone who stumbles across the site, we'll remove login-dependent functionality:
 - [ ] Remove "Create an account" and "Sign in" from sidebar nav
 - [ ] Remove "Join Us" button from homepage hero image
 - [ ] Remove "Add Your Club" button from /mozilla-web-clubs hero image
 - [ ] Remove map and list of clubs from /mozilla-web-clubs
 - [ ] Remove "Add Your Club to the Map" CTA at the bottom of /mozilla-web-clubs and  /clubs/toolkit

toolness:
This fixes #589 by adding an optional `SOFTEST_OF_LAUNCHES` environment variable. If it's defined, we hide all the login-related stuff.


toolness:
For the very near future, you can see what the site looks like with `SOFTEST_OF_LAUNCHES=on` at http://mozteach.toolness.org/.

ScottDowne:
Hm, possible idea to make this a bit more readable/maintainable.

Would you agree or disagree with a react module called `<LoginRequired></LoginRequired>` that we can wrap around things. It them either adds `{this.props.children}` or doesn't depending on the process.env value.

Another thought it to change the env name to `REMOVE_LOGIN`.

Thoughts?

toolness:
Ah, I was a bit lackadaisical with the naming of the environment variable since I assume it'll only be around for a few weeks :)

I think I will just rename the env var to `REMOVE_LOGIN`... I didn't want to modularize this code too much because (a) that would require a `require` which we might forget to remove when excising this stuff, and (b) a neat thing about inlining the `process.env.REMOVE_LOGIN` is that the actual content inside the conditional gets stripped out at minification time as a part of dead code removal, which I thought was neato.

toolness:
I like the idea of a `<LoginRequired>` tag for a situation in which we could make the site dynamically change based e.g. on a checkbox in the dev ribbon modal :)  But if it's a pure build-time setting, I think that doing things the way they're currently being done is nice because it strips out the login-required code. (I don't *think* the same thing happens when the login-required code is passed as an unused argument to a function, but I could be wrong.)

As you can see I also was a bit funky with my code formatting because I didn't want adding or removing the login-required-ness to mess with version history/blame information too much, e.g. because of changed indentation and whatnot.

ScottDowne:
Sounds like this is a very temporary env. If so, R+.

PR #608, opened 2015-04-07
[WIP] Initial support for OAuth2.

### LINKED ISSUE #576: Redirect to teach-api for login/logout flow
As mentioned in https://github.com/mozilla/teach.webmaker.org/pull/445#issuecomment-89299936:

We're moving to using an OAuth2 flow for login via [id.webmaker.org](https://github.com/mozilla/id.webmaker.org).

Since the OAuth2 client secret can only be stored in the teach API, I'm just doing *all* OAuthy stuff on the teach API (see https://github.com/mozilla/teach-api/pull/11).

I've added an `/auth/status` endpoint in https://github.com/mozilla/teach-api/issues/10 with an extremely strict CORS policy; essentially, every time the teach site loads, it needs to ping this endpoint to see if it's currently logged in.

Putting all the pieces together, the login flow will essentially work like this:

oauth2-flow-diagram

* User visits teach.mozilla.org for the first time. It pings `/auth/status` on the teach API and notices that no one is logged in.
* User clicks "login" or "create account" on teach.mozilla.org (either in the sidebar or through the [add a club modal](https://github.com/mozilla/id.webmaker.org/issues/51)).
* User is transparently redirected to teach-api, which immediately redirects to id.webmaker.org, where the user logs in/signs up.
* id.webmaker.org redirects the user back to teach-api (via OAuth2 callback), which then redirects the user back to teach.mozilla.org.
* When teach.mozilla.org loads, it pings `/auth/status` on the teach API and notices that the user is logged in.

Phew. It's not particularly ideal, but I think we will be moving to a simpler solution once we actually give the teach site its own lightweight HTML-rendering server in v2 (we need to do this anyways to fully support our progressive enhancement strategy and some other things).

I should also note that logout follows a similar pattern, because for now we'll want logging out of the teach site to also log the user out of id.webmaker.org for usability reasons--see https://github.com/mozilla/id.webmaker.org/issues/119#issuecomment-89008358 for more details on that.

toolness:
This fixes #576 by using the OAuth2 endpoints defined on teach-api in https://github.com/mozilla/teach-api/pull/11 to ultimately delegate login to id.webmaker.org.

Currently this can only be enabled by setting the `ENABLE_OAUTH2` environment variable to `on`, and the new functionality doesn't yet have any unit tests.


toolness:
This is now superseded by #626.

PR #610, opened 2015-04-07
fix the event resources link

### LINKED ISSUE #485: EVENTS: Add link to "Events Resources"
The bottom left icon of the page with the words "Events Resources" should be linked.

iamjessklein:
This fixes #485

This is dependent on this PR getting merged:
https://github.com/mozilla/teach.webmaker.org/pull/582

### DIFF
--- a/pages/events.jsx
+++ b/pages/events.jsx
@@ -79,7 +79,7 @@ var EventsPage = React.createClass({
           <section>
             <IconLinks>
               <IconLink
-                linkTo="fixme"
+                linkTo="/events/resources"
                 imgSrc="/img/events-page/icon-curriculum.svg"
                 imgAlt="icon resources"
                 head="Event Resources"
--- a/pages/events.jsx
+++ b/pages/events.jsx
@@ -79,7 +79,7 @@ var EventsPage = React.createClass({
           <section>
             <IconLinks>
               <IconLink
-                linkTo="/events/resources"
+                linkTo="event-resources"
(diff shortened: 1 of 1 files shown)


pages/events.jsx
```diff
@@ -79,7 +79,7 @@ var EventsPage = React.createClass({
           <section>
             <IconLinks>
               <IconLink
-                linkTo="fixme"
+                linkTo="/events/resources"
```
ScottDowne: This doesn't work like a standard link.

The linkTo attribute is a custom thing that allows a link to transition the menu colour. It's doesn't actually link, which is why the standard url won't work. You instead give it an id to the page.

To see the ids that are set, view the https://github.com/ScottDowne/teach.webmaker.org/blob/resources/lib/routes.jsx#L26 page.

That's the file from my pull request and the master branch.

You can see I setup a route called `name="event-resources"` so in order to link to it, use `event-resources` in the linkTo attribute.

iamjessklein:
did I do this work on the branch correctly? @ScottDowne

ScottDowne:
@iamjessklein it is correct, but we cannot merge it until event resources exists. That is why Travis is failing, because it knows you're using a link that doesn't exist.

So let's keep this open and merge it along with event resources page.

ScottDowne:
Tested this locally, merges clean and tests pass, so, merging and travis should be fixed once merged with develop.

Issue #592, opened 2015-04-06
Add Club description to list view


hannahkane:
Currently, the Club description is included in the Map callout view of Clubs, but not the list view. Please add it, right below the location and above the "Led by" field.

ScottDowne:
@hannahkane could you elaborate a bit more? I'm not sure what needs to be done yet.

ScottDowne:
I think I figured it out by reading a few more tickets and seeing a pattern.

hannahkane:
Just trying to make the list view and the map view consistent in terms of what information they display!

toolness:
Hey is @iamjessklein cool with this? I actually originally included it in the list but she wanted it removed bc it added clutter to the whole thing.

ScottDowne:
@toolness ah, that's fair.

iamjessklein:
see my comment on #611

PR #611, opened 2015-04-07
display discription in club list

ScottDowne:


ScottDowne:
Fixes #592

toolness:
That said, the actual code here looks fine, so if @iamjessklein is cool with it from a design standpoint, i say go ahead and merge!

iamjessklein:
To :bee: or not to :bee that is the :question: 
To take :muscle: against a :ocean: of troubles 
and by opposing, end them? :skull: or :sleeping: 

Let's go with a :+1:  R+ for consistency and re-evaluate if the content is overwhelming.

Issue #591, opened 2015-04-06
Add username and link to user's Webmaker profile to Club description on map


hannahkane:
The Club Leader's name currently appears in the list view, but not the map callout view.

Just like in the list view, it can be the last field, and should include a label, "Led by:"

PR #612, opened 2015-04-07
add username to map marker

ScottDowne:


ScottDowne:
Fixes #591

components/map.jsx
```diff
@@ -110,6 +110,7 @@ var MarkerPopupClub = React.createClass({
         <br/>
         <br/>
         <p>{this.props.description}</p>
+        <p><small>Led by <a href={"https://webmaker.org/en-US/search?type=user&q=" + this.props.owner}>{this.props.owner}</a></small></p>
```
toolness: Hmm, it would probably be good to add a helper function (or maybe even a super simple react component) somewhere that returns the profile URL for a user, to avoid DRY breakage between this link and the one in the clubs list, but I don't think it's a blocker or anything. R+ I say, R+!

ScottDowne: I mused about this for a bit.

Ultimately, I considered a club-view component. however, I came to the conclusion that the current club list is not scalable and needs to be updated somehow anyway, that doing too much on the engineering end of things might be not all that useful.

I feel like a component just for the link is too much, and I'm not sure how to make it a function a bridge the map to club page file gap.

I say we punt these problems.

toolness: Yep sounds good to me!

Issue #585, opened 2015-04-04
Use a lightweight server that dynamically renders HTML


toolness:
I've alluded to this in other issues and discussions, but I wanted to document it here.

When I was originally asked to come up with a site prototype for https://github.com/mozilla/teach.webmaker.org/issues/75, It seemed implicit that most of the site's content would be static (as opposed to user-generated). That was the assumption I relied on when building the [initial prototype](https://github.com/toolness/mozilla-learning-html/tree/86154883f857a5c954f4795bf25a289fada50967) of the site, which generated a static website that could be uploaded to S3.

However, the next heartbeat, I saw the full teach site roadmap for the first time, and it included quite a bit of dynamic functionality powered by user-generated content, like mentor profiles. Because I had already started productionalizing the prototype and the v1 of the teach site was still going to be largely static content, I didn't propose to change the static site nature of the site.

However, now that we're on the cusp of releasing v1, I'd like to formally propose moving to a slightly different architecture whereby a lightweight server dynamically renders HTML for clients and aggressively caches it. Or, as mentioned in https://github.com/mozilla/teach.webmaker.org/pull/579#issuecomment-89548371:

> ... I think we should evolve this site so it "gracefully degrades" into a standard web-1.0 style website, but transforms into a badass hovercapable 1962 Chevrolet Corvette if the user's browser is cool enough.

This will enable a number of use cases:

* Mentor profiles, clubs, and other user-generated content (UGC) can have their own clean URLs, e.g. `/mentors/toolness`, without needing to constantly regenerate the site every time a user adds/changes their content.
* UGC can be pre-loaded into the initial HTML sent to the user, improving initial page load experience.
* UGC could be easily viewed by users without JS enabled, as well as scraped by web spiders for SEO (if we want to allow that kind of thing).
* UGC could potentially be *submitted* by users without JS enabled.
* The site can potentially deliver different content for certain user agents (see https://github.com/mozilla/teach.webmaker.org/issues/151#issuecomment-77690066).
* Basic locale information will be easier to detect and respond to on the server-side.

Fortunately, because of the way we're currently generating the static site, this isn't actually a very hard thing to do. I was also talking to @jdotpz and there's a way we can still even use the generated static site as a fallback if the server goes down for some reason, or if we need to perform maintenance on it--that way, at least users will *always* be able to see the basic static content of the site.

Any thoughts on this approach are appreciated!

rossbruniges:
[...]

PR #654, opened 2015-04-10
Fixes #481 - add mailinglist signup form

### LINKED ISSUE #481: EVENTS: Add mailinglist sign up form
There is an itsy bitsy button underneath the carousel. It needs to be enlarged. It should link to the Clubs page. 
screen shot 2015-03-31 at 5 25 07 pm

mmmavis:
This fixes #481

### DIFF
--- a/pages/events.jsx
+++ b/pages/events.jsx
@@ -127,7 +127,7 @@ var EventsPage = React.createClass({
             header=""
             dividerImgSrc="/img/pages/events/svg/line-divider.svg">
               <div className="row" id="mailinglist">
-                { (this.context.router.getCurrentQuery().mailinglist === "thank")
+                { (this.context.router.getCurrentQuery().mailinglist === "thanks")
                  ?
                   <div>
                     <p>Thank you for signing up!</p>
--- a/pages/events.jsx
+++ b/pages/events.jsx
@@ -134,7 +134,7 @@ var EventsPage = React.createClass({
               <div className="row" id="mailinglist">
                 { (this.context.router.getCurrentQuery().mailinglist === "thanks")
                  ?
-                  <div className="thank-you-message">
+                  <div>
                     <p>Thank you for signing up!</p>
                   </div>
                  :
--- a/pages/events.jsx
+++ b/pages/events.jsx
@@ -70,7 +70,12 @@ var FormMailingListSignup = React.createClass({
 var EventsPage = React.createClass({
   statics: {
     pageTitle: 'Events',
-    pageClassName: 'events'
+    pageClassName: 'events',
+    FormMailingListSignup: FormMailingListSignup,
+    validateSignupForm: validateSignupForm
+  },
+  contextTypes: {
+    router: React.PropTypes.func.isRequired
   },
   contextTypes: {
     router: React.PropTypes.func.isRequired
@@ -129,7 +134,7 @@ var EventsPage = React.createClass({
               <div className="row" id="mailinglist">
                 { (this.context.router.getCurrentQuery().mailinglist === "thanks")
                  ?
-                  <div>
+                  <div className="thank-you-message">
                     <p>Thank you for signing up!</p>
                   </div>
                  :
--- a/pages/events.jsx
+++ b/pages/events.jsx
@@ -1,14 +1,60 @@
 var React = require('react');
-
+var request = require('superagent');
 var HeroUnit = require('../components/hero-unit.jsx');
 var IconLinks = require('../components/icon-links.jsx');
 var IconLink = require('../components/icon-link.jsx');
 var Illustration = require('../components/illustration.jsx');
 
-var CAROUSEL_IMG_PROPS = {
-  width: 800,
-  height: 533
-};
+var FormMailingListSignup = React.createClass({
+  mixins: [React.addons.LinkedStateMixin],
+  getInitialState: function() {
+    return {
+      formDest: "https://sendto.mozilla.org/page/s/maker-party-signup",
+      email: "",
+      privacyPolicyAgree: false
+    };
+  },
+  handleSubmit: function(e) {
+    e.preventDefault();
+    console.log(this.state.email);
+    console.log(this.state.privacyPolicyAgree);
+    console.log(this.state.formDest);
+
+    request
+      .post(this.state.formDest)
+      .withCredentials()
+      .send({
+        email: this.state.email,
+        "custom-2517": this.state.privacyPolicyAgree
+      })
+      .end(function(err, res) {
+        if (err) {
+          console.log(err);
+          window.alert("error encountered");
+        }
+      });
+  },
+  handleBtnClick: function(e) {
+    window.alert("submit btn clicked");
+  },
+  render: function() {
+    return (
+      <form action={this.state.formDest} method="POST" onSubmit={this.handleSubmit}>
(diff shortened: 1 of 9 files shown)


mmmavis:
Hey @toolness ,

I'm currently blocked on fixing CORS problem when POSTing to BSD.  Basically BSD has no API for us to use and (we can try but) I don't think we can easily ask them to whitelist our domains.

I talked to @jbuck and he suggested 2 options

1. XHR post to teach-api, and teach-api can POST to BSD on the clients behalf
2. have the users browser do a form submission to BSD, and have BSD thank you page be a redirect back to teach site

I would like to see if you have a preferred approach.  :hamster:

toolness:
I recommend just doing option 2, as using teach-api as a go-between is a bit of a hassle right now. Once we implement #585, solutions like option 1 will be a lot easier.

mmmavis:
Sigh, now I think of it I think we probably won't be able to redirect users back to the a page we desire (e.g., back to Teach site).

Reason:
The underlying [BSD form](https://sendto.mozilla.org/page/s/maker-party-signup) has a Thank You page set up.  Our current config is to just replace the content in the wrapper and display thank you message there.  See [this sample thank you page](https://sendto.mozilla.org/page/st/maker-party-signup?action_code=FgxRWxYUOVIKQV0YAFcDTFM&td=TVFNSwMxEP0rISeF7jbZbtmPXjwpgkWhF8FKyGan29BssuSjWMX_bmKLNuQwM-_NvMnLF4aRS4Vb7MF5ejfyo3QplHqgxSLX4PEMW-ilBeFZsIkaK2wIsr-ErONKGX_JzOTZMWE6KDXDzgQr4Iy50F2n095o-ON5E1XYTipwuH17jxU56DCxnbEj-9WiBYldwgL3wPqkVxC6zEiZFRTRpl2SlhT_DDkx3vc20iqSV01Oa5rXVxOCA8v4ADpNWptPqRTfzpc5QTdrLqT2xu1X6FF7UCgW0PMGvSJK8maF7LFdVDm5RQ8gDmY7j3uQeCm6jzbtzMd2nuAkteeTjzJpfzrDnQpxcy7dKQJSn59Vgqi6ciGyDro6oxTKrO4bnpGrE0clHzQfk3drfgCLXrj1J7T5dQk9xV-LpNH0QcGFdjYQf_8A) (notice there's super long query string in the url. it's BSD generated and I have no clue what it is for). In config we *can* tell BSD to redirect to a url for thank you page instead of replacing page content, BUT since I'm not sure how this BSD form is used I can't be sure 100% of the traffic comes from Teach site.  That's why I think a redirect back to the Teach site isn't a good option.

I'm really not sure what the best solution is here.  I could try refreshing the wrapper used by the BSD form and the Thank You page to make it less janky... if that helps...

cc @ScottDowne in case you have suggestions

davidascher:
Can't we create a new form on BSD that is specifically for use from the
teach site, and have the redirect be specifically for that flow?

mmmavis:
@davidascher  oh snap, problem solved! (for now) :+1:

mmmavis:
I guess all we care about it to get people added to the right constituent group?  I'll double check with Hannah :dancer:

hannahkane:
wfm.

pages/events.jsx
```diff
@@ -1,15 +1,69 @@
...
+        </div>
+      );
+    }
+  },
+  render: function() {
+    return (
+      <form className="mailinglist-signup" action="https://sendto.mozilla.org/page/s/maker-party-signup-for-teach-site" method="POST" onSubmit={this.handleSubmit}>
+        <div className="col-sm-offset-1 col-sm-8 col-md-offset-1 col-md-8 col-lg-offset-1 col-lg-8">
+          <fieldset>
+            <input name="email" type="email" size="30" placeholder="Your email address" valueLink={this.linkState("email")} required />
```
toolness: Can you add a `<label>` with the [`sr-only`](http://getbootstrap.com/css/#conveying-meaning-to-assistive-technologies) class above this input field that links to the `id` of the input field? That way it'll be nice and accessible, which will help us resolve #722.

pages/events.jsx
```diff
@@ -1,15 +1,69 @@
...
+          </fieldset>
+          <fieldset>
+            <input name="custom-3460" type="checkbox" checked readOnly required hidden />
+            <p className="pp-note">&#10003; I'm okay with you handling this info as you explain in your <a href="https://www.mozilla.org/en-US/privacy/websites/">privacy policy</a>.</p>
```
toolness: Hmm, I'm not sure how this works w/ assistive technologies. Hopefully the description you gave is enough here.

mmmavis: I added a `<label>` here as well... better repeat then not inform I guess :open_mouth:

pages/events.jsx
mmmavis: yes, should've done so.

mmmavis:
when this is on staging we should verify that the redirect works.  ie. `http://teach.mofostaging.net/events/?mailinglist=thanks#mailinglist`

toolness:
Nice! Do you think you could add tests for this?

mmmavis:
@toolness tests have been added :bowtie: https://github.com/mmmavis/teach.webmaker.org/commit/a04708729ea799bfb82611139876707b6b1847fd

pages/events.jsx
```diff
@@ -63,6 +128,23 @@ var EventsPage = React.createClass({
                 <p className="callout-heading">Check out the highlights from Maker Party and see more photos in our <a href="https://www.flickr.com/photos/mozilladrumbeat/galleries/72157643962655534/">Flickr gallery</a>.</p>
               </div>
             </div>
+            <PageEndCTA
+            header=""
+            dividerImgSrc="/img/pages/events/svg/line-divider.svg">
+              <div className="row" id="mailinglist">
+                { (this.context.router.getCurrentQuery().mailinglist === "thanks")
+                 ?
+                  <div className="thank-you-message">
```
toolness: Since this class is only used for testing, could you call it maybe `test-thank-you-message' or something similar, to indicate that it's only used for testing, and not for CSS styling?

Alternatively, you could just do `node.textContent.should.match(/thank you/i)` of an ancestor node in your test suite.

toolness:
Nice, this is looking really good! Thanks for adding the comprehensive tests. 

Where does the "thank you" message appear? I tried submitting the form on my local dev instance and the URL changed to have `?mailinglist=thanks` in it, and the mailing list signup form isn't visible on the page anymore, but I don't see a "thank you" message either...

toolness:
Oh wait! LOL the issue we anticipated in https://github.com/mozilla/teach.webmaker.org/issues/481#issuecomment-97882062 just happened to me: when I filled out the form, I was redirected to teach.webmaker.org rather than my own local dev instance, and since production doesn't yet have this PR landed on it, of course there was no "thank you" message. So in other words, forget my previous comment. :)

toolness:
Aside from my [one note](https://github.com/mozilla/teach.webmaker.org/pull/654#discussion_r30031916), this looks great mavis! Feel free to fix it and then merge.

mmmavis:
>  I was redirected to teach.webmaker.org rather than my own local dev instance, and since production doesn't yet have this PR landed on it, of course there was no "thank you" message.

:stuck_out_tongue_closed_eyes: heh, I totally feel you.  With things with BSD we have to test on "PROD" (sad face) that's why it often causes confusion.  I'm gonna set the redirect URL to `https://teach.mofostaging.net/events/?mailinglist=thanks#mailinglist` and see if that works out first.

PR #666, opened 2015-04-10
Show warning message if js is disabled/browser is old.

toolness:
This is a potential fix for #573. If the user's browser isn't supported for JS functionality (basically IE <= 9) or JS is disabled entirely, the following message appears at the top of the page:

The "update your browser" link goes to https://browser-update.org/update.html, which seems legit.


iamjessklein:
I like what this is saying, however the styling is so 1999 :stuck_out_tongue_winking_eye: 
Did @sabrinang or @vazquez design any treatment for alerts or notifications yet?

toolness:
> I like what this is saying, however the styling is so 1999 :stuck_out_tongue_winking_eye: 

Well, given that they're using an older browser, maybe they *like* 1999 :stuck_out_tongue:

iamjessklein:
:ok_hand: snaps @toolness :ok_hand:

ScottDowne:
@toolness status on this? Is it blocked on review only?

toolness:
That's a good question... I'm not sure if @iamjessklein was approving the retro 1999 styling in her [comment](https://github.com/mozilla/teach.webmaker.org/pull/666#issuecomment-92428176) but I'm going to be optimistic and assume she was. We should just review and land this, and then if it's decided that retro 1999 is *not* the way to go, we can file a separate issue to design and tweak the CSS a bit.

ScottDowne:
@toolness I say R+ and land it, we can open a new ticket to change the style.

PR #667, opened 2015-04-10
Lazy-load Tabzilla.

### LINKED ISSUE #658: Lazy-load tabzilla
Tabzilla is a bit annoying to me because it loads jQuery, which the rest of our site doesn't need (React-based code often doesn't need it, since React takes care of DOM manipulation for us). This will slow down page load times for our users, and it's particularly annoying because lots of our users will never even click on the tabzilla tab!

There's a few potential things we can do to alleviate this:

* An easy, but not incredibly effective strategy is to add the `async` attribute to the `tabzilla.js` script, so that at least loading tabzilla doesn't slow down the page load time. However, this won't save much because the `tabzilla.js` script is actually not very big--it's its dynamic injection of jQuery that takes a long time.
* An easy, but not very user-friendly strategy is to simply not load `tabzilla.js` at all. The tabzilla link, when un-embellished by `tabzilla.js`, is actually just an `<a href="http://mozilla.org">`--so when a user clicks on the link, they'll simply be taken to mozilla org. This isn't as slick as the tabzilla widget sliding down, though.
* A harder, but more useful strategy is to lazy-load tabzilla JS only when the user hovers/clicks/taps on it. We'd have to also trigger the opening of tabzilla once it loaded, if the user indeed did click/tap on it. A potentially annoying side effect of this, though, is that the first time a user clicks/taps on tabzilla, there may be a perceptible delay. That said, we could wrap this up into a handy `<Tabzilla>` React component that our other properties could use, which might be useful.

Before actually investigating too far into this, though, we should profile the performance of the whole site. As [Amdahl's Law](http://en.wikipedia.org/wiki/Amdahl%27s_law) suggests, if Tabzilla's load time is minimal compared to other factors slowing down our site, spending lots of time to optimize Tabzilla's load time won't do us a lot of good.

toolness:
This fixes #658 by only loading Tabzilla when the user has clicked/tapped on the link.

This *does* mean there is a delay between when the user clicks on the link and when Tabzilla opens. If it's really confusing for users with slow internet connections, we could either add a throbber, or just redirect to mozilla.org (the normal behavior of the link when JS is disabled) if Tabzilla/jQuery haven't loaded within a reasonable amount of time (say, 250ms).

### DIFF
--- a/components/tabzilla.jsx
+++ b/components/tabzilla.jsx
@@ -1,0 +1,64 @@
+// This is a Tabzilla "lazy-loader" for sites that don't use jQuery.
+// Because Tabzilla requires jQuery and jQuery is relatively large,
+// we delay loading it until the last possible moment, when the user
+// has expressed intent to explore Tabzilla.
+//
+// Only one instance of this component is expected to ever be on the
+// page at once, and it's expected to exist for the lifetime of the
+// whole page.
+//
+// This component does require the following stylesheet to be loaded,
+// however:
+//
+//   //mozorg.cdn.mozilla.net/media/css/tabzilla-min.css
+
+var React = require('react');
+
+// This is currently the minimum version of jQuery that Tabzilla requires
+// to load.
+var JQUERY = '//mozorg.cdn.mozilla.net/media/js/libs/jquery-1.7.1.min.js';
+var TABZILLA = '//mozorg.cdn.mozilla.net/tabzilla/tabzilla.js';
+
+var Tabzilla = React.createClass({
+  propTypes: {
+    className: React.PropTypes.string
+  },
+  handleClick: function(e) {
+    if (!this.lazyLoadStarted) {
+      this.lazyLoadStarted = true;
+      e.preventDefault();
+      loadScript(JQUERY, function() {
+        loadScript(TABZILLA, function() {
+          window.Tabzilla.open();
+        });
+      });
+    }
+  },
+  render: function() {
+    return <a 
+      href="https://www.mozilla.org/"
+      id="tabzilla"
+      onClick={this.handleClick}
+      className={this.props.className}>mozilla</a>;
+  }
+});
+
+// This is based on Tabzilla's own jQuery-loading code.
+function loadScript(src, cb) {
+  var script = document.createElement("script");
+  if (script.readyState) {
+    script.onreadystatechange = function () {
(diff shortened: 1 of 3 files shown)


components/tabzilla.jsx
```diff
@@ -0,0 +1,64 @@
...
+// This component does require the following stylesheet to be loaded,
+// however:
+//
+//   //mozorg.cdn.mozilla.net/media/css/tabzilla-min.css
+
+var React = require('react');
+
+// This is currently the minimum version of jQuery that Tabzilla requires
+// to load.
+var JQUERY = '//mozorg.cdn.mozilla.net/media/js/libs/jquery-1.7.1.min.js';
```
toolness: I am a little bit worried about this--if Tabzilla changes its minimum jQuery version below our feet, then clicking on Tabzilla will (A) ultimately load *two* versions of jQuery and (B) might not actually open Tabzilla, because `window.Tabzilla.open` could potentially throw an exception (if it uses jQuery APIs that aren't in 1.7.1).

toolness: That said, the consequences aren't *that* dire, as a confused user could just click/tap on Tabzilla again moments later and either be taken to mozilla.org or see the pane slide down. Once the anomaly is reported to us, we can easily fix it.

Pomax: We may want to rely on https://api.jquery.com/jquery.noconflict/ here so that we can at least control what happens should there be a jQuery conflict.

toolness: Hmm, do you mean if there's a jQuery conflict with Tabzilla, or with other code that uses jQuery on the Teach site? One nice thing about Tabzilla is that if it *does* decide to load jQuery itself, it automatically noconflict's *its* version of jQuery to make sure it doesn't interfere with the one already on the page, which is why I figured I didn't need to do anything from my end.

Pomax: ah. yeah I was thinking to ensure that if tabzilla loads its own. If we use jQuery on teach, it might be worth not even bothering to load 1.7.1 ourselves. If our modern version does everything tabzilla needs (and from what I know, it needs very little) it hopefully doesn't try loading 1.7.1

toolness:
Ah well, there's too many potentially weird things with this PR and I guess nobody's complaining about slow page-load speeds, so I will just close this PR. We can revisit it if we need to optimize stuff.

Issue #577, opened 2015-04-03
Add asterisks to indicate required fields on Clubs form


hannahkane:
This helps with the usability issue identified in #540.

toolness:
So now that #540 is actually fixed, I'm wondering if we still need this?

Also, if we do still need it, I have questions about the design. For example, is just an asterisk next to the field label enough, or do should we also put a "* indicates required field" sort of message somewhere on the dialog? If so, what exactly should that message say, and where should it go? Alternatively, do we want to just put "(required)" next to the field labels instead of an asterisk?

hannahkane:
I don't think we need this now that #540 has been resolved.

PR #688, opened 2015-04-14
Validate club website, prepend http://.

### LINKED ISSUE #424: URL validation on 'website' field in club modal should be friendlier

toolness:
This fixes #424.

### DIFF
--- a/pages/clubs.jsx
+++ b/pages/clubs.jsx
@@ -219,6 +219,16 @@ var ModalRemoveYourClub = React.createClass({
   }
 });
 
+var normalizeClub = function(clubState) {
+  var state = _.extend({}, clubState);
+
+  if (state.website && !/^https?:\/\//.test(state.website)) {
+    state.website = 'http://' + state.website;
+  }
+
+  return state;
+};
+
 var validateClub = function(clubState) {
   var errors = [];
 
@@ -230,6 +240,9 @@ var validateClub = function(clubState) {
   }
   if (!clubState.location) {
     errors.push("You must provide a location for your club.");
+  }
+  if (clubState.website && !/\./.test(clubState.website)) {
+    errors.push("Your club's website must be a valid URL.");
   }
 
   return errors;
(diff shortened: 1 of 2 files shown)


pages/clubs.jsx
```diff
@@ -231,6 +241,9 @@ var validateClub = function(clubState) {
   if (!clubState.location) {
     errors.push("You must provide a location for your club.");
   }
+  if (clubState.website && !/\./.test(clubState.website)) {
```
toolness: Note that this is super basic validation; we're essentially just testing to make sure that a `.` appears somewhere in what they've typed. It's way less rigorous than [what Django uses](http://stackoverflow.com/a/7160778), which means that technically, users could submit a URL that our front-end accepts but the teach-api backend doesn't (in which case the user will just get a cryptic message saying that something went wrong and  to retry later).

test/browser/clubs-page.test.jsx
```diff
@@ -283,7 +309,7 @@ describe("ClubsPage.ModalAddOrChangeYourClub", function() {
           validationErrors: ['old error that should be removed'],
           name: 'blorpy',
           location: 'chicago',
-          website: 'http://example.org',
+          website: 'example.org',
```
toolness: This has the side effect of ensuring that submitting the form normalizes the `website` field into a proper URL.

mmmavis:
r+ :ok_hand:

mmmavis:
I'm wondering if we should add some visual hints to required fields. (asterisks? maybe?)

toolness:
Thanks!

> I'm wondering if we should add some visual hints to required fields. (asterisks? maybe?)

Yeah, that was #577, but it does need some design work as I mention in https://github.com/mozilla/teach.webmaker.org/issues/577#issuecomment-91236434... and then Hannah closed it...

mmmavis:
Let me open a new ticket and reference to [your comment](https://github.com/mozilla/teach.webmaker.org/issues/577#issuecomment-91236434)

mmmavis:
k, just entered a new ticket for that: https://github.com/mozilla/teach.webmaker.org/issues/744

PR #690, opened 2015-04-14
Fixes #480 shrink carousel images

### LINKED ISSUE #480: EVENTS: Carousel images should be smaller
screen shot 2015-03-31 at 5 12 03 pm

mmmavis:


toolness:
Looks great! I have a feeling this won't be the last we've seen of react-bootstrap, but we can always bring it back another day, so thanks for removing it from the `package.json`.

PR #646, opened 2015-04-09
Fixes #622, #624 - footer padding

mmmavis:
This fixes both
https://github.com/mozilla/teach.webmaker.org/issues/622 (Footer links have no left margin/padding) and https://github.com/mozilla/teach.webmaker.org/issues/624 (Footer links overflow their container on mobile)

Issue #622, opened 2015-04-09
Footer links have no left margin/padding


toolness:
I'm not sure if this is a duplicate but this is what it looks like on my screen:

![2015-04-09_7-43-37](https://cloud.githubusercontent.com/assets/124687/7066042/665c4eca-de8c-11e4-8956-4474e795a59c.jpg)

The links appear flush against the very edge of the page.

toolness:
Hey @mmmavis, want to work on this one?

PR #692, opened 2015-04-14
Upgrade to react 0.13 and dependencies

### LINKED ISSUE #294: Upgrade to React v0.13
It was just released [a few days ago](https://facebook.github.io/react/blog/2015/03/10/react-v0.13.html). I'm not concerned about the new features so much as ensuring that we continue to avoid deprecated APIs, ensuring that it's easy to keep up with React's bugfixes and such as it continues to mature.

It's probably best that I take this one on, as resolving this issue may involve modifying any part of the codebase.

toolness:
This fixes #294 and upgrades us to React 0.13, as well as the latest versions of `react-bootstrap`, `react-ga`, `react-router`, and `react-select`.

In general this consisted of the following changes:

* Getting rid of the HTML prettification of React's generated HTML. React now notices that there are more text nodes than before and issues a warning, which disabling prettification eliminates. This is a bummer for view-source but will make our HTML files smaller and folks can view the prettified version with goggles/devtools anyways.
* Ensuring that `getChildContext` always returns the exact same properties in tests to avoid spurious [Parent vs. Owner Context](https://gist.github.com/jsfb/0eb6e61f300a8c1b2ce7) warnings.
* Upgrading to `react-router`'s new use of [`this.context.router`](https://github.com/rackt/react-router/blob/master/UPGRADE_GUIDE.md#012x---013x) instead of the `Router.State` and `Router.Navigation` mixins. This entailed creating a new `StubRouter` class for our tests.

In general, having the really good test coverage of our front-end codebase really helped here, as it surfaced problems very quickly.


toolness:
Note that this PR will likely collide with #626. We should probably land #626 first, since it's more urgent, which means this PR is blocking on id.webmaker.org's stability.

toolness:
Hey @k88hudson, curious what you think of this, since a lot of this code is testing code, and you're familiar w/ both react and react-router.

toolness:
Ok, I think this is ready for review. @ScottDowne can you take a look?

toolness:
By the way, whoever reviews this (and also, once it is merged) may need to manually `npm remove react-a11y` before running `npm install`. We stopped using `react-a11y` a while ago but it was still in our `package.json`, and because it has a peer dependency that requires React 0.12, an error may be raised, which manually uninstalling `react-a11y` fixes.

ScottDowne:
Not sure if this is due to needing a rebase, or because of the upgrade.

The login logo is super large and it's not on staging.

logo

The footer is missing some left padding or margin:

footer

ScottDowne:
It worked otherwise, just visual quirks.

toolness:
Hmm, the footer issue you're seeing seems like a regression of #622, which was fixed by #646... but I'm not seeing that or the big Webmaker logo on my end. Can you re-try with my latest merge? If you're still experiencing those problems, what browser+version are you using?

ScottDowne:
That seems to have done the trick, looks good now.

R+

Issue #443, opened 2015-03-27
Implement the design for the accessible Clubs list


hannahkane:
Design was done in #245.

@iamjessklein - can you provide the assets?

iamjessklein:
The only additional assets here are the map icons. @toolness - should I go ahead and purchase them?

iamjessklein:
I would actually start to implement without the flags. I am having a difficult time finding a set that permits open source use. [This is the set](https://www.iconfinder.com/iconsets/flags-of-the-world-2) that I was going to get, however these restrictions in the TOS seem to be restricting the use:
![deh0c_e4mp](https://cloud.githubusercontent.com/assets/535012/6876818/97bbea38-d4a1-11e4-911a-e4960de80696.png)

However, it does say they permit usage in "On-line content" so I am going to try contacting them to see if it's permissible.

toolness:
I can bust this up. Jess and I were talking about an approach that doesn't involve using the flags, where we could riff off the [LittleBits Chapters](http://littlebits.cc/community/chapters) design:

![2015-04-09_13-03-59](https://cloud.githubusercontent.com/assets/124687/7072272/f812d07e-deb8-11e4-9fe6-f800d084e624.jpg)

Basically, instead of the flags, and instead of the three-letter city codes on the LittleBits site, we could show a unique color combined with the [2 or 3-letter country code](http://www.worldatlas.com/aatlas/ctycodes.htm) for the country that the club is in.

hannahkane:
woohoo! excited for the busting that's going to happen. :ghost:

iamjessklein:
Really nice @toolness - this is what I was visualizing. How are these colors generated? It would be nice to build off of our core palette and then move into shades.

Issue #245, opened 2015-03-11
Determine an accessible solution for the webmaker clubs map


toolness:
As @adamlofting mentioned in https://github.com/mozilla/teach.webmaker.org/issues/151#issuecomment-78243236:

> As a fallback, you can always search and see a list of local clubs without needing to see a map. The data is essential, the map presentation is the progressive enhancement.

So we need to actually implement that!

This will require some **design**, since we don't currently have any mock-ups that show what such a list of clubs might look like.

Then it will require **implementation**. It's likely that for v1 we can constrain the implementation in some ways. For instance, it's likely that there won't be a ton of webmaker clubs at first, so we probably don't need pagination, and even search might not be needed (though ordering the list in a reasonable way, such as grouping the clubs by country or geographic region, might be necessary).

Aside from that, there's also the question of *when* we progressively enhance to the map. Currently, we only really know what the map will look like on desktops, since that's all we have mockups for; we'll likely need mockups for how things should look on mobile, and how interactions should work. If we don't have time to design/implement that for v1 then we may have to fallback to the list on mobile.

hannahkane:
@toolness - do you want this to start out as a design ticket? Wondering how we push this forward.

iamjessklein:
Can this be pushed to the March 27 heartbeat?

toolness:
@hannahkane - Yeah, i think it makes sense to start out as a design ticket... and @iamjessklein I think it's cool to push it out of this heartbeat!

iamjessklein:
I'm thinking that we can give it a similar treatment to the [Contacts, Spaces and Offices map on mozilla.org](https://www.mozilla.org/en-US/contact/spaces/)
This is what it looks like in desktop:
![screen shot 2015-03-23 at 8 44 32 am](https://cloud.githubusercontent.com/assets/535012/6780163/12627e44-d139-11e4-9b5f-f6141929e7d0.png)

These are mobile views:

![screen shot 2015-03-23 at 8 43 56 am](https://cloud.githubusercontent.com/assets/535012/6780165/1beb4dec-d139-11e4-9f98-58d113833b41.png)

![screen shot 2015-03-23 at 8 44 12 am](https://cloud.githubusercontent.com/assets/535012/6780166/1d4a60a6-d139-11e4-8e1c-537b3fc278ce.png)

cc/ @toolness @amirad @chrislarry33

hannahkane:
[...]

PR #702, opened 2015-04-15
[WIP] Style accessible club list

toolness:
This is an attempt at fixing #245. So far it looks like this:

It's based on the LittleBits design mentioned in https://github.com/mozilla/teach.webmaker.org/issues/443#issuecomment-91293411, with some inspiration from the [BAM logo](http://bam.org/).

The colors are just the sidebar colors, and are based on a hash of the full location name.

### DIFF
--- a/less/pages/clubs.less
+++ b/less/pages/clubs.less
@@ -1,0 +1,74 @@
+.page.clubs main ul.clubs li {
+  margin: 0;
+  border: none;
+  padding: 1rem 0;
+}
+
+.place-square {
+  width: 7rem;
+  height: 7rem;
+  overflow: hidden;
+}
+
+.place-square > span {
+  font-size: 2.25rem;
+  margin-left: -0.5rem;
+  text-transform: uppercase;
+  color: white;
+  font-weight: 700;
+}
+
+.club .media-left, .club .media-body {
+  background: #FBF9F9;
+}
+
+.club h4 {
+  padding-top: 0.9rem;
+  margin-bottom: 0.1rem;
+  text-transform: uppercase;
+  font-weight: 600;
+  font-size: 1.4rem;
+}
+
+.club .location {
+  font-style: italic;
+  font-size: 1rem;
+}
+
+.club .leader {
+  font-size: 1.4rem;
+}
+
+.place-square.hash-color-0 {
+  background-color: @blueSidebarColor;
+}
+
+.place-square.hash-color-1 {
+  background-color: @darkBlueSidebarColor;
+}
+
+.place-square.hash-color-2 {
+  background-color: @purpleSidebarColor;
+}
+
+.place-square.hash-color-3 {
+  background-color: @tealSidebarColor;
+}
+
+.place-square.hash-color-4 {
+  background-color: @redSidebarColor;
+}
+
+.place-square.hash-color-5 {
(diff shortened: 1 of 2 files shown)


less/pages/clubs.less
```diff
@@ -0,0 +1,74 @@
+.page.clubs main ul.clubs li {
+  margin: 0;
+  border: none;
```
toolness: We need this super-specific selector so we can override the margin and border for all `li` elements, which I don't like. We should probably change the styling for the page so that `li` elements don't have orange borders and big margins by default.

pages/clubs.jsx
```diff
@@ -18,6 +18,29 @@ var ga = require('react-ga');
 
 var Illustration = require('../components/illustration.jsx');
 
+var PlaceSquare = React.createClass({
+  NUM_HASH_COLORS: 6,
```
toolness: This has to be kept in-sync with the number of `.place-square.hash-color-*` rules in the LESS, which is a bit annoying.

pages/clubs.jsx
```diff
@@ -18,6 +18,29 @@ var ga = require('react-ga');
...
+    var sum = 0;
+
+    for (var i = 0; i < loc.length; i++) {
+      sum += loc.charCodeAt(i);
+    }
+
+    return 'hash-color-' + (sum % this.NUM_HASH_COLORS);
+  },
+  render: function() {
+    var abbr = this.props.location.slice(0, 3);
```
davidascher: This won't actually generate country or city codes, just so you know.  As an example, Vancouverites proudly wear the YVR airport code.

chrislarry33: PTO was great, appreciated the otherworldlyness of all aspects

On April 16, 2015 12:03:13 AM EDT, David Ascher <notifications@github.com> wrote:
>> @@ -18,6 +18,29 @@ var ga = require('react-ga');
>>  
>>  var Illustration = require('../components/illustration.jsx');
>>  
>> +var PlaceSquare = React.createClass({
>> +  NUM_HASH_COLORS: 6,
>> +  hashColor: function() {
>> +    var loc = this.props.location;
>> +    var sum = 0;
>> +
>> +    for (var i = 0; i < loc.length; i++) {
>> +      sum += loc.charCodeAt(i);
>> +    }
>> +
>> +    return 'hash-color-' + (sum % this.NUM_HASH_COLORS);
>> +  },
>> +  render: function() {
>> +    var abbr = this.props.location.slice(0, 3);
>
>This won't actually generate country or city codes, just so you know. 
>As an example, Vancouverites proudly wear the YVR airport code.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/mozilla/teach.webmaker.org/pull/702/files#r28482699

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

toolness:
Hey @hannahkane, any idea how high-priority this is? If we want to merge it, I'll need to resolve some conflicts it's got with the latest version of the site.

hannahkane:
Hmmm....I forgot this had gotten this far! There's a ticket in the current heartbeat to revise the Clubs page anyway, so I wouldn't work on this now.

toolness:
Ok cool, I will close this PR then.

PR #715, opened 2015-04-17
Allow forms to disable HTML5 form validation for manual testing

toolness:
This is an experimental PR and I'm not sure how I feel about it yet.

A few days ago @mmmavis brought up the question of how to test to make sure that our forms worked on browsers that don't support HTML5 form validation. I told her that she could use an iPhone or iPad, since Mobile Safari is the only remaining browser whose latest version still doesn't support it, but obviously this solution is suboptimal, especially for developers who don't have iOS devices on-hand.

Then I learned about the nifty [`novalidate` attribute](http://www.wufoo.com/html5/attributes/11-novalidate.html) for form elements.

This PR adds a `HTML5_NOVALIDATE` environment variable that, if set to `on`, hacks `React.createElement` to automatically add the `novalidate` attribute to all form elements generated by React code (on both the client and server/static-side).

Another option is to just point developers to http://novalidate.com/, a bookmarklet that does roughly the same thing. Or we could just tell devs to temporarily add `novalidate` to the relevant form while testing it. Or we could add a checkbox to the developer ribbon modal that allows anyone to dynamically change the value at runtime, which could make manual testing by QA folks easier. The possibilities are endless!

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,10 @@ JS disabled, just add `safemode=on` to the querystring when visiting the
 page. This will cause the page to operate in "safe mode", meaning that
 almost no JS will run, even if JS is enabled in the browser.
 
+If you need to make sure that a form is usable with browsers that
+don't support HTML5 form validation, see the documentation for the
+`HTML5_NOVALIDATE` environment variable below.
+
 ## Generating A Static Site
 
 Run `npm run build` to generate a static site in `dist/` that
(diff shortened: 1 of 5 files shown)


toolness:
This is way too hacky and complicated. I've opted to just add a section to `CONTRIBUTING.md` instead, in 69f398ba22861ed2a4e525a9257743a9bcdda070.

PR #721, opened 2015-04-17
Fixes #550 - add 3 CTAs to bottom of the Activities page

### LINKED ISSUE #550: Change CTAs on Teaching Activities page
Right now this is one of the only pages that doesn't have the three CTAs at the bottom. Wondering if we should replace the single Clubs CTA with three:

1) Clubs
2) Teach Like Mozilla
3) Web Literacy

@LauraHilliger @ldecoursy @iamjessklein - what do you think?

mmmavis:
This fixes #550.


ScottDowne:
I love how easy reusable react components make reviews like this :D

Issue #722, opened 2015-04-19
Not all of our form labels have the "for" attribute


toolness:
I don't believe our form labels have the `for` attribute that associates them with their relevant input fields. The main reason for this is because I wasn't sure how to create an `id` for the input fields that was guaranteed to be globally unique, but we might just want to have some kind of id prefix get passed in by the component's parent for that or something.

PR #729, opened 2015-04-20
Link labels to form fields in add/change club modal.

toolness:
This helps with #722. Note that the "where are you located" field is not currently hooked up, because `react-select` doesn't seem to support a11y at all, which is a big bummer.

Also note that this adds an optional **`idPrefix`** property to the modal component, which determines what the ultimate `id` for each form field is. While this isn't strictly needed for these forms, since they're singleton modals, I thought it would be good to follow the convention of making them configurable by the parent component, since `id` attributes are in a global namespace and therefore parent components should be able to tweak them if needed.


toolness:

mmmavis:
r+, @toolness could you please rebase :)

toolness:
Ah, forsaken #688!

Ok done :)

PR #762, opened 2015-04-21
Rename /mozilla-web-clubs/ to /clubs/.

### LINKED ISSUE #750: Rename /mozilla-web-clubs/ to /clubs/
We discussed this in another issue and decided it was a good idea, particularly since we'd like it to be a permalink yet the names of clubs in flux (it was "mozilla web clubs", then just "mozilla clubs", etc).

Most importantly, though, in #716 we include a link to this page in the email we send out to users when they've added a club. **The email uses the /clubs/ URL** so we really need the page to exist there. (And if we decide on a different URL, we need to update the email text to refer to that URL instead.)

toolness:
This fixes #750.

Note that the name of the route, used internally by the code, is still `mozilla-web-clubs`, which allows existing links (currently in sidebar, activities and clubs toolkit pages) to still point to the proper place.

@mmmavis can you review this?

### DIFF
--- a/lib/routes.jsx
+++ b/lib/routes.jsx
@@ -25,7 +25,7 @@ var routes = (
      handler={require('../pages/events.jsx')}/>
     <Route name="event-resources" path="/events/resources/"
      handler={require('../pages/event-resources.jsx')}/>
-    <Route name="mozilla-web-clubs" path="/mozilla-web-clubs/"
+    <Route name="mozilla-web-clubs" path="/clubs/"
      handler={require('../pages/clubs.jsx')}/>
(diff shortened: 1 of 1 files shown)


toolness:
I've done some digging into why Travis is complaining at https://github.com/mozilla/teach.webmaker.org/pull/770#issuecomment-94865623.

ScottDowne:
Looks good to me.

ScottDowne:
Ignoring travis for now because it's known, we should be able to just merge this, but I'll let mavis or atul do the final merge in case I'm the only one that feeld comfortable with merging while travis is failing :P

PR #772, opened 2015-04-21
downgrade react select module because breakage

ScottDowne:


ScottDowne:
Fixed in https://github.com/mozilla/teach.webmaker.org/commit/af25115156a69c1d62bc203ef1b29a411712a305

toolness:
Ah this is basically a duplicate of af25115156a69c1d62bc203ef1b29a411712a305, except my version uses 0.4.6, which is the latest working one. Sorry about the duplication of work!

PR #779, opened 2015-04-22
Support unique per-page titles.

### LINKED ISSUE #774: Pages on the site should have their own <title>
I was listening to [Accessibility Easy Checks](http://rosenfeldmedia.com/a-web-for-everyone/accessibility-easy-checks-with-sharron-rush/) and one of the most low-hanging fruit that can really help out vision-impaired users is simply ensuring that the `<title>` tag is correct:

> **Whitney:** So, let’s pick one of them apart, and talk about how it works and why it’s important. The first one on the list is ‘Page Title.’ It seems pretty basic, doesn’t it?
>
> **Sharron:** One of the reasons that we put that as the first one is because it’s relatively easy to check, and it’s the first thing that you often encounter when you come to a website. We thought, “We’ll start with the page title. Does it have a page title, does it not”? And also relatively easy to understand the importance of, because **people who are listening to the Web need that for orientation**, to understand where they are, “Am I on the right page? Am I where I thought I was going to be”? If the page title is announced and it’s clear, and gives that information, then they have success.

And we don't actually change the page title based on the page content! Every single page is just called "Mozilla Learning". So we should fix that.

toolness:
This fixes #774.

### DIFF
--- a/components/page.jsx
+++ b/components/page.jsx
@@ -12,6 +12,20 @@ var DevRibbon = (process.env.NODE_ENV === 'production' &&
                 : require('./dev-ribbon.jsx');
 
 var Page = React.createClass({
+  statics: {
+    handlerForPage: function(router, url) {
+      return router.match(url).routes[1].handler;
+    },
+    titleForHandler: function(handler) {
+      var title = 'Mozilla Learning';
+
+      if (handler.pageTitle) {
+        title = handler.pageTitle + ' - ' + title;
+      }
+
+      return title;
+    }
+  },
   contextTypes: {
     router: React.PropTypes.func
   },
@@ -19,6 +33,9 @@ var Page = React.createClass({
     showModal: React.PropTypes.func.isRequired,
     hideModal: React.PropTypes.func.isRequired,
     teachAPI: React.PropTypes.object.isRequired
+  },
+  getCurrentPageHandler: function() {
+    return this.context.router.getCurrentRoutes()[1].handler;
   },
   getInitialState: function() {
     return {
@@ -38,6 +55,7 @@ var Page = React.createClass({
     } else if (!this.state.modalClass && prevState.modalClass) {
       document.body.classList.remove('modal-open');
     }
+    document.title = Page.titleForHandler(this.getCurrentPageHandler());
   },
   getTeachAPI: function() {
     if (!this.teachAPI) {
@@ -63,8 +81,7 @@ var Page = React.createClass({
     }
   },
   render: function() {
-    var routes = this.context.router.getCurrentRoutes();
-    var pageClassName = routes[1].handler.pageClassName || '';
+    var pageClassName = this.getCurrentPageHandler().pageClassName || '';
     return (
       <div>
         <div className={"page container-fluid " + pageClassName}
--- a/components/page.jsx
+++ b/components/page.jsx
@@ -10,6 +10,7 @@ var DevRibbon = (process.env.NODE_ENV === 'production' &&
                  process.env.SHOW_DEV_RIBBON !== 'on')
                 ? null
                 : require('./dev-ribbon.jsx');
+var config = require('../lib/config');
 
 var Page = React.createClass({
   statics: {
@@ -48,6 +49,15 @@ var Page = React.createClass({
   },
   hideModal: function() {
     this.setState({modalClass: null, modalProps: null});
+  },
+  componentDidMount: function() {
+    if (process.env.NODE_ENV !== 'production' && !config.IN_TEST_SUITE) {
+      var title = Page.titleForHandler(this.getCurrentPageHandler());
+      if (document.title !== title) {
+        console.warn("Document title is '" + document.title + "' but " +
+                     "expected it to be '" + title + "'.");
+      }
+    }
   },
   componentDidUpdate: function(prevProps, prevState) {
     if (this.state.modalClass && !prevState.modalClass) {
--- a/test/browser/page.test.jsx
+++ b/test/browser/page.test.jsx
@@ -14,29 +14,57 @@ var FakeModal = React.createClass({
 });
 
 describe("page", function() {
-  var handler, page, xhr;
+  var handler, page, xhr, originalTitle;
 
-  beforeEach(function(done) {
-    xhr = sinon.useFakeXMLHttpRequest();
-    Router.run(routes.routes, '/', function(Handler) {
+  function visitPage(url, cb) {
+    Router.run(routes.routes, url, function(Handler) {
       handler = TestUtils.renderIntoDocument(<Handler/>);
       page = TestUtils.findAllInRenderedTree(handler, function(c) {
         return !!c.showModal;
       })[0];
+      cb();
+    });
+  }
+
+  beforeEach(function() {
+    xhr = sinon.useFakeXMLHttpRequest();
+    originalTitle = document.title;
+    handler = null;
+    page = null;
+  });
+
+  afterEach(function() {
+    if (handler) {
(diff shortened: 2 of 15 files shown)


toolness:
I noticed that the join page doesn't need to exist anymore, so I didn't add a page title to it, and instead filed #782 to get rid of it separately.

toolness:

components/page.jsx
```diff
@@ -12,6 +12,20 @@ var DevRibbon = (process.env.NODE_ENV === 'production' &&
...
 var Page = React.createClass({
+  statics: {
+    handlerForPage: function(router, url) {
+      return router.match(url).routes[1].handler;
+    },
+    titleForHandler: function(handler) {
+      var title = 'Mozilla Learning';
+
+      if (handler.pageTitle) {
+        title = handler.pageTitle + ' - ' + title;
```
mmmavis: I'm not sure if the convention is to put general site title first or title of the subpage.  But Webmaker seems to be doing this the same way so I guess this is the preferred way.

mmmavis:
@toolness , fyi, I'm reading through docs on `Context` as I have no idea what it is so this review will take a big longer to finish.

components/page.jsx
toolness: Yeah, I think it's the preferred way b/c that way if the text gets cut off, e.g. how long titles get cut off on browser tabs, the text that makes the tab "different" from other tabs on the same site is visible.

toolness:
Oh yeah, sorry I didn't explain that--it is confusing b/c it's an undocumented feature of React but they say that they will eventually document it, so tools like `react-router` are already using it. This is where I learned about them:

https://www.tildedave.com/2014/11/15/introduction-to-contexts-in-react-js.html

Also, we use them for a few things ourselves:

* The modal manager mixin at [`mixins/modal-manager.js`](https://github.com/mozilla/teach.webmaker.org/blob/develop/mixins/modal-manager.js) accesses context information provided by [`components/page.jsx`](https://github.com/mozilla/teach.webmaker.org/blob/develop/components/page.jsx).
* The teach API client mixin at [`mixins/teach-api-client.js`](https://github.com/mozilla/teach.webmaker.org/blob/develop/mixins/teach-api-client.js) also accesses context information provided by [`components/page.jsx`](https://github.com/mozilla/teach.webmaker.org/blob/develop/components/page.jsx).

Let me know if you have any questions about anything!

mmmavis:
> https://www.tildedave.com/2014/11/15/introduction-to-contexts-in-react-js.html

heh, this is exactly what I've been reading :smile_cat:

components/page.jsx
```diff
@@ -10,8 +10,23 @@ var DevRibbon = (process.env.NODE_ENV === 'production' &&
                  process.env.SHOW_DEV_RIBBON !== 'on')
                 ? null
                 : require('./dev-ribbon.jsx');
+var config = require('../lib/config');
 
 var Page = React.createClass({
+  statics: {
+    handlerForPage: function(router, url) {
+      return router.match(url).routes[1].handler;
```
mmmavis: hey @toolness can you point me to the docs for `.match` method?  I could probably guess what that is for but I wanted to read through the docs to get a more clear sense.  I couldn't find it anywhere on https://github.com/rackt/react-router/tree/master/docs :confused:

components/page.jsx
mmmavis: @toolness I don't understand how `handler.pageTitle` is hooked up with w/e `pageTitle` is set on the actul page, e.g., https://github.com/mozilla/teach.webmaker.org/pull/779/files#diff-f61a39ec11b5b6b336c2b6fdc3a59c4bR76.  Guidance?  :cold_sweat: :wink:

mmmavis: > @toolness I don't understand how `handler.pageTitle` is hooked up with w/e `pageTitle` is set on the actul page, e.g., https://github.com/mozilla/teach.webmaker.org/pull/779/files#diff-f61a39ec11b5b6b336c2b6fdc3a59c4bR76.  Guidance?  :cold_sweat: :wink: 

nvm I think I got it!  it's via `<Route name="about" path="/about/"
     handler={require('../pages/about.jsx')}/>` in `routes.jsx` :bowtie:

components/page.jsx
toolness: Ah yeah unfortunately i actually had to dig into the source code to find that--it was really hard trying to figure out how to access route information outside of a component!

Two options:

1. We can submit docs for that method to react-router, which will hopefully make it a "public" method rather than potentially a private implementation detail.
2. We just use it and an exception will be thrown or a warning logged if we upgrade react-router and things break.

This message was typed with haste on a tiny keyboard with no tactile feedback.

components/page.jsx
toolness: Yep you got it :)

components/page.jsx
```diff
@@ -10,8 +10,23 @@ var DevRibbon = (process.env.NODE_ENV === 'production' &&
...
+
+      if (handler.pageTitle) {
+        title = handler.pageTitle + ' - ' + title;
+      }
+
+      return title;
+    }
+  },
   contextTypes: {
     router: React.PropTypes.func
```
mmmavis: wondering if this should be `required` as well?

toolness: You mean the `router`? I think that all `contextTypes` are required by default--at least it seems like React logs warnings if they're not present when the component is instantiated...

mmmavis: Just did some experiment, yea @toolness you're right!  cooool I got confused 'cuz https://www.tildedave.com/2014/11/15/introduction-to-contexts-in-react-js.html has examples w/ and w/o `isRequired` for `contextTypes` props

test/browser/page.test.jsx
```diff
@@ -14,29 +14,57 @@ var FakeModal = React.createClass({
...
-  var handler, page, xhr;
+  var handler, page, xhr, originalTitle;
 
-  beforeEach(function(done) {
-    xhr = sinon.useFakeXMLHttpRequest();
-    Router.run(routes.routes, '/', function(Handler) {
+  function visitPage(url, cb) {
+    Router.run(routes.routes, url, function(Handler) {
       handler = TestUtils.renderIntoDocument(<Handler/>);
       page = TestUtils.findAllInRenderedTree(handler, function(c) {
```
mmmavis: Hi @toolness ,
Do you mind telling me what `findAllInRenderedTree` does?  The [official doc](https://facebook.github.io/react/docs/test-utils.html#findallinrenderedtree) doesn't help much...

toolness: Ah yeah, it's a bit confusing, as is my use of it here... Basically I was just trying to find the `<Page>` component in the heirarchy of rendered components that the router is giving us, but it is surprisingly hard to do that.

`findAllInRenderedTree` essentially craws through the entire component heirarchy, calling our test function, which returns `true` if the given component is one of the ones we're looking for. Since our `<Page>` component has a method called `showModal`, and it's the only component that has that method, we basically just test for that... It's not a great way of finding the `<Page>` component but I couldn't figure out anything else at the time. :/

mmmavis: Ah, got it!  :sunglasses:  

Hopefully I will be able to find some useful methods to use as I'm getting familiar with the test framework and API!

mmmavis:
Hey @toolness , sorry for taking so long to review this PR :turtle: :turtle: :turtle:  I'm done and it's a r+ :+1: 
:dancers: :dancers: :dancers: :dancers: :dancers:

toolness:
No worries mavis! Thanks for the great questions--they also required me to explain my own thought process, which is always a good thing. :)

Issue #784, opened 2015-04-22
Clubs toolkit page is no longer keyboard-accessible


toolness:
Hmm, so @ScottDowne's [original solution](https://github.com/mozilla/teach.webmaker.org/commit/f0981c96352f2da5031bd865f38f70d50a17b8b7#commitcomment-10569879) to make the Clubs toolkit sections collapsible without JS was actually dependent on the toolkit sections not having any focusable elements inside them. Unfortunately, links were added in #683, which breaks this.

You can see the problem if you use solely the tab key to navigate through the links on the page. When one of the collapsible section headers is focused, that section is expanded, but as soon as you tab to one of the hyperlinks *inside* that expanded section, the section is collapsed, preventing you from actually navigating to the link. :(

toolness:
I should note that the Web Literacy page is, I think, implemented using the same mechanism--but it's still working fine for keyboards because AFAIK there are no hyperlinks or other focusable elements inside the collapsible sections. If that ever changes, however, it'll have the same problem as the Clubs Toolkit page.

ScottDowne:
Interesting. I thought because all the contents including the links were in a button that is focused, it would work.

Example, you can select text and it stays open.

I think we can fix this with some css.

ScottDowne:
Hm, maybe cannot do this in css.

toolness:
Yeah, I think until CSS supports that funky thing where you can style an ancestor based on its descendants, this may be impossible to do with CSS >.<

ScottDowne:
We can make this a js thing and just open them all if js disabled. It was an interesting experiment but I think it's more of a hack now than useful.

It would still be tabbable with js, and if you tab to a child anchor, it would use js to set a class on the parent to keep it open.

toolness:
Rad!

We could potentially use `react-bootstrap`, although I don't know how awesome its collapsible/accordion component is when JS is disabled.

toolness:
This was fixed by #788!

Issue #698, opened 2015-04-15
Expander component doesn't have unit tests


toolness:
I noticed that `components/expander.jsx` doesn't seem to have any unit tests, which is concerning because there are JS event handlers in it.

@ScottDowne would you be interested in writing tests for this? I think you could probably use the [sidebar tests](https://github.com/mozilla/teach.webmaker.org/blob/develop/test/browser/sidebar.test.jsx) as a jumping-off point.

ScottDowne:
Yeah, I should do this. It'll be a p2 for me though.

ScottDowne:
Is the idea we just test cs related things? And not css/html related things?

If that's the case, I think this only has one js interaction, and it's an enhancement and not required. So this should be a good starting point for my test making experience. (It's been a long time)

toolness:
Rad!

Yeah, I think testing css/html is hard for various reasons, so just JS stuff would be great.

toolness:
Hey @mmmavis, it looks like Scott's pretty busy with other stuff, would you be interested in taking this one on, to familiarize yourself with unit testing?

mmmavis:
YEAH! :metal:

PR #788, opened 2015-04-22
fix expander a bit

ScottDowne:


toolness:
Er what things in particular is this supposed to fix? What functionality should I be verifying?

ScottDowne:
Ah, sorry, it fixes: #784

toolness:
aww the branch name :(

ScottDowne:
haha, yeah

toolness:
Hey want to take this opportunity to add unit tests to the expander too, fixing #698? :grin:

ScottDowne:
I'm not super familiar with the test suite, so adding tests would not be quick.

toolness:
o ok. :(

This looks good. I verified that everything is auto-expanded when JS is disabled, and both collapse/expend and keyboard accessibility work great when JS is enabled.

Out of curiosity, is there a reason you're calling the expand/collapse caret "ion"? It's kinda hard to search for since the word "function" includes it :)

Anyhow, thanks for the fix, merging now!

ScottDowne:
ion is from http://ionicons.com/

Same icon used in webmaker app and the design team was moving towards them

ionicons creates the class.

That said, we could also add our own class along with ion, and when we reference it in css use our own class.

toolness:
ohhh i see! got it, no worries.

PR #795, opened 2015-04-22
Fixes #782 - removed join.jsx

### LINKED ISSUE #782: Get rid of /pages/join.jsx
It looks like this is a placeholder page that isn't linked-to from anywhere and was obsoleted according to #154 by our use of id.webmaker.org, so we should get rid of it.

mmmavis:
This fixes #782


toolness:
Thanks mavis!

Issue #714, opened 2015-04-17
/events/resources/ fails to load ionicons.ttf and iconicons.woff


toolness:
When loading `/events/resources/`, this shows up in my browser console:

![2015-04-17_6-53-37](https://cloud.githubusercontent.com/assets/124687/7200803/88a1795e-e4ce-11e4-8450-afbf04a09e91.jpg)

Note that this is logged *only* when loading `/events/resources/`, other pages load without raising the warning.

It looks like this has probably been happening since #396.

toolness:
It looks like this is happening because we're inlining `ionicons.css` into our CSS bundle. However, that file has a `@font-face` rule that points to fonts with relative URLs that don't exist in our `dist` directory. We then have a `@font-face` rule in our own LESS that points to the correct files.

I'm not sure why our `@font-face` rule isn't overriding the one in `ionicons.css` entirely... it does appear after the original rule in the generated CSS, which I *think* means it should override them.

Regardless, I am a bit concerned about the "override" approach because if we upgrade ionicons and it changes, those changes might no longer be fully reflected by our override rule.

Anyways, one way to fix this is to simply directly link to the ionicons CSS in our HTML pages, which we're already doing with bootstrap and tabzilla.

PR #809, opened 2015-04-23
Spider for broken links during npm test.

### LINKED ISSUE #804: Add a test that checks for broken links
It shouldn't be too hard to use a library like [simplecrawler](https://www.npmjs.com/package/simplecrawler) to crawl the site and ensure that, at the very least, all internal links (i.e. links that don't go to external sites) are valid. This could prevent issues like #603 and #714, and if we crawl `srcset` attributes it can also prevent ones like #664.

Before deploying to production, we could also automatically check to ensure that no links are going to `fixme` (#500, #597).

(Originally this issue also included searching for huge images, but that has been moved to #810).

toolness:
This fixes #804. Example output on error:

On success:

### DIFF
--- a/less/index.less
+++ b/less/index.less
@@ -4,6 +4,3 @@
 @import "components/index";
 @import "pages/index";
 @import "../node_modules/react-select/less/default.less";
-@import (less) "../node_modules/webmaker-app-icons/css/ionicons.css";
-
-@font-face { font-family: "Ionicons"; src: url("vendor/webmaker-app-icons/fonts/ionicons.eot?v=2.0.1"); src: url("vendor/webmaker-app-icons/fonts/ionicons.eot?v=2.0.1#iefix") format("embedded-opentype"), url("vendor/webmaker-app-icons/fonts/ionicons.ttf?v=2.0.1") format("truetype"), url("vendor/webmaker-app-icons/fonts/ionicons.woff?v=2.0.1") format("woff"), url("vendor/webmaker-app-icons/fonts/ionicons.svg?v=2.0.1#Ionicons") format("svg"); font-weight: normal; font-style: normal; }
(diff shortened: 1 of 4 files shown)


toolness:
Travis is going to fail because the spidering actually finds broken links, causing `npm test` to fail. :)

Specifically, it's running into #714.

toolness:
Unfortunately, the spidering library we're using doesn't seem to parse `srcset` tags, so not all of our responsive images will be checked, which is a bummer. 

Still, this is better than nothing right now.

toolness:
Ok, this now fixes #714 using the suggestion described in https://github.com/mozilla/teach.webmaker.org/issues/714#issuecomment-95921298, since the spidering tests can't actually pass unless there are no 404s.

toolness:
@ScottDowne can you take a look at this?

less/index.less
```diff
@@ -4,6 +4,3 @@
 @import "components/index";
 @import "pages/index";
 @import "../node_modules/react-select/less/default.less";
-@import (less) "../node_modules/webmaker-app-icons/css/ionicons.css";
-
-@font-face { font-family: "Ionicons"; src: url("vendor/webmaker-app-icons/fonts/ionicons.eot?v=2.0.1"); src: url("vendor/webmaker-app-icons/fonts/ionicons.eot?v=2.0.1#iefix") format("embedded-opentype"), url("vendor/webmaker-app-icons/fonts/ionicons.ttf?v=2.0.1") format("truetype"), url("vendor/webmaker-app-icons/fonts/ionicons.woff?v=2.0.1") format("woff"), url("vendor/webmaker-app-icons/fonts/ionicons.svg?v=2.0.1#Ionicons") format("svg"); font-weight: normal; font-style: normal; }
```
ScottDowne: Interesting, I thought we needed this for the fonts, but turns out we do not?

toolness: Yep see https://github.com/mozilla/teach.webmaker.org/issues/714#issuecomment-95921298 for more explanation.

ScottDowne: I think I see why it still works! Yay!

PR #814, opened 2015-04-24
Fixes #811 - updated 'legal' link in footer

### LINKED ISSUE #811: Change "Legal" link in footer
We've just heard from the legal dept that we should be linking to https://webmaker.org/en-US/terms instead.

mmmavis:
Fixes #811

### DIFF
--- a/components/footer.jsx
+++ b/components/footer.jsx
@@ -17,8 +17,8 @@ var Footer = React.createClass({
             </div>
             <div className="col-xs-6">
               <ul className="list-unstyled">
-                <li><a href="https://www.mozilla.org/en-US/about/legal/terms/mozilla">Legal</a></li>
-                <li><a href="https://www.mozilla.org/en-US/privacy/websites">Privacy</a></li>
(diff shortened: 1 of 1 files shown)


components/footer.jsx
```diff
@@ -17,8 +17,8 @@ var Footer = React.createClass({
             </div>
             <div className="col-xs-6">
               <ul className="list-unstyled">
-                <li><a href="https://www.mozilla.org/en-US/about/legal/terms/mozilla">Legal</a></li>
-                <li><a href="https://www.mozilla.org/en-US/privacy/websites">Privacy</a></li>
+                <li><a href="https://webmaker.org/terms">Legal</a></li>
+                <li><a href="https://www.mozilla.org/privacy/websites">Privacy</a></li>
```
mmmavis: Spec says use `https://webmaker.org/en-US/terms ` but I think we shouldn't specify locale in the urls.

toolness: Good call!

toolness:
This looks good, thanks!

PR #839, opened 2015-04-29
Use webpack to (re)generate index HTML files

toolness:
While researching #585 I ran into jlongster's [Backend Apps with Webpack](http://jlongster.com/Backend-Apps-with-Webpack--Part-I) series of posts and decided to familiarize myself with using webpack in node (as opposed to the browser) by using it to watch and rebuild our static index files during `gulp watch`.

This has a number of benefits:

* **Rebuilding index HTML files is a *lot* faster.** We don't have to spawn a subprocess anymore, and webpack rebuilds only the code that's changed.
* Index HTML files are only rebuilt when an actual file they use changes. Currently, for example, changing `lib/travis.js` will trigger a full rebuild of the index HTML files, even though that file has nothing to do with rebuilding index HTML files.
* We can reuse more of our webpack code pipeline in node, rather than relying on a completely separate build toolchain (in the form of `node-jsx`). ~~We haven't completely decoupled ourselves from that toolchain yet, as everything outside of `gulp watch` still uses it, but the plan is to migrate away from it, so that our build toolchain is ultimately simplified.~~ ~~We're still using `node-jsx` but only for running node-side tests.~~ This should also make migrating to #585 (and quickly iterating on the server during development) easier.

Limitations:

* We're not rebuilding the `sitemap.xml` file in `gulp watch` anymore. Hopefully not that big a deal.
* ~~Until we've completely replaced our use of `node-jsx` with webpack, we're moving away from [dev/prod parity](http://12factor.net/dev-prod-parity), which isn't awesome. So we should finish that migration ASAP.~~
* Running the `test/index-static.test.js` test is slower because it builds the webpack bundle now, but I'd rather it just run slower than assume the developer manually builds it before running the test, and then have devs get frustrated when they change the code but the tests still fail.
* ~~We're building the index HTML generator bundle in `dist/index-static`, which probably isn't awesome because it's really just an intermediate file that isn't intended for distribution. As such, we should probably put it in some sort of intermediate directory that's also in our `.gitignore`.~~ The index HTML generator bundle is in the `build` directory, intended for intermediate build files rather than files to be distributed.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -186,7 +186,7 @@ gulp.task('test-react-warnings', function() {
         this.emit('error', new Error('At least one warning was logged.'));
       }
     })
-    .pipe(gulp.dest('./dist'));
+    .on('data', function() {});
 });
 
 gulp.task('generate-index-files', function() {
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -18,8 +18,6 @@ var jshint = require('gulp-jshint');
 var autoprefixer = require('gulp-autoprefixer');
 var rename = require('gulp-rename');
 
-require('node-jsx').install({ extension: '.jsx' });
-
 // TODO: Some of our third-party components are triggering warnings
 // from react-a11y, so we need to disable it for now to prevent
 // warning spam. Hopefully in the future we can find a way to
@@ -75,12 +73,35 @@ function handleError() {
   });
 }
 
+function createIndexFileStream() {
+  var meta = {};
+  var execSync = require('child_process').execSync;
+
+  try {
+    meta['git-rev'] = execSync('git rev-parse HEAD', {
+      cwd: __dirname,
+      encoding: 'utf8'
+    }).slice(0, 40);
+  } catch (e) {}
+
+  return new IndexFileStream(indexStaticWatcher.getBundle(), {
+    meta: meta
+  });
+}
+
 gulp.task('sitemap', ['generate-index-files'], function() {
   gulp.src('dist/**/*.html')
     .pipe(sitemap({
       siteUrl: config.ORIGIN
     }))
     .pipe(gulp.dest('./dist'));
+});
+
+gulp.task('compile-index-static-bundle', function(cb) {
+  indexStaticWatcher.build(function ok() { cb(); }, function fail() {
+    cb(new Error('compiling ' + indexStaticWatcher.outputFilename +
+                 ' failed'));
+  });
 });
 
 gulp.task('copy-test-dirs', function() {
@@ -149,7 +170,9 @@ gulp.task('smoketest', BUILD_TASKS.concat([
   gutil.log(gutil.colors.green.bold('Yay, smoke test passes!'));
 });
 
-gulp.task('test-react-warnings', function() {
+gulp.task('test-react-warnings', [
+  'compile-index-static-bundle'
+], function() {
   var oldWarn = console.warn;
   var warnings = 0;
 
@@ -158,7 +181,7 @@ gulp.task('test-react-warnings', function() {
     gutil.log(gutil.colors.red.bold(message));
   };
 
-  return new IndexFileStream(require('./lib/index-static.jsx'))
+  return createIndexFileStream()
     .on('end', function() {
       console.warn = oldWarn;
       if (warnings) {
@@ -168,21 +191,10 @@ gulp.task('test-react-warnings', function() {
     .pipe(gulp.dest('./dist'));
 });
 
-gulp.task('generate-index-files', function() {
-  var meta = {};
-  var execSync = require('child_process').execSync;
-
-  try {
-    meta['git-rev'] = execSync('git rev-parse HEAD', {
-      cwd: __dirname,
-      encoding: 'utf8'
-    }).slice(0, 40);
-  } catch (e) {}
-
-  return new IndexFileStream(require('./lib/index-static.jsx'), {
-    meta: meta
-  })
-    .pipe(gulp.dest('./dist'));
+gulp.task('generate-index-files', [
+  'compile-index-static-bundle'
+], function() {
+  return createIndexFileStream().pipe(gulp.dest('./dist'));
 });
 
 gulp.task('beautify', function () {
@@ -217,8 +229,8 @@ gulp.task('watch', _.without(BUILD_TASKS, 'webpack'), function() {
     }, webpackConfig)))
     .pipe(gulp.dest('./dist'));
 
-  indexStaticWatcher.watch(200, function(indexStatic) {
-    new IndexFileStream(indexStatic, {})
+  indexStaticWatcher.watch(200, function() {
+    createIndexFileStream()
       .on('error', function(err) {
         gutil.log('Error rebuilding index HTML files.');
         gutil.log(gutil.colors.red.bold(err.stack));
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -28,11 +28,15 @@ require('node-jsx').install({ extension: '.jsx' });
 // require('react-a11y')();
 
 var IndexFileStream = require('./lib/gulp-index-file-stream');
-var indexStaticWatcher = require('./lib/index-static-watcher');
 var webpackConfig = require('./webpack.config');
 var config = require('./lib/config');
 var travis = require('./lib/travis');
 var server = require('./test/browser/server');
+var indexStaticWatcher = require('./lib/index-static-watcher')({
+  nodeModulesDir: path.join(__dirname, 'node_modules'),
+  outputDir: path.join(__dirname, 'dist', 'index-static'),
+  externalModules: ['react/addons']
+});
 
 var BUILD_TASKS = [
   'beautify',
@@ -213,12 +217,7 @@ gulp.task('watch', _.without(BUILD_TASKS, 'webpack'), function() {
     }, webpackConfig)))
     .pipe(gulp.dest('./dist'));
 
-  indexStaticWatcher({
-    delay: 200,
-    nodeModulesDir: path.join(__dirname, 'node_modules'),
-    outputDir: path.join(__dirname, 'dist', 'index-static'),
-    externalModules: ['react/addons']
-  }, function(indexStatic) {
+  indexStaticWatcher.watch(200, function(indexStatic) {
     new IndexFileStream(indexStatic, {})
       .on('error', function(err) {
         gutil.log('Error rebuilding index HTML files.');
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -18,7 +18,7 @@ var jshint = require('gulp-jshint');
 var autoprefixer = require('gulp-autoprefixer');
 var rename = require('gulp-rename');
 
-require('node-jsx').install();
+require('node-jsx').install({ extension: '.jsx' });
 
 // TODO: Some of our third-party components are triggering warnings
 // from react-a11y, so we need to disable it for now to prevent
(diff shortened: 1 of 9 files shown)


toolness:
@mmmavis do you want to take a look at this one? It's a bit of a doozy so no worries if not, but I think you could learn a lot from it. This stuff is also pretty new to me, so I could use a second opinion :)

If you accept the mission, I recommend trying out the following to manually test things:
* While running `gulp watch`, intentionally add syntax and runtime errors in various places in the code. This is to simulate real developers making real editing mistakes. Then make sure the console logs useful errors!
* While running `gulp watch`, add basic edits in various parts of the code, and make sure the site regerates itself properly.
* Run `gulp` without this PR, make a copy of the `dist` directory, then run `gulp` with this PR, and make sure the `index.html` files in the two `dist` directories are identical.

gulpfile.js
```diff
@@ -70,6 +70,32 @@ function handleError() {
   });
 }
 
+function createIndexFileStream() {
+  var stream = new PassThrough({ objectMode: true });
```
toolness: This returns an `IndexFileStream` of all the `index.html` files in the site. It does a bit of trickery because we first need to generate the webpack bundle that *generates* the index files.

mmmavis:
@toolness I'll get to this PR today - hopefully I won't take too long this time :no_mouth:

toolness:
No worries! It's not a big rush either, so feel free to take your time and ask any questions that come up.

mmmavis:
Sorry for the delay - my other tickets took longer than I expected.  Will work on this one today :grin:

lib/index-static-watcher.js
```diff
@@ -0,0 +1,120 @@
...
+  var outputDir = options.outputDir;
+  var outputFilename = 'index-static.bundle.js';
+  var indexStatic = null;
+  var buildInProgress = false;
+  var entry = path.join(__dirname, 'index-static.jsx');
+  var compiler = webpack({
+    entry: entry,
+    target: 'node',
+    devtool: 'sourcemap',
+    externals: function(context, request, callback) {
```
mmmavis: Hey @toolness could you help me understand what `context` and `request` are?  I see they being mentioned on http://webpack.github.io/docs/configuration.html#externals but the doc doesn't go in detail.

toolness: Ah yeah, this was annoying for me to figure out, because webpack's documentation is sparse in some places. :( I think I literally just `console.log`'d the arguments to get an idea of what they actually were, lol. Not a very strong sign of forwards-compatibility, but the good news is that at least we'll know when it breaks, since that code is executed so often!

mmmavis: heh, `console.log` is always good!  I think it'll help a lot for understanding webpack related stuff especially... :no_mouth:

toolness: Yeah, webpack can be really confusing. :( Hopefully it will improve with age...

lib/index-static-watcher.js
```diff
@@ -0,0 +1,120 @@
...
+      buildInProgress = false;
+      cb(null, indexStatic);
+    };
+    if (indexStatic) {
+      return process.nextTick(success);
+    } else if (buildInProgress) {
+      return setTimeout(self.build.bind(null, cb), 100);
+    }
+    buildInProgress = true;
+    compiler.run(onBuild.bind(null, success, function failure() {
```
mmmavis: I'm lost again.  Is there doc for the `run` method?  I can't find it anywhere :cry:

toolness: Ah yes, there is for that one! Just hard to find in the labyrinthine docs. It's part of the [webpack node.js API](http://webpack.github.io/docs/node.js-api.html).

mmmavis: Thanks!  Heh, I tried looking for it in the config section, grunt section, and on https://nodejs.org/api but had no luck :stuck_out_tongue:

gulpfile.js
```diff
@@ -153,31 +179,18 @@ gulp.task('test-react-warnings', function() {
...
-  return new IndexFileStream(require('./lib/index-static.jsx'))
+  return createIndexFileStream()
     .on('end', function() {
       console.warn = oldWarn;
       if (warnings) {
         this.emit('error', new Error('At least one warning was logged.'));
       }
     })
-    .pipe(gulp.dest('./dist'));
+    .on('data', function() {});
```
mmmavis: Hey @toolness , could you help me understand why `on('data', function() {})` is needed here?

mmmavis:
Hi @toolness 

> * While running `gulp watch`, intentionally add syntax and runtime errors in various places in the code. This is to simulate real developers making real editing mistakes. Then make sure the console logs useful errors!

Working perfectly.

> * While running `gulp watch`, add basic edits in various parts of the code, and make sure the site regerates itself properly.

Yes, this works great as well.  (files got regenerated themselves and with manual refresh on browser I was able to see the changes)

> * Run `gulp` without this PR, make a copy of the `dist` directory, then run `gulp` with this PR, and make sure the `index.html` files in the two `dist` directories are identical.

Question: by identical did you mean comparing the code in those files or?

lib/gulp-index-file-stream.js
```diff
@@ -34,7 +34,10 @@ IndexFileStream.prototype._read = function() {
 
   this._indexStatic.generate(url, {
     meta: this._options.meta || {}
-  }, function(html) {
+  }, function(err, html) {
+    if (err) {
+      return this.emit('error', err);
+    }
```
mmmavis: (commenting on this line instead since Github doesn't let me do it on [line 19](https://github.com/mozilla/teach.webmaker.org/pull/839/files#diff-5c5ee15f5e04d0241e2101cf5b6ac58aR19))

I'm just curious why we have to reverse the array.  Does it matter [which url gets popped first](https://github.com/mozilla/teach.webmaker.org/pull/839/files#diff-5c5ee15f5e04d0241e2101cf5b6ac58aR28)?

gulpfile.js
toolness: Oh yeah! This has to do with the way that node streams work, and it's done to put the stream in **flowing mode**. Previously, the pipe to `gulp.dest('./dist')` (i.e., the filesystem) was what made node actually read from the stream. If that `.on('data', function() {})` isn't there, then node will just sit on the stream forever, waiting for it to be piped somewhere or be manually read from. You can read more about it at https://nodejs.org/api/stream.html#stream_class_stream_readable.

You can sort of think of that line as the node equivalent of piping to `/dev/null`, I think. At least that's how I read it. Though it would be more helpful for me to have added a comment like this:

```javascript
.on('data', function() { /* Drain the stream. */ });
```

which I should probably just do, since if you were confused by this, I'm sure others will be too.

toolness:
> Question: by identical did you mean comparing the code in those files or?

Yeah, um, I really should've done this myself before issuing this PR, but I think it'd be useful for both of us to do it, just to make sure. Basically I mean e.g. using the current develop branch without this PR and generating the files, then renaming the `dist` directory to `dist-old`, then switching to this PR and generating the files, and comparing the `index.html` files in `dist` and `dist-old` using a tool like `diff`. Since theoretically, both *should* be identical.

lib/gulp-index-file-stream.js
toolness: Ah, yeah, it is to preserve the ordering. Instead of using `.pop()` in the `_read` method I should've just used [`.shift()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift), but because I constantly get confused between shift and unshift and didn't want to look up the docs, at the time I just opted to reverse the array being popped-from instead! :sweat_smile:

gulpfile.js
mmmavis: I was reading https://nodejs.org/api/stream.html#stream_class_stream_readable yesterday but got confused cuz for things like https://github.com/mozilla/teach.webmaker.org/pull/839/files#diff-b9e12334e9eafd8341a6107dd98510c9R229 we are not adding the `on data` event listener there.  But yea, things will get a lot more clear to me after our Vidyo chat :smile:

toolness: Ah yeah, that is because `pipe`'ing a readable stream into another stream puts it into flowing mode (which is what's happening in the [example](https://github.com/mozilla/teach.webmaker.org/pull/839/files#diff-b9e12334e9eafd8341a6107dd98510c9R229) you linked to). It additionally allows node to automatically regulate how quickly it reads from the stream through **back pressure**, which is nicely explained in Tim Caswell's [Daddy, what's a stream?](http://howtonode.org/streams-explained) post.

Basically with a stream, you want to either add a `data` event listener *or* pipe it somewhere--otherwise node won't bother reading from it, because it doesn't think anyone's listening to it, and it also doesn't want to exhaust  available memory by buffering it all up. It's not very common to have a `data` event listener that does nothing, but we want to do it here because we're really just trying to see if any errors were logged while the readable stream was doing its thing--we don't care about the actual *output* of the stream, which is a bit unusual.

toolness: Oh! You should totally do substack's [stream adventure](https://github.com/substack/stream-adventure)! I should too, for that matter, as I haven't actually done that one yet.

lib/index-static-watcher.js
```diff
@@ -0,0 +1,120 @@
...
+      // re-require it.
+
+      delete require.cache[filename];
+
+      indexStatic = require(filename);
+      successCb(indexStatic);
+    }
+  };
+
+  self.build = function(cb) {
```
mmmavis: Hi @toolness , I just wanna check I'm getting this right.

So this part means 
if `index-static.bundle.js` has been generated, move on
if it's still being build, we make sure it doesn't get rebuilt within 100ms?

test/index-static.test.js
```diff
@@ -1,12 +1,24 @@
...
 
-require('node-jsx').install();
-
-var indexStatic = require('../lib/index-static.jsx');
+var indexStaticWatcher = require('../lib/index-static-watcher').create();
 
 describe('index-static', function() {
+  var indexStatic;
+
+  this.timeout(10000);
```
mmmavis: (was reading http://mochajs.org/#test-specific-timeouts, but didn't quite get it)
is this to increase Mocha's default timeout in case our bundled js takes longer to build?

lib/index-static-watcher.js
toolness: Very close--basically we don't want multiple builds going on simultaneously, so we just block any additional requests after the first one to wait until the first one is finished. They "wait" by coming back every 100ms and checking to see if the first one is done yet.

It's kind of an overly-complicated way to do it honestly. Maybe I should use a singleton [promise](https://www.promisejs.org/) instead?

test/index-static.test.js
toolness: Yep! It takes longer than mocha's default of 2000ms on my 5-yr old laptop, so i wanted to make sure it had plenty of time...

mmmavis:
Hey @toolness could you rebase this PR with `develop` so I can `diff` comparing the 2 copies of `dist`? thanks!

mmmavis:
I rebased on my local branch but `diff dist/ dist-develop/` did return some code differences back. :open_mouth: I don't always trust my rebasing skill so @toolness could you rebase and double check?

toolness:
Ugh, yeah, actually don't worry about diffing the files--even doing this kind of diff on the exact same branch appears to fail, because the `data-reactid` being given to elements is totally changing between runs. It must be based on the current timestamp or something? Anyways, super annoying.

mmmavis:
Alrighty! This PR totally make sense to me now! Thanks for walking me through @toolness 
Hooray R+ :+1: and I can finally unblock you :joy: 

:dancers: :dancers: :dancers: :dancers: :dancers:

toolness:
Wooot thanks mavis!

PR #849, opened 2015-04-30
Fixes #848 - cleanup /dist in the initial build steps

### LINKED ISSUE #848: Build step doesn't clean up old files
Our build step doesn't **delete** old files that were previously copied to `/dist`.   Maybe we can have a gulp task to handle that?

cc @toolness

mmmavis:
(WIP) 
This fixes #848.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -72,9 +72,9 @@ function handleError() {
   });
 }
 
-// cleanup /dist in the initial build steps
+// cleanup ./dist in the initial build steps
 gulp.task('cleanup-dist', function (cb) {
-  del(['/dist/*'], cb);
+  del(['./dist/*'], cb);
 });
 
 gulp.task('sitemap', ['generate-index-files'], function() {
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -17,6 +17,7 @@ var jscs = require('gulp-jscs');
 var jshint = require('gulp-jshint');
 var autoprefixer = require('gulp-autoprefixer');
 var rename = require('gulp-rename');
+var del = require('del');
 
 require('node-jsx').install();
 
@@ -34,6 +35,7 @@ var travis = require('./lib/travis');
 var server = require('./test/browser/server');
 
 var BUILD_TASKS = [
+  'cleanup-dist',
   'beautify',
   'copy-test-dirs',
   'copy-images',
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -70,6 +72,11 @@ function handleError() {
   });
 }
 
+// cleanup /dist in the initial build steps
+gulp.task('cleanup-dist', function (cb) {
+  del(['/dist/*'], cb);
+});
+
```
mmmavis: I think this task is enough... I guess we don't really have to watch for changes and clean up `/dist` every time something is changed.  That seem like it will slow down build time and is only useful for local development.

gulpfile.js
```diff
@@ -70,6 +72,11 @@ function handleError() {
   });
 }
 
+// cleanup ./dist in the initial build steps
+gulp.task('cleanup-dist', function (cb) {
+  del(['./dist/*'], cb);
+});
+
```
mmmavis: I think this task is enough... I guess we don't really have to watch for changes and clean up /dist every time something is changed. That seem like it will slow down build time and is only useful for local development.

mmmavis:
oops this won't work.  let me work on the fix

gulpfile.js
toolness: Yeah, it might even mess things up if tasks are run in parallel and the cleanup task is run at the same time as the building task--eek.

toolness: Um, speaking of running things in parallel, I *think* that's what might happen if you add `cleanup-dist` to the `BUILD_TASKS` array... I might be wrong, but I think those tasks are all run in parallel, so it might cause a race condition! It might actually be easiest for now to just call the task `cleanup` and have it be run manually?

toolness:
Hey, I was just watching http://learn.humanjavascript.com/react-ampersand/setting-up-webpack and around 10:30 he talks about using [pre and post hooks](http://blog.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/#preandposthooks) for `npm` scripts to do the same thing you're trying to do in this issue, which might be an alternative solution.

mmmavis:
:+1: :+1: :+1: that's a fantastic suggestion @toolness .  I feel like I've seen the `pre` and `post` hooks somewhere before but guess they got buried deep in my memory... and I shouldn't restrict myself with `gulp` for this task initially.  Heh!  Gonna close this PR and open a new one! :grin:

PR #851, opened 2015-04-30
Remove jsbeautify

### LINKED ISSUE #426: "gulp beautify" pollutes dist directory
I know I ought to know this since I reviewed #187, where the related code was first introduced, but I'm quite confused by our `gulp beautify` task:

```js
gulp.task('beautify', function () {
  gulp.src(LINT_DIRS)
      .pipe(beautify({ config: 'node_modules/mofo-style/linters/.jsbeautifyrc' }))
      .pipe(gulp.dest('dist'));
});
```

As far as I can tell, this just puts the beautified versions of some of our JS files in `dist`, which are completely separate from the bundles that webpack is creating. These files don't ever seem to be loaded by any code (which makes sense, since they're also being included in webpack's bundles). They just seem to pollute the `dist` directory and I've occasionally made the mistake of accidentally editing one of them (I've actually had to tell my Sublime Text 2 editor to ignore the `dist` directory to prevent this from happening).

~~I *assume* that we're running our code through beautify as a form of linting, so that any errors raised by beautify make our test suite fail... If that's the case, I wonder if there's a way to pipe the file output into the Vinyl equivalent of `/dev/null`, or at least a different directory than `dist`.~~ Nope, I think the intent here is to actually rewrite our code to conform to style guide. See #851 for more thoughts on this.

toolness:
This fixes #426 by removing "gulp beautify" entirely.

After a bit of research it appears that jsbeautify isn't a linter, it actually *reformats* your code to conform to a style guide. I don't mind linting, but I think there are potentially multiple solutions to a code style warning, and I want to be able to choose between them when I reformat my code--not have a beautifier make a decision for me.

At most, I think the beautification should be an *optional* step a developer can take to fix their lint warnings, rather than a required part of the build/test process. And since jsbeautify isn't working right now (as explained in #426) I'd rather just remove it.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,7 +12,6 @@ var plumber = require('gulp-plumber');
 var merge = require('merge-stream');
 var sourcemaps = require('gulp-sourcemaps');
 var sitemap = require('gulp-sitemap');
-var beautify = require('gulp-jsbeautify');
 var jscs = require('gulp-jscs');
 var jshint = require('gulp-jshint');
 var autoprefixer = require('gulp-autoprefixer');
@@ -34,7 +33,6 @@ var travis = require('./lib/travis');
 var server = require('./test/browser/server');
 
 var BUILD_TASKS = [
-  'beautify',
   'copy-test-dirs',
   'copy-images',
   'copy-event-resources',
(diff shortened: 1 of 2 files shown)


toolness:
@simonwex do you have an opinion on this? Is it in violation of mofo style stuff?

simonwex:
@gvn what's in jsbeautify that's missing from .jscsrc & .jshintrc?

gvn:
Ostensibly nothing. Jsbeautify is our crutch while JSCS lands autoformatting. It's just there to make conforming to JSCS and JSHint less painful. I'd consider it optional at this point...

simonwex:
R+

gvn:
Related: https://github.com/MozillaFoundation/javascript-style-guide/issues/19

PR #852, opened 2015-04-30
Fixes #720 organize assets

### LINKED ISSUE #720: Organize /img and its subdirectory
We should find time to organize assets and directories in `/img`.  

[`/img/activities-page`](https://github.com/mozilla/teach.webmaker.org/tree/develop/img/activities-page) is a case that the assets organization is different from all other directories for pages.  It has subdirectories for relevant assets to go.
vs
[`/img/about-page`](https://github.com/mozilla/teach.webmaker.org/tree/develop/img/about-page) where all assets live on the same level.

We should set a convention to follow and reorganize the `/img` folder after.

cc @toolness

mmmavis:
This fixes #720 

I kinda mocked what we did for `\less`

- Image and image@2x files are on the same level.  That way I think it'll be easier for us to spot missing file at a glance
- Separated `.svg` from non-scalable image files for the same reason
- I tried my best to remove old files (basically by searching them in the code) but I feel there could be some I missed.  We can file new tickets if we do find one.

```
img
│ 
├── components
│   │ 
│   ├── footer
│   │   └── svg
│   │       └── hive-logo.svg
│   ├── hero-unit
│   │   └── svg
│   │       └── mozilla_wordmark_white.svg
│   ├── ...
│   │
│   │       
└── pages
    │ 
    ├── about
    │   ├── about-illustration.svg
    │   ├── hive-nyc-summer-quest.jpg
    │   ├── icon-city.png
    │   ├── icon-city@2x.png
    │   └── svg
    │       ├── icon-connect-blue.svg
    │       ├── icon-get-help-blue.svg
    │       ├── icon-horizontal-divider.svg
    │       └── icon-twitter-blue.svg
    │   
    ├── ...
```


toolness:
Awesome!  This looks like a good organization, and thanks for the helpful diagram.

Regarding the removal of old files, I'm realizing that our URL checker could possibly be enhanced to check for this, though it wouldn't be able to detect links/images that are shown based on conditional logic.

Anyhow, I'll take a closer look at this soon.

toolness:
This is looking great! Just one thing though, it seems like the globe image in the add/remove club modal might be 404'ing now. When I run the test suite, the following is logged:



Hmm, I should probably make those kinds of things actually cause the test suite to fail.

Anyways, can you check those to make sure they're valid?

mmmavis:
@toolness , oops!  Thanks for spotting those.  Didn't know why I didn't catch those in the first place.  Anyways, I've updated my PR again.  :grin:

PR #855, opened 2015-05-01
Log safe mode info in Google Analytics

### LINKED ISSUE #846: No analytics for "safe mode" browsers
I just realized that, as far as I know, there actually isn't any analytics being done for browsers that either have no JS or are running in [safe mode](https://github.com/mozilla/teach.webmaker.org/pull/579).

Things we can do to resolve this:

* For browsers running in safe mode, we actually *can* run JS, we just don't. So we can still load `react-ga` and initialize it and send some sort of event indicating that the user is viewing the site in safe mode.
* For browsers that have JS disabled entirely, we could either do some kind of S3-based logging, or if GA supports image-based tracking, we could use that.

Any thoughts on this @adamlofting?

toolness:
This fixes #846 by moving GA initialization from `routes.jsx` to `main.jsx` (which is really a more appropriate place for it anyways) and emitting a `ga.pageview` and a custom `ga.event` if the user is in [safe mode](https://github.com/mozilla/teach.webmaker.org/pull/579) when visiting the site.

@adamlofting can you take a look at this?

### DIFF
--- a/lib/main.jsx
+++ b/lib/main.jsx
@@ -1,8 +1,12 @@
 var React = require('react');
 var Router = require('react-router');
+var ga = require('react-ga');
 
 var config = require('./config');
 var routes = require('./routes.jsx');
+
+var GA_ACCOUNT = process.env.GA_ACCOUNT || 'UA-49796218-20';
+var GA_DEBUG = process.env.GA_DEBUG || 'off';
 
 function startRunningSite() {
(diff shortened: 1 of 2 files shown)


adamlofting:
Sorry for the delay, I was on holiday last week. I'll look at this asap.

adamlofting:
R+ this is great.

toolness:
Woot thanks!!

PR #877, opened 2015-05-08
Fixes #698 - added unit test for Expander component

### LINKED ISSUE #698: Expander component doesn't have unit tests

mmmavis:
This fixes #698

### DIFF
--- a/components/expander.jsx
+++ b/components/expander.jsx
@@ -16,7 +16,7 @@ var Expander = React.createClass({
       expanded: true
     });
   },
-  onClick: function() {
+  handleClick: function(e) {
     if (this.state.expanded) {
       this.collapse();
     } else {
@@ -32,7 +32,7 @@ var Expander = React.createClass({
     return (
       <div className="expander-container">
         <div className={className} tabIndex="0" onBlur={this.collapse} onFocus={this.expand}>
-          <div onMouseDown={this.onClick} className="expander-header">
+          <div onClick={this.handleClick} className="expander-header">
             {this.props.head}
             <span className="ion"></span>
           </div>
--- a/test/browser/expander.test.jsx
+++ b/test/browser/expander.test.jsx
@@ -1,0 +1,38 @@
+var should = require('should');
+var React =require('react/addons');
+var TestUtils = React.addons.TestUtils;
+
+var stubContext = require('./stub-context.jsx');
+var Expander = require('../../components/expander.jsx');
+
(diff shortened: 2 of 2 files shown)


components/expander.jsx
```diff
@@ -16,7 +16,7 @@ var Expander = React.createClass({
       expanded: true
     });
   },
-  onClick: function() {
+  handleClick: function(e) {
```
mmmavis: renamed this handler since `onClick` could be confusing with the 'onClick' event

components/expander.jsx
```diff
@@ -32,7 +32,7 @@ var Expander = React.createClass({
     return (
       <div className="expander-container">
         <div className={className} tabIndex="0" onBlur={this.collapse} onFocus={this.expand}>
-          <div onMouseDown={this.onClick} className="expander-header">
+          <div onClick={this.handleClick} className="expander-header">
```
mmmavis: https://facebook.github.io/react/docs/events.html#mouse-events

I'm not sure why but handler of `onMouseDown` doesn't return me an `event` object in callback.  Changing the event to `onClick` seem to solve the problem.

Also, wondering if `e.preventDefault();` in the handler is really needed? or maybe we should this from `<div>` to `<a>` to indicate this is something togglable.

components/expander.jsx
toolness: Good idea--I've noticed this is a convention they use in the React tutorial and I've tried to stick to it myself.

components/expander.jsx
toolness: Hmm, good question, and I'm not sure what the answer is, though the MDN page on [using the button role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role) seems helpful. Perhaps it might be better to focus on the tests for this PR, and file an accessibility bug that we can address after landing this?

toolness:
Ok, I filed #879 and am landing this one. Thanks mavis!

mmmavis:
Awesome! Thanks @toolness

PR #902, opened 2015-05-14
Fixes #897 - removed ref to 'Add your event'

### LINKED ISSUE #897: Remove references to "add your event"
There are two places on /events where we tell people to add their event via events.webmaker.org. We've decided we don't actually need to collect data about individual MP events this year, so let's remove those references.

1. In the main "Join us" text, let's just remove the second clause ("and add your events to our global list any time of year")

2. Remove the "Add Your Event" CTA from the bottom. Can we just center the two remaining CTAs for now? Soon enough we'll be adding new CTAs to this page (see #843), but for now, let's just get the "Add Your Event" CTA off the page.

There's also a reference on /events/resources. Let's remove the entire "Add Your Events to the Map" section of the "Before Your Event" tab

mmmavis:
This fixes #897.

### DIFF
--- a/components/icon-links.jsx
+++ b/components/icon-links.jsx
@@ -1,15 +1,21 @@
 var React = require('react');
 
 var IconLinks = React.createClass({
+  componentDidMount: function() {
+    var numChild = this.props.children.length;
+    if (process.env.NODE_ENV !== 'PRODUCTION' && (numChild !== 2 && numChild !== 3) ) {
+      console.error("We should only have 2 or 3 icons in a IconLinks. Otherwise please assign different classes to wrap around <IconLink> component.");
+    }
+  },
   render: function() {
-    var links = this.props.children;
+    var links = Array.isArray(this.props.children) ? this.props.children : [this.props.children];
     var classes;
     return (
       <div className="row icon-links">
         {
           links.map(function(link, i){
             // we either have 2 or 3 icons in a IconLinks row
-            classes = (links.length === 2 & i === 0) ?
+            classes = (links.length === 2 && i === 0) ?
                       "col-sm-4 col-md-4 col-lg-4 col-sm-offset-2 col-md-offset-2 col-lg-offset-2" :
                       "col-sm-4 col-md-4 col-lg-4";
             return(
--- a/components/icon-links.jsx
+++ b/components/icon-links.jsx
@@ -2,8 +2,22 @@ var React = require('react');
 
 var IconLinks = React.createClass({
   render: function() {
+    var links = this.props.children;
+    var classes;
     return (
-      <div className="row icon-links">{this.props.children}</div>
+      <div className="row icon-links">
+        {
+          links.map(function(link, i){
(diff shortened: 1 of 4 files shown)


components/icon-links.jsx
```diff
@@ -2,8 +2,22 @@ var React = require('react');
...
+    var classes;
     return (
-      <div className="row icon-links">{this.props.children}</div>
+      <div className="row icon-links">
+        {
+          links.map(function(link, i){
+            // we either have 2 or 3 icons in a IconLinks row
+            classes = (links.length === 2 & i === 0) ?
+                      "col-sm-4 col-md-4 col-lg-4 col-sm-offset-2 col-md-offset-2 col-lg-offset-2" :
+                      "col-sm-4 col-md-4 col-lg-4";
```
toolness: Ah cool, I like your solution here.

toolness: Er wait, should `&` be `&&`? `&` is a [bitwise AND](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Bitwise_AND).

Also, I wonder if it'd be useful during `ComponentDidMount` to check if `node.env !== 'PRODUCTION'` and log a dev warning if the number of children is an unexpected value (0, 1, or greater than 3 I think?). Could be useful both as a dev warning and as self-documenting code, but it's up to you.

toolness:
Aside from the Bitwise AND, this looks good!

components/icon-links.jsx
mmmavis: > Er wait, should `&` be `&&`? `&` is a [bitwise AND](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Bitwise_AND).

Ah hahahah, nice catch.  Yes it should be `&&`.

> Also, I wonder if it'd be useful during `ComponentDidMount` to check if `process.env.NODE_ENV !== 'PRODUCTION'` and log a dev warning if the number of children is an unexpected value (0, 1, or greater than 3 I think?). Could be useful both as a dev warning and as self-documenting code, but it's up to you.

Yes, I like that plan!  Will update this PR.

components/icon-links.jsx
```diff
@@ -1,9 +1,29 @@
...
 
 var IconLinks = React.createClass({
+  componentDidMount: function() {
+    var numChild = this.props.children.length;
+    if (process.env.NODE_ENV !== 'PRODUCTION' && (numChild !== 2 && numChild !== 3) ) {
+      console.error("We should only have 2 or 3 icons in a IconLinks. Otherwise please assign different classes to wrap around <IconLink> component.");
+    }
+  },
   render: function() {
+    var links = Array.isArray(this.props.children) ? this.props.children : [this.props.children];
```
mmmavis: when `this.props.children` only has 1 `<IconLink>` component, it returns an object instead of array of objects and that breaks the `.map` call

toolness:
Woot thanks!

PR #912, opened 2015-05-14
Fixes #744 - added 'optional' next to non-required fields

### LINKED ISSUE #744: Visual treatment to indicate required fields on forms
As I was reviewing a [PR](https://github.com/mozilla/teach.webmaker.org/pull/688) related to the Add Your Club form validation I realized we currently don't have any visual indication for required fields.  

Re: @toolness's https://github.com/mozilla/teach.webmaker.org/issues/577#issuecomment-91236434 on the same matter.

I +1 on adding an asterisk next to required fields and a "* indicates required field" note somewhere near the bottom (just above the submit button?).  I personally dislike seeing "required" repeatedly as it feels a bit demanding and pushy to me.

cc @hannahkane @iamjessklein for suggestions

mmmavis:
This fixes #744.

### DIFF
--- a/less/common.less
+++ b/less/common.less
@@ -150,3 +150,20 @@ html.no-js .no-js-warning {
     }
   }
 }
+
+/**********
+* Form
+*/
+
+.optional-text {
+  color: @grey;
+  font-style: italic;
+  font-weight: normal;
+  &:before {
+    content: " (";
+  }
+  &:after {
+    content: ")";
+  }
+}
+
--- a/pages/clubs.jsx
+++ b/pages/clubs.jsx
@@ -407,7 +407,7 @@ var ModalAddOrChangeYourClub = React.createClass({
                valueLink={this.linkState('name')} />
             </fieldset>
             <fieldset>
-              <label>Where are you located?<span className="optional-text">optional</span></label>
+              <label>Where are you located?</label>
               <Select
                disabled={isFormDisabled}
                placeholder="Type in a city or a country"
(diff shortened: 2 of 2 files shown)


less/common.less
```diff
@@ -150,3 +150,20 @@ html.no-js .no-js-warning {
...
+
+/**********
+* Form
+*/
+
+.optional-text {
+  color: @grey;
+  font-style: italic;
+  font-weight: normal;
+  &:before {
```
mmmavis: Doing this so that screenreader will not read out `(` and `)`.   Otherwise it will say something like "bracket, optional, bracket" (if I remember it correctly)

pages/clubs.jsx
```diff
@@ -439,7 +439,7 @@ var ModalAddOrChangeYourClub = React.createClass({
                onChange={this.handleLocationChange} />
             </fieldset>
             <fieldset>
-              <label htmlFor={idPrefix + "website"}>What is your Club&lsquo;s website?</label>
+              <label htmlFor={idPrefix + "website"}>What is your Club&lsquo;s website?<span className="optional-text">optional</span></label>
```
mmmavis: Is this the `location` field supposed to be optional?

toolness: Oh, it's because the autocomplete widget we're using doesn't support that attribute :/ It's checked by our custom fallback validation code, though, so it is indeed required!

less/common.less
toolness: Ahh nice!!!!! Thanks for thinking about this.

toolness:
Aside from making the location field non-optional, this looks good! Feel free to make that change and merge.

mmmavis:
:tada: merged

Issue #818, opened 2015-04-24
Link checker doesn't spider srcset attributes


toolness:
As mentioned in https://github.com/mozilla/teach.webmaker.org/pull/809#issuecomment-95735829, our link checker isn't spidering `srcset` attributes. It'd be rad if it did.

PR #919, opened 2015-05-15
Use srcset and CSS resolution media queries (no JS) for retina displays

### LINKED ISSUE #841: React checksum warning on systems with Retina displays
Because the static index files render the `ImageTag` component assuming non-retina displays, there's a checksum mismatch when the page loads on devices with retina displays:

```
Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) e:url(/img/hero-unit@2x.png);" data-reac
 (server) e:url(/img/hero-unit.png);" data-reactid
```

Most unfortunate. We could resolve this either by removing the JS logic from `ImageTag` and just expecting browsers to use the `srcSet` attribute, or by having the initial render on the client-side *always* match what was rendered on the server.

**Update:** Oh, it's not just `ImageTag` that's causing this bug, it's also the `HeroUnit::calculateDensity()` function (since hero units use background image CSS to do their job).

toolness:
This fixes #841 by removing all our custom JS logic for checking [`window.devicePixelRatio`](https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio) and relying on the browser to choose the best image via the `srcset` attribute for image elements and CSS resolution media queries for background images.

Note that as an added bonus, our [link checker](https://github.com/mozilla/teach.webmaker.org/pull/809) now spiders both retina and non-retina background images, which it wasn't actually doing before due to the dynamic nature of our `HeroUnit` component's inline CSS.

I've verified this works on my non-retina Desktop PC and my retina iPad. However, I don't have a development environment set up for my iPad so I haven't actually been able to verify that the React checksum warning isn't logged. I'll do that early next week when I'm at the office. In the meantime, though, can you take a look at this @mmmavis?


toolness:
I just fixed #818, so once this lands, all our retina images will be link-checked! :dancer:

components/hero-unit.jsx
```diff
@@ -3,33 +3,6 @@ var React = require('react');
 var config = require('../lib/config');
 
 var HeroUnit = React.createClass({
-  calculateDensity: function () {
```
rossbruniges: Deleting all this code is awesome!

toolness:
Hehe thanks ross!

@mmmavis any idea when you might be able to review this one?

mmmavis:
yes for sure!  sorry i'm on this now

mmmavis:
this makes it SO MUCH cleaner! R+++++++ :dancers: :dancers: :dancers:

mmmavis:
btw, I like how I could test this on my retina MacBook Pro AND my non-retina external monitor under the same development environment :neckbeard:

toolness:
Ooo nice! I hadn't realized the multi-monitor situation... when you drag your browser from one monitor to the other, does the image change immediately, or do you have to reload the page?

mmmavis:
For the `srcSet` cases you *do* have to reload the page.  `media query` loads corresponding image right away without a page refresh.   (tested in on FF 38)

I should verify this finding with Chrome as well.

mmmavis:
It's the same on Chrome!

I wonder if there's an easier to check the actual source of the image.  Can't easily tell from inspecting the DOM.  I had to right click and "copy image location" or check Network panel to get the actual image source.

(same for both retina and non-retina displays)
screen shot 2015-05-29 at 2 42 47 pm

PR #926, opened 2015-05-18
Various easy accessibility fixes

toolness:
Made a few simple accessibility fixes, many surfaced by tenon.io:

* Set `role` on the navbar, main content, and footer. In NVDA, for example, pressing `d` allows users to easily navigate between [landmark roles](http://www.w3.org/WAI/GL/wiki/Using_ARIA_landmarks_to_identify_regions_of_a_page), which makes it much easier for them to navigate to important parts of the page if these are defined.

* The list of Hives in `/activities/` has purely [decorative](https://github.com/mozilla/teach.webmaker.org/pull/905#issue-76379588) Hive logos next to each Hive city, whose `alt` attributes should be nulled.

* The "Learn more" link in the footer is unhelpful to screen reader users:

    > Do not use generic text in links. Provide text in the link that accurately and concisely indicates what will be found at the destination.

    So for screen readers this now reads as "Learn more about Hive Learning Networks". (The [`sr-only`](http://getbootstrap.com/css/#conveying-meaning-to-assistive-technologies) class is used for the extra text, so it still looks unchanged for sighted users.)

* Set `lang="en"` on the `<html>` element so that screen readers know how to pronounce the words on the page.

* Added [skip navigation](http://getbootstrap.com/getting-started/#skip-navigation). I was originally going to file this as a separate PR, but it would've conflicted with this one, so I just added it as another commit.

### DIFF
--- a/components/footer.jsx
+++ b/components/footer.jsx
@@ -5,7 +5,7 @@ var Link = Router.Link;
 var Footer = React.createClass({
   render: function() {
     return (
-      <footer className="row">
+      <footer className="row" role="contentinfo">
         <div className="sidebar col-md-3">
           <div className="row">
             <div className="col-xs-6">
--- a/components/footer.jsx
+++ b/components/footer.jsx
@@ -27,10 +27,10 @@ var Footer = React.createClass({
         <div className="content col-md-9">
           <div className="row">
             <div className="col-sm-4 col-sm-offset-1">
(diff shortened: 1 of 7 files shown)


toolness:
@mmmavis, when you take a look at this, feel free to take some extra time to learn how to use [VoiceOver](http://webaim.org/articles/voiceover/) (assuming you're on OS X) to test things out manually.

mmmavis:
Yay, just tested this PR with VoiceOver and everything works perfectly.  Thanks @toolness !

PR #928, opened 2015-05-18
Use <h4> for Expander heading.

### LINKED ISSUE #879: Expander may not be accessible
As @mmmavis noted in https://github.com/mozilla/teach.webmaker.org/pull/877#discussion_r29916167, the Expander component, which was originally based entirely on CSS focus rules (and thus automatically accessible) but changed in #788 to use custom JS, may not be accessible anymore.

On the one hand, it may still be accessible: since the expander is only used to hide content to make it not look as visually overwhelming for sighted users, people who use screen readers will still see all the page content. That said, they might not be able to easily navigate through it since we don't actually use header tags for the `expander-header` class.

Another alternative may be to use the [button role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role) on our markup, though I'm not sure how this will impact users running in safe mode (or with JS disabled).

toolness:
This fixes #879 and makes navigating & understanding Expander content easier for users with screen readers. Visual styling and interactivity of Expanders for sighted users is not affected.

### DIFF
--- a/less/components/expander.less
+++ b/less/components/expander.less
@@ -21,6 +21,9 @@ html.no-js {
     font-weight: bold;
     text-transform: uppercase;
     padding: 2rem;
+    margin: 0;
+    font-size: inherit;
+    line-height: inherit;
     transition: background-color 0.5s ease;
     .ion {
       float: right;
--- a/components/expander.jsx
+++ b/components/expander.jsx
@@ -32,10 +32,10 @@ var Expander = React.createClass({
     return (
       <div className="expander-container">
         <div className={className} tabIndex="0" onBlur={this.collapse} onFocus={this.expand}>
-          <div onMouseDown={this.handleMouseDown} className="expander-header">
+          <h4 onMouseDown={this.handleMouseDown} className="expander-header">
             {this.props.head}
             <span className="ion"></span>
-          </div>
+          </h4>
           <div className="expander-items-container">
             <div className="items-margin">
               {this.props.children}
--- a/less/pages/clubs-toolkit.less
+++ b/less/pages/clubs-toolkit.less
@@ -10,7 +10,6 @@
     }
   }
   .expander-header {
-    font-size: 1.5rem;
     background-color: @headerBackground;
     color: @headerText;
   }


less/components/expander.less
```diff
@@ -21,6 +21,9 @@ html.no-js {
     font-weight: bold;
     text-transform: uppercase;
     padding: 2rem;
+    margin: 0;
+    font-size: inherit;
```
mmmavis: This gets overriden by https://github.com/toolness/teach.webmaker.org/blob/639a89c0a366d91be0dc0029303dbcf1f31e3f1d/less/pages/clubs-toolkit.less#L13.  I'm guessing we can just have `font-size: 1.5rem` (or `1.6rem` to match default body copy size) to unify all `.expander-header` font sizes?

toolness: Oh, I think it is supposed to override it. I just added the rules necessary for the `h4` to not have any visual difference from the `div`... I believe that `font-size: inherit` is needed to make the headings look like they used to on the web literacy page, while the overridden `font-size: 1.5rem` is needed to make them look like they used to on the clubs toolkit page?

mmmavis: Oh yea what you said makes sense @toolness !  

I guess I'm just wondering if we can remove those 2 `font-size` rules altogether.  

- For the Clubs Toolkit page, it's kinda odd to me that we specifically set a "header" type of text just 1px smaller than default body copy size (16px) and that rule only applies on a single page on our site.  (maybe it was falsely set or things got changed along the way?)
- For the `.expander-header` class, `font-size: inherit` essentially means set the font size to 16px now.

For these reasons I'm thinking if we should just remove those CSS rules. @sabrinang thoughts?

mmmavis: @sabrinang 
FYI, this is the "Expander" we are talking about

screen shot 2015-05-19 at 4 14 42 pm

You can find them on https://teach.mozilla.org/teach-like-mozilla/web-literacy/ and https://teach.mozilla.org/clubs/toolkit/

toolness:
Ahhh gotcha, yeah that makes sense! will wait for @sabrinang's comment before proceeding.

sabrinang:
please unify size to 16px

mmmavis:
@toolness R++++++ if you could just remove the two `font-size` rules.  Feel free to merge this PR once that's done.  Thanks!!!

toolness:
Er, *both* `font-size` rules? Removing the one you mentioned at https://github.com/toolness/teach.webmaker.org/blob/639a89c0a366d91be0dc0029303dbcf1f31e3f1d/less/pages/clubs-toolkit.less#L13 has a negligible effect on the clubs toolkit page, so I think it is a useful simplification, but removing the `font-size: inherit` is a bit more of a major change because it actually sets the font size to 18px, rather than 16px (I guess because `<h4>` is set to 18px by bootstrap, it looks like).

toolness:
I could alternatively change `font-size: inherit` to `font-size: 1.6rem`, which also appears to change the font size to 16px.

mmmavis:
oh yea sorry, ignore me.  removing the one from `clubs-toolkit.less` should be sufficient. thanks!

Issue #911, opened 2015-05-14
Implement approval flow for Clubs


hannahkane:
detail to come (pending final decisions on https://teach.etherpad.mozilla.org/clubs-signup)

toolness:
I'm noticing that the flow described on the etherpad seems a bit intricate, and someone left the following note there:

> At this stage, don't need to optimize with numbers. Do manually so have a better feel.

Given this, we'll want to find a good balance between writing up new code/UI for the changes versus doing things manually. I suspect there are just a few relatively minor tweaks we can make to the system that will allow the Teach team and RCs to freely iterate on an approval flow; we want to find this sweet spot, so that the staff feels empowered and the engineering team has enough time to focus on the other priorities of the current milestone!

toolness:
Given the fact that clubs will be managed by different individuals with different roles at different times, and that the approval flow will likely involve lots of different phases, I'm wondering if we should add a "staff notes" field to the Club model.

This would be a field that is only visible to staff users of the [admin UI](https://github.com/mozilla/teach.webmaker.org/issues/925#issuecomment-104705395), and they could update it with any information relevant to the current approval flow status of the club that isn't otherwise captured by the system, e.g. "RC alice has set up a 1:1 interview w/ the club lead on june 5". This way more information is stored in the DB for all staff to see, rather than e.g. stuffed in an email thread somewhere that only a few people can see.

hannahkane:
@toolness - does this feel blocked anymore? I think we've removed all of the blockers.

Re: staff notes field—as we talked about at today's stand-up, I'm hoping that, initially, all of the Club management (assigning RCs to clubs, tracking phone calls and emails, scheduling follow-up tasks) will happen in a spreadsheet so that staff and RCs can tweak the process as much as they need without developer involvement. Once the process is stabilized, we can decide whether to add more functionality to the site, to automate or otherwise streamline.

toolness:
Right on, sounds good hannah! This is indeed not blocked anymore IMO, I'll remove the label.

toolness:
We've got a whole lot of the parts in place for this issue, but one important thing that we shouldn't forget to do, and which this issue can perhaps track, is write **documentation** for the new approval flow process, for staff/RCs/admins. We'll probably want to do this collaboratively with the teach team.

toolness:
[...]

PR #929, opened 2015-05-18
Show "Site Administration" link in sidebar when user is staff

toolness:
In https://github.com/mozilla/teach-api/commit/11ad5456a9696df1a08581414f8268f037d7492e I added support for the Teach API to tell clients if the current user is staff and, if so, what the URL for the Admin UI is.

This PR uses that new functionality to show a "Site Administration" link in the sidebar if (and only if) the user is staff:

This should be particularly useful as we move towards implementing the approval flow for clubs in #911, which will require regional coordinators and teach staff to visit the admin UI more frequently, and we don't want them constantly having to remember what the link to the admin UI is. :grin:

### DIFF
--- a/test/browser/login.test.jsx
+++ b/test/browser/login.test.jsx
@@ -60,7 +60,7 @@ describe("Login", function() {
     login.getDOMNode().textContent.should.not.match(ADMIN_RE);
   });
 
-  it("doesn't show admin link for non-staff users", function() {
+  it("shows admin link for staff users", function() {
     teachAPI.getAdminURL.returns("http://admin");
     login.setState({username: "blop"});
     login.getDOMNode().textContent.should.match(ADMIN_RE);
--- a/test/browser/login.test.jsx
+++ b/test/browser/login.test.jsx
@@ -11,6 +11,8 @@ var LogoutLink = Login.LogoutLink;
 var StubTeachAPI = require('./stub-teach-api');
 var StubRouter = require('./stub-router');
 
+var ADMIN_RE = /administration/i;
+
 describe("Login", function() {
   var login, teachAPI;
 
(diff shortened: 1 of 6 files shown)


test/browser/login.test.jsx
```diff
@@ -53,6 +55,17 @@ describe("Login", function() {
     login.getDOMNode().textContent.should.match(/blop/);
   });
 
+  it("doesn't show admin link for non-staff users", function() {
+    login.setState({username: "blop"});
+    login.getDOMNode().textContent.should.not.match(ADMIN_RE);
+  });
+
+  it("doesn't show admin link for non-staff users", function() {
```
mmmavis: a typo here I think, should this be for "staff" users instead?  :hamster:

toolness: ah yeah, nice catch! will fix.

mmmavis:
Hey @toolness ,

I don't know Python :disappointed:  and I don't have an admin account to test this.

Could you help me understand what the following snippet returns? (from https://github.com/mozilla/teach-api/commit/11ad5456a9696df1a08581414f8268f037d7492e#diff-bc7ccec6a99b9a600a491d8133aa0b84R136)

```
body['admin_url'] = '%s%s' % (settings.ORIGIN,
                                      reverse('admin:index'))
```

mmmavis:
Ah, I guess that means
if user is a staff, `admin_url` will be set to `settings.ORIGIN` + `reverse('admin:index')`?
I'm reading https://docs.djangoproject.com/en/1.8/ref/urlresolvers/#reverse now but still unsure what `admin:index` is in the snippet

toolness:
Oh, heh, you don't need to review that, but thanks for looking at it. The `%` is the [string interpolation](https://docs.python.org/2/library/stdtypes.html#string-formatting-operations) operation in Python, and `ORIGIN` ultimately comes from an environment variable documented in the [teach-api `README.md`](https://github.com/mozilla/teach-api#environment-variables). [`reverse()`](https://docs.djangoproject.com/en/1.8/ref/urlresolvers/#reverse) is a [DRY-friendly](http://en.wikipedia.org/wiki/Don%27t_repeat_yourself) way to get a URL for a view--semantically, it's very similar to the way that we use react-router's [`<Link>`](http://rackt.github.io/react-router/#Link) with a route name (rather than a hard-coded URL path) in the `to` prop, which guarantees that the link will always point at where we need it to go, even if we decide to change the URL path for the view.

toolness:
Oh `admin:index` is the view name for the root of the [django admin site](https://docs.djangoproject.com/en/1.8/ref/contrib/admin/).

mmmavis:
Heh, I primarily wanted to know what a sample login url will be for staff users

mmmavis:
and R+ thanks @toolness :dancer:

PR #932, opened 2015-05-20
Fixes #848 - cleaned up dist dir before running npm test and npm start

### LINKED ISSUE #848: Build step doesn't clean up old files

mmmavis:
This fixes #848 and thanks for @toolness's the [great suggestion](https://github.com/mozilla/teach.webmaker.org/pull/849#issuecomment-101396329)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,9 +53,13 @@
     "test": "gulp smoketest && mocha test/*.test.js && npm run browsertest && gulp lint-test && npm run spider",
     "spider": "node test/browser/spider.js",
     "browsertest": "node test/browser/run-in-phantom.js",
+    "presmoketest": "rm -rf dist",
     "smoketest": "gulp smoketest",
+    "pres3": "rm -rf dist",
     "s3": "gulp s3",
+    "travis-after-success": "rm -rf dist",
     "travis-after-success": "gulp travis-after-success",
+    "prebuild": "rm -rf dist",
     "build": "gulp",
     "prestart": "rm -rf dist",
     "start": "gulp watch"
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "smoketest": "gulp smoketest",
     "pres3": "rm -rf dist",
     "s3": "gulp s3",
-    "travis-after-success": "rm -rf dist",
+    "pretravis-after-success": "rm -rf dist",
     "travis-after-success": "gulp travis-after-success",
     "prebuild": "rm -rf dist",
     "build": "gulp",
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
   },
   "devDependencies": {},
   "scripts": {
-    "pretest": "rm -r dist",
+    "pretest": "rm -rf dist",
     "test": "gulp smoketest && mocha test/*.test.js && npm run browsertest && gulp lint-test && npm run spider",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -49,13 +49,19 @@
...
   "scripts": {
+    "pretest": "rm -rf dist",
     "test": "gulp smoketest && mocha test/*.test.js && npm run browsertest && gulp lint-test && npm run spider",
     "spider": "node test/browser/spider.js",
     "browsertest": "node test/browser/run-in-phantom.js",
+    "presmoketest": "rm -rf dist",
     "smoketest": "gulp smoketest",
+    "pres3": "rm -rf dist",
     "s3": "gulp s3",
+    "travis-after-success": "rm -rf dist",
```
toolness: shouldn't this be `pretravis-after-success`?

toolness:
Other than my one note, this looks good! Feel free to merge once you've fixed things.

package.json
mmmavis: heh, yes. copy and paste mistake :stuck_out_tongue:

mmmavis:
I'm waiting for Travis to finish then I'll merge :)

toolness:
Awsm thanks!

We should also double-check to make sure that Travis pushes this particular changeset to staging. Because we're technically changing some of the code that is only run when Travis pushes to staging/prod, we can only really make sure it's still working properly by checking the commit link in the dev ribbon modal after Travis is done pushing to https://teach.mofostaging.net/ !

mmmavis:
Merging it now.

@toolness 
When things get merged into develop, how long do we have to wait for them to show on staging usually?

toolness:
Thanks! 

The push to staging seems to vary a lot depending on how much load Travis is under. I find, for instance, that in the morning EST, it seems to be extremely fast, but by midday it can take a really long time. :/

I guess I recommend just keeping a tab open to https://travis-ci.org/mozilla/teach.webmaker.org and monitoring that?

toolness:
Woot it looks like it is indeed pushing stuff properly to staging!

One interesting thing it is *not* doing, is wiping the S3 bucket before pushing the new content. Technically that might be a good idea, I'm not really sure.

mmmavis:
> One interesting thing it is *not* doing, is wiping the S3 bucket before pushing the new content. Technically that might be a good idea, I'm not really sure.

is it some permission problem?  maybe running `sudo` for `pres3` script could help? (just guessing)

toolness:
Oh no, it's not a problem w/ permissions at all--it's just faster for the S3 deployment script to write files into that bucket, overwriting any existing files, rather than wipe it first, or do some kind of synchronization logic. That's not really that big a deal usually, but it could have repercussions if we delete pages from the site, as they'll actually still be accessible on the S3 bucket.

For pages that have moved, like `/clubs/curriculum/`, this isn't a big deal because we're actually overwriting those pages with [redirect pages](https://teach.mofostaging.net/clubs/curriculum/?safemode=on). But for ones that we actually decide to *delete* and not redirect, then unless we wipe the bucket before deploying (or perform some kind of synchronization logic), the old pages won't actually 404!

I guess this problem will go away once we move to a lightweight dynamic server though (#585).

PR #942, opened 2015-05-23
Make LogoAsset keyboard accessible.

### LINKED ISSUE #941: "Logos & Assets" section of Event Resources isn't keyboard accessible
As mentioned in https://github.com/mozilla/teach.webmaker.org/issues/724#issuecomment-104774041:

> I'm also noticing that the Logos & Assets above the Tabulator on the Events Resources also isn't keyboard accessible--it's a problem somewhat similar to #927, where the keyboard focus appears to "disappear" while the focus is inside one of the resource download links.

toolness:
This fixes #941.

### DIFF
--- a/less/pages/event-resources.less
+++ b/less/pages/event-resources.less
@@ -109,7 +109,7 @@ html.no-js {
         }
         opacity: 0;
         transition: opacity .3s ease-in-out;
-        &:hover {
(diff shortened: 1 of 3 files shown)


mmmavis:
yay yay yay R+

Issue #906, opened 2015-05-14
Remove the imgAlt prop from IconLink components for better accessibility


toolness:
As mentioned in #905, the [W3C Techniques for providing useful text alternatives](http://dev.w3.org/html5/alt-techniques/#sec5) states:

> Purely decorative images must be marked up so they can be ignored by assistive technology with a null alt attribute (`alt=""`) or preferably use CSS techniques. If the image isn't providing the user any informative content or enhancing greater understanding of the content, then the `alt` attribute must be empty.

The icons in our `IconLink` components are purely decorative and should be ignored by assistive technology, so we should remove the `imgAlt` prop from them and simply hard-code `alt=""` for the icons in the component's `render()` method.

PR #943, opened 2015-05-26
(WIP) Fixes #856 new homepage

### LINKED ISSUE #856: Implement homepage redesign
Pending completion of #800 

Important: the blog feed should be limited to articles tagged with "#teachtheweb" rather than the entire blog feed.

mmmavis:
This fixes #856

**[ WIP ]**
- [ ] Missing proper tests
- [x] should render longer blog post excerpt
- [ ] 3 buttons on hero unit
- [ ] optimize images (see https://github.com/mozilla/teach.webmaker.org/pull/943#issuecomment-105556993)
- [x] update changelog


less/index.less
```diff
@@ -4,3 +4,4 @@
 @import "components/index";
 @import "pages/index";
 @import "../node_modules/react-select/less/default.less";
+@import "../node_modules/react-select/less/default.less";
```
toolness: Hmm why is the same file imported twice?

less/pages/home.less
```diff
@@ -1,56 +1,84 @@
...
+  h2 {
+    margin-bottom: 3rem;
   }
+}
 
-  @media (min-width: @screen-sm-min) {
-    .row {
-      align-items: stretch;
-      display: flex;
+#featured-post {
```
toolness: Is there a reason we're using an id-based selector instead of a class/tag-based one here? It seems like a best practice to not use ids whenever possible, but if there's a good reason for it then that's cool.

less/pages/home.less
```diff
@@ -1,56 +1,84 @@
...
 }
 
-.values {
-  padding: @grid-gutter-width 0;
-  font-size: 3rem;
-
-  @media (min-width: @screen-sm-min) {
-    align-items: center;
-    display: flex;
+#recent-posts {
```
toolness: Same concern as https://github.com/mozilla/teach.webmaker.org/pull/943/files#r31041302

pages/home.jsx
```diff
@@ -5,15 +5,156 @@ var Link = Router.Link;
...
+  }
+});
+
+var FeaturedPost = React.createClass({
+  propTypes: {
+    data: React.PropTypes.object.isRequired
+  },
+  render: function() {
+    var publishedDate = new Date(this.props.data.publishedDate);
+    var month = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"][publishedDate.getMonth()];
```
toolness: I'm wondering if we should just use a library like [moment.js](http://momentjs.com/) here? That way we won't have to write unit tests for this month logic, and [i18n](http://momentjs.com/docs/#/i18n/) should hopefully be easier too (when we get to that).

pages/home.jsx
```diff
@@ -5,15 +5,156 @@ var Link = Router.Link;
...
+  },
+});
+
+var LatestPosts = React.createClass({
+  propTypes: {
+    data: React.PropTypes.array.isRequired
+  },
+  render: function() {
+    function formatDate(theDate) {
+      return ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"][theDate.getMonth()] + " " +
```
toolness: Same concern as [above](https://github.com/mozilla/teach.webmaker.org/pull/943/files#r31041939) regarding using Moment.js.

pages/home.jsx
```diff
@@ -25,20 +166,50 @@ var HomePage = React.createClass({
...
+              href="mailto:teachtheweb@mozillafoundation.org"
+              imgSrc="/img/pages/about/svg/icon-get-help-blue.svg"
+              imgAlt="question mark icon"
+              head="Get Help"
+              subhead="Email us anytime"
+            />
+            <IconLink
+              href="http://discourse.webmaker.org/category/meet"
+              imgSrc="/img/pages/about/svg/icon-connect-blue.svg"
+              imgAlt="illustration of two people connecting"
```
toolness: We'll be removing `imgAlt` in #906 because the icons are purely decorative, but in the meantime it might be a good idea to have `imgAlt=""` for any new code we're adding. What do you think?

pages/clubs.jsx
```diff
@@ -139,10 +139,10 @@ var Quote = React.createClass({
     return(
       <div className="row">
         <div className="col-sm-offset-2 col-sm-8 col-md-offset-2 col-md-8 col-lg-offset-2 col-lg-8">
-          <Blockquote className="primary-quote" author="Mikko K, Helsinki, Finland"
```
toolness: Weird, so is `primary-quote` never actually used in a CSS rule? I guess it seems like it isn't, according to a [GitHub search for `primary-quote`](https://github.com/mozilla/teach.webmaker.org/search?q=primary-quote&type=Code&utf8=%E2%9C%93) at least. So I guess it's a good idea to just get rid of it...

toolness:
The new `hero-unit.png` and its 2x variant are 160k and 560k, respectively, which is quite large. Can you [optimize](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization?hl=en) the file sizes a bit?

pages/clubs.jsx
```diff
@@ -139,10 +139,10 @@ var Quote = React.createClass({
...
+          <Blockquote author="Mikko K, Helsinki, Finland"
               imgSrc="/img/pages/clubs/mikko-finland.png" imgSrc2x="/img/pages/clubs/mikko-finland@2x.png" imgAlt="Mikko Finland Quote">
 
-            <p>The idea of teachers and students learning at the same time is what makes me excited about this work.</p>
+            <p>"The idea of teachers and students learning at the same time is what makes me excited about this work."</p>
```
toolness: I think it'd be swell to use `&ldquo;` and `&rdquo;` for typographic awesomeness here but it's up to you :grin:

pages/home.jsx
```diff
@@ -5,15 +5,156 @@ var Link = Router.Link;
...
+            <time className="published" title={this.props.data.publishedDate} dateTime={this.props.data.publishedDate} >
+              <span className="posted-month">{month}</span>
+              <span className="posted-date">{publishedDate.getDate()}</span>
+              <span className="posted-year">{publishedDate.getFullYear()}</span>
+            </time>
+          </p>
+        </div>
+        <div className="entry-header-container">
+          <h3 className="entry-title"><a href={this.props.data.link}>{this.props.data.title}</a></h3>
+          <cite className="author">{this.props.data.author}</cite>
```
toolness: Nice use of `<time>`, `<h3>`, and `<cite>`!

pages/home.jsx
```diff
@@ -5,15 +5,156 @@ var Link = Router.Link;
...
+    gScript.setAttribute("src", "https://www.google.com/jsapi");
+    gScript.onload = callback;
+    head.appendChild(gScript);
+    this.setState({googleAPILoaded: true});
+  },
+  componentDidMount: function() {
+    var self = this;
+    var google;
+    this.loadGoogleAPI(function() {
+      google = window.google;
```
toolness: Might want to guard these callbacks with a check to see if [`this.isMounted()`](https://facebook.github.io/react/docs/component-api.html#ismounted) is true. Otherwise, for instance, if the network is slow, the user may navigate to a different section of the site before the callback is called, and then an exception will be thrown (and unnecessary network activity may occur) when the callback is finally triggered.

pages/home.jsx
```diff
@@ -5,15 +5,156 @@ var Link = Router.Link;
...
+        author: "",
+        publishedDate: "",
+        contentSnippet: "",
+        link: ""
+      },
+      latestPostsData: []
+    }
+  },
+  loadGoogleAPI: function(callback) {
+    var head = document.getElementsByTagName("head")[0];
```
toolness: It seems like this should first check to see if `window.google` already exists, and if so, just `return process.nextTick(callback)`. Otherwise, you may be loading a new script tag every time the user navigates to the homepage.

toolness:
Err, sorry I started reviewing this before you were finished--just figured I'd get a head start, but please let me know if my comments were unhelpful.

less/pages/home.less
mmmavis: I was thinking there should only be **one** featured post on a page so I used `id` instead of `class`.  I guess over specific selector might not be a good idea so I'm happy to change it into `class`.

less/index.less
mmmavis: purely by mistake :grin:

pages/clubs.jsx
mmmavis: Yea, I realized the majority of the CSS code in [`home.less`](https://github.com/mozilla/teach.webmaker.org/blob/develop/less/pages/home.less) is outdated/no longer in use so I cleaned that up.  Same with `primary-quote`.

pages/clubs.jsx
mmmavis: Maybe using CSS pseudo-elements `:before` and `:after` could help?  And I guess we don't have to pass `imgAlt` here as well since the image is decorative

mmmavis: OR...
I wonder if we could use `<q>` to wrap around the actual quote instead of `<p>`.  I tested it with VoiceOver and it interpreted the two(`<q>` vs `<p>`) the same.

mmmavis:
not a problem as all, and thanks for those suggestions and comments :bowtie:

mmmavis:
it's better to know early whether or not i'm on the right track! :dancer:

pages/clubs.jsx
toolness: Ah cool, thanks!

pages/clubs.jsx
toolness: Oh neat, I didn't even know `<q>` was a tag!

pages/home.jsx
mmmavis: Lovely! :tada: Didn't realize such thing existed!

pages/home.jsx
```diff
@@ -1,19 +1,157 @@
...
+  propTypes: {
+    data: React.PropTypes.object.isRequired
+  },
+  render: function() {
+    var parsedDate = moment(this.props.data.publishedDate);
+    return(
+      <div className="featured-post">
+        <div className="entry-posted-container">
+          <p className="entry-posted">
+            <time className="published" title={this.props.data.publishedDate} dateTime={this.props.data.publishedDate} >
```
mmmavis: leaving `dateTime` value unparsed here since it's in machine readable format already

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time
> This element is intended to be used presenting dates and times in a machine readable format. This can be helpful for user agents to offer any event scheduling for user's calendar.

pages/home.jsx
```diff
@@ -1,19 +1,157 @@
...
+  },
+  render: function() {
+    return (
+      <ul className="recent-posts">
+        {
+          this.props.data.map(function(post, i) {
+            return (
+              <li key={i}>
+                <a className="post-title">{post.title}</a>
+                <time className="published" title={post.publishedDate} dateTime={post.publishedDate}>
```
mmmavis: same reason as mentioned in https://github.com/mozilla/teach.webmaker.org/pull/943/files#r31094020

pages/home.jsx
mmmavis: Hi @toolness , sorry I'm not so familiar with `process.nextTick`, did you mean

```
loadGoogleAPI: function(callback) {
    if (window.google) {
      return process.nextTick(callback);
    }
    var head = document.getElementsByTagName("head")[0];
    var gScript = document.createElement("script");
    gScript.setAttribute("src", "https://www.google.com/jsapi");
    gScript.onload = callback;
    head.appendChild(gScript);
    this.setState({googleAPILoaded: true});
  },
```

If that's the case wouldn't everything starting from `var head = ...` gets run again before `process.nextTick`? :open_mouth: OR that means "return to the loadGoogleAPI method caller and continue"?

mmmavis:
hey @toolness 

(from https://github.com/mozilla/teach.webmaker.org/issues/946)
> The new `hero-unit.png` and its 2x variant are 160k and 560k, respectively, which is quite large. Can you [optimize](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization?hl=en) the file sizes a bit?

I just created [this ticket](https://github.com/mozilla/teach.webmaker.org/issues/946) and was thinking maybe we could do the image optimization all at once?

This PR is getting bigger and bigger and I started to feel a bit worn out. :cry:  Do you mind reviewing what I have here first and I'll work on adding tests on Wed (probably after I reviewed [your PR](https://github.com/mozilla/teach.webmaker.org/pull/936) :cold_sweat: )?  I'm just trying to think of an effective way to unblock you / the upcoming release so let me know what you think!  :eyes:

pages/home.jsx
toolness: Yep, it's just async code for "return to the loadGoogleAPI method caller and continue". Technically we could just do `return callback()` but it can result in problems sometimes, e.g. consider this:

```javascript
var x;
doSomethingAsync(function() {
  x++;
});
x = 0;
```

You'd think that this code would run the callback *after* setting `x` to `0`, but if `doSomethingAsync` has a `return callback()` anywhere in it, then the callback will actually get called *before* `doSomethingAsync` even returns. This can result in really hard-to-figure-out bugs in your code, so it's generally best practice to *always* wait until the current thread runs to completion before calling the callback, which is basically what `process.nextTick()` does (think of it like `setTimeout(callback, 0)`).

pages/home.jsx
```diff
@@ -1,19 +1,167 @@
...
+      latestPostsData: []
+    }
+  },
+  loadGoogleAPI: function(callback) {
+    var head = document.getElementsByTagName("head")[0];
+    var gScript = document.createElement("script");
+    gScript.setAttribute("src", "https://www.google.com/jsapi");
+    gScript.onload = callback;
+    head.appendChild(gScript);
+    this.setState({googleAPILoaded: true});
```
toolness: Er, I just realized that this `googleAPILoaded` state variable isn't actually being used anywhere. Since you're just checking the existence of `window.google` to see if the API is loaded, maybe just get rid of this variable?

toolness:
Oh, no worries about feeling worn out--this PR is actually a bit of a doozy because it involves both CSS work *and* some tricky JS logic, and I have a hard time juggling between those different modes of thinking. :tired_face:

I'm fine with moving some of the issues, like image optimization, to a different ticket. (That said, for the image optimization at least, I'd like to either switch to [Git LFS](https://git-lfs.github.com/) ASAP or only commit optimized images after this PR, as it bloats the repository and doing a fresh clone of the repo is already starting to take a while!).

Also, don't feel too pressured about reviewing #936--it's not super-urgent, and if it's easier for you to just focus on this for now, that's fine. On the other hand, if you're worn out on this PR and could use a review break, that's cool too.

toolness:
FWIW, I recommend abstracting out the Google Feed API code into a separate module/function/class. This will decouple your `BlogSection` class from the details of the Google Feed API, which will make it easier to unit test (which is why I bring it up now), and it'll also make it easier to switch to a different implementation (such as one of the solutions we discussed in https://github.com/mozilla/teach.webmaker.org/issues/856#issuecomment-104449278) in the future. If you're not sure what I mean, I can bust out a PR to your PR, and/or we can talk about it over vidyo.

toolness:
Hey @mmmavis, on the stand today hannah mentioned that this is top priority and should get finished ASAP, so I proposed actually splitting the work between you and me. If you want, you can focus on the CSS/HTML part and I can deal with productionalizing the google feed API code... let me know!

mmmavis:
Hey @toolness , I quickly refactored the blog feed loader code into a module in `/lib`.  See https://github.com/mozilla/teach.webmaker.org/commit/809171e0f20019813730510453827958af906456 (it's a squashed version of https://github.com/mmmavis/teach.webmaker.org/commit/4196d839dd9eaf87b97737896aa219781654164a and https://github.com/mmmavis/teach.webmaker.org/commit/1654564cd3e1220b5934f557027eda52c1f53f5a)  If I'm way too off track please go ahead and take that over and save me from being the blockerrrrrr :sweat_smile: :sweat_smile: :sweat_smile:

lib/blog-feed-loader.js
```diff
@@ -0,0 +1,65 @@
...
+      contentSnippet: sanitizeHtml(featured.content, {
+        allowedTags: []
+      }).split(' ').slice(0, 70).join(' ') + '...',
+      link: post.link
+    },
+    latestPostsData: latestPosts
+  };
+}
+
+var loadBlogFeed = function(requestedComp, window, feedUrl, callback) {
```
mmmavis: I had hard time naming these parameters, suggestions?

mmmavis:
I think the CSS/HTML part is nearly done.  Though I haven't got the icon buttons on hero unit to fully working.  Would like your feedback on how we should make icon colour change on hover(see https://github.com/mozilla/teach.webmaker.org/issues/856#issuecomment-105761602).

mmmavis:
I will review the CSS changes again to see if those can be better organized.

pages/home.jsx
```diff
@@ -1,19 +1,131 @@
...
+        contentSnippet: "",
+        link: ""
+      },
+      latestPostsData: []
+    }
+  },
+  componentDidMount: function() {
+    var self = this;
+    blogFeedLoader(this, window, "https://blog.webmaker.org/tag/teachtheweb/feed", function(data) {
+      self.setState({
```
mmmavis: maybe I could just do `self.setState(data);` here

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ to [Semantic Versioning](http://semver.org/).
   new releases of the site.
 - The dev version modal now links to the latest release
   of the site if the current git revision is unavailable.
+- New homepage design
```
toolness: Sorry, this is something I'm kind of a curmudgeon about--please use complete sentences in the changelog :grin:

components/icon-buttons.jsx
```diff
@@ -0,0 +1,21 @@
+var React = require('react');
+
+var IconButtons = React.createClass({
+  render: function() {
+    var links = Array.isArray(this.props.children) ? this.props.children : [this.props.children];
```
toolness: Rather than creating this and mapping over it, maybe use [`React.Children.map()`](http://facebook.github.io/react/docs/top-level-api.html#react.children.map)?

less/common.less
```diff
@@ -113,6 +122,11 @@ html.no-js .no-js-warning {
   border-top: 0.2rem solid @white;
 }
 
+#content {
+  padding-left: 0;
+  padding-right: 0;
+}
```
toolness: Hmm, unfortunately this seems to break other pages. For instance, in `/about/`, the heading "About Mozilla Learning Networks" bleeds all the way to the very edge of the screen on mobile, whereas there's normally 15px of padding being applied from bootstrap's `.col-*` rules. :cry:

lib/blog-feed-loader.js
toolness: Hmm actually I think this module should be completely decoupled from React--it's a tiny bit more inefficient in some cases, but I think it's conceptually cleaner to keep the `isMounted()` in the component that requests the blog feed.

lib/blog-feed-loader.js
```diff
@@ -0,0 +1,65 @@
+var sanitizeHtml = require('sanitize-html');
```
toolness: I appreciate the concern for security, but I'm concerned about the weight an entire HTML sanitizer will add to our JS bundle. Do you think you can just use the `feed.entries[].contentSnippet` property of the [JSON result](https://developers.google.com/feed/v1/devguide#resultJson), which doesn't contain markup? Or perhaps use some simple [HTML tag stripping code](http://stackoverflow.com/a/822464) from StackOverflow and set the content using standard JSX text syntax (i.e., not `dangerouslySetInnerHTML`)?

When we switch to getting this from the Teach API, we can potentially make sure that the server sanitizes HTML before sending it to the client. (I was hoping that the Google Feed API would do this for us, but there doesn't seem to be *any* documentation about whether the HTML is sanitized or not, which is quite annoying.)

lib/blog-feed-loader.js
```diff
@@ -0,0 +1,65 @@
...
+    });
+  }
+  return {
+    featuredPostData: {
+      title: featured.title,
+      author: featured.author,
+      publishedDate: featured.publishedDate,
+      contentSnippet: sanitizeHtml(featured.content, {
+        allowedTags: []
+      }).split(' ').slice(0, 70).join(' ') + '...',
```
toolness: Err, this could actually result in malformed HTML if the slice occurs e.g. in the middle of a tag. But [avoiding HTML entirely for now](https://github.com/mozilla/teach.webmaker.org/pull/943/files#r31177174) will obviate this.

pages/home.jsx
```diff
@@ -1,19 +1,131 @@
...
+              <span className="posted-year">{parsedDate.format("YYYY")}</span>
+            </time>
+          </p>
+        </div>
+        <div className="entry-header-container">
+          <h3 className="entry-title"><a href={this.props.data.link}>{this.props.data.title}</a></h3>
+          <cite className="author">{this.props.data.author}</cite>
+        </div>
+        <p className="excerpt">
+          {this.props.data.contentSnippet}
```
toolness: Hmm, since this is actual HTML and not text, any tags in the HTML will actually get escaped, so this will just look like raw HTML code to the user, rather than rendered HTML. I think you'll want to [dangerously set inner HTML](https://facebook.github.io/react/tips/dangerously-set-inner-html.html) here, but I recommend [avoiding HTML entirely for now](https://github.com/mozilla/teach.webmaker.org/pull/943/files#r31177174).

pages/home.jsx
toolness: You could do that, though I actually like the explicitness of listing the specific properties you need. Another alternative is [`_.pick()`](http://documentcloud.github.io/underscore/#pick) for a compromise between explicitness and verbosity, but it's really up to you.

pages/home.jsx
```diff
@@ -1,19 +1,131 @@
 var React = require('react');
 var Router = require('react-router');
 var Link = Router.Link;
+var moment = require('moment');
+var sanitizeHtml = require('sanitize-html');
```
toolness: You're only using this in the other module, not `home.jsx`...

lib/blog-feed-loader.js
mmmavis: the `contentSnippet` google returns is < 120 characters which is too short for what we need.  I'll switch over to use the stripping code (ie. `myString.replace(/<(?:.|\n)*?>/gm, '');`) instead.

toolness: Er, on the other hand, we *are* retrieving the HTML from Google, which is retrieving it from https://blog.webmaker.org, so as long as we trust those, we can trust the HTML too. I do like the idea of sanitizing the HTML for extra security, to ensure that compromising the blog doesn't also compromise teach, but the extra weight of a client-side HTML sanitizer... :sweat:

lib/blog-feed-loader.js
```diff
@@ -0,0 +1,65 @@
...
+      link: post.link
```
toolness: Should this be `featured.link`? I'm noticing when manually testing that the featured post says **What's next for Thimble?** but the link goes to **Learning Through Making: The Best Kind of Education**.

lib/blog-feed-loader.js
mmmavis: other than security concern I think escaping all HTML could

1. make our page look cleaner
2. make our contentSnippet word count logic easier :hushed: 

:no_mouth:

CHANGELOG.md
toolness: Also, the new homepage design is a BIG DEAL so you should put it at the very top of the list :dancer:

pages/home.jsx
mmmavis: heh I actually prefer what we already have here :smile:

toolness:
Hmm, I guess don't actually worry about fixing any of the feed API stuff now. I should've seen this complexity coming and had us split this up into two separate PRs, one for the HTML/CSS of the page and the other for the feed stuff. It's way too much work for one person!

toolness:
I just noticed that the rightmost icon link actually looks a bit wonky when the page is narrow:

lib/blog-feed-loader.js
mmmavis: hahah yea it should be `featured.link`.  purely a copy and replace mistake (sorry)

CHANGELOG.md
mmmavis: :+1:

less/common.less
mmmavis: arghhh, if `#content` does have left/right padding we won't be able to have a full width section with background colour (e.g., the quote section on the new homepage).  I probably have to go through all the pages and make sure thing are being wrapped within extra inner containers... :sob:

toolness:
Ok I just filed https://github.com/mmmavis/teach.webmaker.org/pull/3! If you accept that, then we can work in parallel and eventually UNITE.

toolness:
By the way, when I load your PR in Chrome, the following is logged to my browser console when viewing the fancy new homepage:

> Deprecation warning: moment construction falls back to js Date. This is discouraged and will be removed in upcoming major release. Please refer to https://github.com/moment/moment/issues/1407 for more info.

less/common.less
toolness: Oh, can you set negative margin to offset the padding? I actually did that with the hero unit for the original site, way back in February: https://github.com/mozilla/teach.webmaker.org/blob/8804e410ce96df5d2983fe829b6a81c335e94d74/styles.less#L23-L26

I'm not sure if that CSS has actually survived to present-day or not, hehe.

mmmavis: Argh, I tried the following on that full width coloured section

```
margin-left: -(@grid-gutter-width / 2);
padding-left: (@grid-gutter-width / 2);
```
This leaves a gap to the right due to the negative offset on the left.

Adding `margin-right: -(@grid-gutter-width / 2);` simply just makes the section wider but the gap is still there, which introduces a horizontal scrollbar to the page.

Also, this approach won't work on larger screen either as we can't specify how much offset we want to move. (see https://github.com/mmmavis/teach.webmaker.org/blob/develop/less/components/page.less#L21-L23)

:sob:

mmmavis:
oh darn I don't get why 
screen shot 2015-05-27 at 6 05 05 pm the text didn't get wrappppppppped.  Gonna come back to this issue tonight

mmmavis:
Hey @toolness ,

As I investigated [the margin issue](https://github.com/mozilla/teach.webmaker.org/pull/943#discussion_r31175376) I went through all the pages and wrapped sections within `<section>`.  We have done that on some pages but it's just we haven't been consistent about it.  If you wanna view the code changes you can add `?w=1` to the URL to hide all the whitespace differences in the PR.  e.g., https://github.com/mozilla/teach.webmaker.org/pull/943/files?w=1

less/components/btn.less
```diff
@@ -1,4 +1,5 @@
 .btn {
+  white-space: normal;
```
mmmavis: button text should be wrapped.  we shouldn't assume button text can always fit into one line, especially the site will be localized in the future.

mmmavis:
@toolness unit test PR is [here](https://github.com/mmmavis/teach.webmaker.org/pull/4).  Still WIP though but I had some questions for you there :information_desk_person:

less/components/btn.less
toolness: That's a [good point](https://github.com/mozilla/teach.webmaker.org/pull/943/files#r31200930), do you think you can actually add that as a comment in the code?

pages/about.jsx
```diff
@@ -10,17 +10,15 @@ var config = require('../lib/config');
...
-            <Illustration width={226} height={226}
-            className="img-circle"
-            src1x="/img/pages/about/hive-nyc-summer-quest.jpg"
-            alt="HiveLearningNYC 2014 Hive NYC Summer Quest Maker Party">
-              <p>We want more people to see themselves as citizens of the web. Mozilla Learning Networks offers programs and a global community dedicated to helping people learn the most important skills of our age: <em>the ability to read, write and participate in the digital world.</em></p>
-            </Illustration>
-          </section>
-        </div>
+      <section className="intro">
+        <h1>About Mozilla Learning Networks</h1>
```
toolness: Oof, changes like this made this PR harder to understand, with all the other changes going on in it :sob: 

If it's still possible, do you think you can actually undo these kinds of changes that are purely indentation changes due to being wrapped in a new element? We could then redo the indentation fixes in a separate PR after the homepage lands... if it's too much trouble though, don't worry about it.

toolness: Oh wait nevermind! I didn't see your note about [`?w=1`](https://github.com/mozilla/teach.webmaker.org/pull/943#issuecomment-106129354), that's awwwwwwwwweeesomeeeeeee

mmmavis:
I opened [a cleaner version of this PR](https://github.com/mozilla/teach.webmaker.org/pull/952) and it got merged.  Gonna close this one now.

PR #950, opened 2015-05-28
Add google-blog-feed-loader w/ automated and manual tests

toolness:
This adds an implementation of a google blog feed loader, based on @mmmavis's original code from #943.

Automated tests ensure that the raw feed provided by Google is processed properly.

Manual tests ensure that we can connect to the Google feed API and obtain proper results. It adds a new "Internet Services" section to the Manual Tests page:

Clicking on the "Load Blog Posts" button logs links below it, which are Data URLs containing the JSON response for testers to peruse:

We also show how many `<script>` tags pointing at Google's JS API have been loaded, which helps us ensure that we're not accidentally loading multiple script tags or anything.

### DIFF
--- a/lib/google-blog-feed-loader.js
+++ b/lib/google-blog-feed-loader.js
@@ -1,0 +1,91 @@
+var FEED_URL = 'https://blog.webmaker.org/tag/teachtheweb/feed';
+var DEFAULT_CONTENT_SNIPPET_WORD_COUNT = 70;
+var SCRIPT_SRC = 'https://www.google.com/jsapi';
+
+// http://stackoverflow.com/a/822464
+var HTML_TAG_RE = /<(?:.|\n)*?>/gm;
+
+var scriptLoading = false;
+var scriptLoadCallbacks = [];
+
+function handleScriptLoad() {
+  scriptLoadCallbacks.forEach(process.nextTick);
+  scriptLoadCallbacks = [];
+}
+
+function ensureGoogleAPI(callback) {
+  var head, gScript;
+
+  if (window.google) {
+    return process.nextTick(callback);
+  }
+  scriptLoadCallbacks.push(callback);
+  if (!scriptLoading) {
+    head = document.getElementsByTagName('head')[0];
+    gScript = document.createElement('script');
+    gScript.setAttribute('src', SCRIPT_SRC);
+    gScript.onload = handleScriptLoad;
+    head.appendChild(gScript);
+    scriptLoading = true;
+  }
+}
+
+function getBlogPosts(feedUrl, callback) {
+  var google = window.google;
+  google.load('feeds', '1', {
+    callback: function() {
+      var feed = new google.feeds.Feed(feedUrl);
+      feed.load(function(result) {
+        callback(formatBlogPosts(result.feed));
+      });
+    }
+  });
+}
+
+function makeContentSnippet(html, wordCount) {
+  var text = html.replace(HTML_TAG_RE, '');
+  var words;
+
+  wordCount = wordCount || DEFAULT_CONTENT_SNIPPET_WORD_COUNT;
+
+  words = text.split(' ');
+
+  if (words.length > wordCount) {
+    text = words.slice(0, wordCount).join(' ') + '\u2026';
+  }
+
+  return text;
+}
+
+function formatBlogPosts(feed) {
+  var featured = feed.entries[0];
+  var latestPosts = [];
+  var post;
+  for (var i = 1; i < 4; i++) {
+    post = feed.entries[i];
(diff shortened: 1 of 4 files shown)


lib/google-blog-feed-loader.js
```diff
@@ -0,0 +1,91 @@
+var FEED_URL = 'https://blog.webmaker.org/tag/teachtheweb/feed';
+var DEFAULT_CONTENT_SNIPPET_WORD_COUNT = 70;
+var SCRIPT_SRC = 'https://www.google.com/jsapi';
+
+// http://stackoverflow.com/a/822464
+var HTML_TAG_RE = /<(?:.|\n)*?>/gm;
+
+var scriptLoading = false;
+var scriptLoadCallbacks = [];
```
mmmavis: question: why do we need an `array` to store callback? when will it be the case that we have multiple callbacks?

lib/google-blog-feed-loader.js
```diff
@@ -0,0 +1,91 @@
...
+
+function handleScriptLoad() {
```
mmmavis: Should we maybe set `scriptLoading` back to `false` since the script has been *loaded*?

lib/google-blog-feed-loader.js
toolness: Hmm, I guess it's actually not *really* needed unless multiple requests to fetch the blog posts are issued almost simultaneously, which is never the case in our current site--except on the manual test page if you click the button really fast with a slow internet connection--so it's technically over-engineering. But it was easier to just make it possible rather than document the limitation, I suppose. :stuck_out_tongue: 

Also, it could be useful if we decide to use other Google JS API features in the future *other* than just the feed API, so I guess there's that. e.g., imagine if we had another component below the blog which used some other Google JS API feature--both components would request the JS API at the same time, but we'd want to ensure that only one `<script>` tag was added to the DOM, and we'd want the callbacks given to both components to be called once that script was loaded.

lib/google-blog-feed-loader.js
toolness: I was thinking of doing that, but it's not really necessary at all to make the code work, and just adds extra complexity. If anything, I guess I'd rather rename the variable to something a bit more accurate, like, um, `scriptTagInjectedIntoDOM` maybe? That would definitely make the variable less misleading.

I suppose an alternative would be to just call it `scriptEl`, initialize it to `null`, and have it be a reference to the actual `<script>` element, which we currently define as a function-level variable called `gScript` in `handleScriptLoad()`--then we can get rid of `gScript` and just test whether `scriptEl === null` instead of having the `scriptLoaded` boolean. Not a bad simplification, and makes the code more straightforward too!

mmmavis: > I suppose an alternative would be to just call it `scriptEl`, initialize it to `null`, and have it be a reference to the actual `<script>` element, which we currently define as a function-level variable called `gScript` in `handleScriptLoad()`--then we can get rid of `gScript` and just test whether `scriptEl === null` instead of having the `scriptLoaded` boolean. Not a bad simplification, and makes the code more straightforward too!

I like this!

toolness: Awsm I will do that! Heading out for the weekend but will do it first thing monday :)

lib/google-blog-feed-loader.js
mmmavis: Heh, that's what I thought too.  Just wanted verify my thinking :smiling_imp: :smiling_imp: :smiling_imp:

toolness:
Ok just fixed up the [`scriptEl` thing](https://github.com/mozilla/teach.webmaker.org/pull/950#discussion_r31371799)!

test/browser/google-blog-feed-loader.test.js
```diff
@@ -0,0 +1,114 @@
...
+  "link": "https://blog.webmaker.org",
+  "author": "",
+  "description": "People, products and programs to teach, learn and make the Web",
+  "type": "rss20",
+  "entries": [
+    {
+      "title": "What’s next for Thimble?",
+      "link": "https://blog.webmaker.org/whats-next-for-thimble",
+      "author": "Hannah Kane",
+      "publishedDate": "Tue, 12 May 2015 10:47:33 -0700",
```
mmmavis: Just a note:

I'm not sure if it's the [API doc](https://developers.google.com/feed/v1/devguide#resultJson) being out of date - it doesn't include the "day of week" bits (see `publishedDate` section).  However when I ran Google API it did return date info in this format "Tue, 12 May 2015 10:47:33 -0700".

mmmavis:
yayayay R :heavy_plus_sign:  thanks @toolness !

PR #965, opened 2015-06-01
Add front-end UI for club approval flow.

### LINKED ISSUE #920: Changes to Clubs submission success message
- [ ] Change success message copy from "We've added your Club!" to "Thanks for your submission! We'll review it and be in touch shortly."

- [ ] remove "Take me to my club" button action

toolness:
When finished, this will fix #920 and #911.

When a user creates a new club, the success message now looks like this:

Approved, pending, and denied clubs look like this in the club list, which now has a separate section for the current user's clubs:

teach-clubs-feedback

The map popup also reflects the club pending/denied status through a Bootstrap label, with more detail provided via the `title` attribute:

Still to do:
- [X] Add pending/denied information to the club list.
- [x] Add pending/denied information to the map pop-up information.
- [x] Modify the club submission success message (#920).
- [x] Add tests for `ClubStatusLabel `.
- [x] Add tests for `ClubLists`.
- [x] There's a bug where logging out will still show the unapproved clubs for the previous user. This is because our existing code assumes that the club list won't change when the current user changes; however, this assumption is broken with the club approval flow changes, so we need to account for that.

### DIFF
--- a/components/map.jsx
+++ b/components/map.jsx
@@ -9,6 +9,8 @@ var DEFAULT_STYLESHEETS = config.IN_TEST_SUITE ? [] : [
   'https://api.tiles.mapbox.com/mapbox.js/plugins/leaflet-markercluster/v0.4.0/MarkerCluster.css',
   'https://api.tiles.mapbox.com/mapbox.js/plugins/leaflet-markercluster/v0.4.0/MarkerCluster.Default.css'
 ];
+var CLUB_PENDING_TEXT = "This club is pending approval and is not visible to other users.";
+var CLUB_DENIED_TEXT = "This club has been denied approval and is not visible to other users.";
 
 var mapboxId = process.env.MAPBOX_MAP_ID || 'alicoding.ldmhe4f3';
 var accessToken = process.env.MAPBOX_ACCESS_TOKEN || 'pk.eyJ1IjoiYWxpY29kaW5nIiwiYSI6Il90WlNFdE0ifQ.QGGdXGA_2QH-6ujyZE2oSg';
@@ -38,6 +40,7 @@ function geoJSONit(data) {
       "description": i.description,
       "website": i.website,
       "location": i.location,
+      "status": i.status,
       "title": i.name
     });
   });
@@ -77,6 +80,7 @@ var MarkerPopupClub = React.createClass({
   render: function() {
     var website = null;
     var actions = null;
+    var status = null;
 
     if (this.props.website) {
       website = (
@@ -84,6 +88,12 @@ var MarkerPopupClub = React.createClass({
           {this.getWebsiteDomain()}
         </a></p>
       );
+    }
+
+    if (this.props.status === 'pending') {
+      status = <span className="label label-warning" title={CLUB_PENDING_TEXT}>pending</span>;
+    } else if (this.props.status === 'denied') {
+      status = <span className="label label-danger" title={CLUB_DENIED_TEXT}>denied</span>;
     }
 
     if (this.props.isOwned) {
@@ -105,7 +115,7 @@ var MarkerPopupClub = React.createClass({
 
     return (
       <div>
-        <b>{this.props.title}<br/></b>
+        <b>{this.props.title} {status}<br/></b>
         <i>{this.props.location}</i>
         <br/>
         <br/>
@@ -129,6 +139,8 @@ var Map = React.createClass({
     onReady: React.PropTypes.func
   },
   statics: {
+    CLUB_PENDING_TEXT: CLUB_PENDING_TEXT,
+    CLUB_DENIED_TEXT: CLUB_DENIED_TEXT,
     MarkerPopup: MarkerPopup,
     clubsToGeoJSON: geoJSONit,
     setAccessToken: function(value) {
--- a/components/map.jsx
+++ b/components/map.jsx
@@ -9,8 +9,6 @@ var DEFAULT_STYLESHEETS = config.IN_TEST_SUITE ? [] : [
   'https://api.tiles.mapbox.com/mapbox.js/plugins/leaflet-markercluster/v0.4.0/MarkerCluster.css',
   'https://api.tiles.mapbox.com/mapbox.js/plugins/leaflet-markercluster/v0.4.0/MarkerCluster.Default.css'
 ];
-var CLUB_PENDING_TEXT = "This club is pending approval and is not visible to other users.";
-var CLUB_DENIED_TEXT = "This club has been denied approval and is not visible to other users.";
 
 var mapboxId = process.env.MAPBOX_MAP_ID || 'alicoding.ldmhe4f3';
 var accessToken = process.env.MAPBOX_ACCESS_TOKEN || 'pk.eyJ1IjoiYWxpY29kaW5nIiwiYSI6Il90WlNFdE0ifQ.QGGdXGA_2QH-6ujyZE2oSg';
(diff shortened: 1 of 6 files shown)


toolness:
@mmmavis, note that to manually test this, you'll likely want to use the [clubs admin view on staging](https://teach-api.herokuapp.com/admin/clubs/club/). If you don't have admin access to that, let me know and I'll give it to you.

mmmavis:
> @mmmavis, note that to manually test this, you'll likely want to use the [clubs admin view on staging](https://teach-api.herokuapp.com/admin/clubs/club/). If you don't have admin access to that, let me know and I'll give it to you.

@toolness I actually don't.  Please grant me access!  :see_no_evil:

toolness:
Ok done!

mmmavis:
@sabrinang this is the PR @toolness just mentioned on the call

pages/clubs.jsx
```diff
@@ -52,6 +59,7 @@ var ClubListItem = React.createClass({
         <p><em>{club.location.split(',')[0]}</em></p>
         <p>{club.description}</p>
         <p><small>Led by <a href={"https://webmaker.org/en-US/search?type=user&q=" + club.owner}>{club.owner}</a></small></p>
+        {status}
```
mmmavis: I personal would like a "pending" label shown next to `<h4>{clubName}</h4>`, similar to what we have on the popup on Map.   It makes the labelling more consistent across the page and make it easier to spot "pending" clubs in a long list.

test/browser/clubs-page.test.jsx
```diff
@@ -97,6 +97,7 @@ describe("ClubsPage.ClubList", function() {
     owner: 'foo',
     website: 'http://example.org',
     location: 'Somewhere, USA',
+    status: 'approved',
     name: 'foo club'
   }];
```
mmmavis: (Github didn't let me make a comment at [line 104](https://github.com/mozilla/teach.webmaker.org/pull/965/files#diff-23085ec9521eccdf034ed187c1843e06R104), there was no "+" when I hovered over)

Maybe we can also check in the test below to see if any `<div>` with classes `.alert.alert-warning ` or `.alert.alert-danger` is presented?

mmmavis: oh wait, nvm, I think I read it wrong :)

mmmavis: urghhhh stupid me, my [question above](https://github.com/mozilla/teach.webmaker.org/pull/965/files#r31479842) is still valid.  sorry!

test/browser/map.test.jsx
```diff
@@ -129,6 +131,22 @@ describe("Map.MarkerPopup", function() {
     findButtons(popup).length.should.equal(2);
```
mmmavis: Same as https://github.com/mozilla/teach.webmaker.org/pull/965/files#r31479842 - could we also check to see if a "pending" or "denied" label is presented?

mmmavis:
Hi @toolness , I finished reviewing this PR and I think it would be great if we could double check the `pending` or `denied` message isn't presented for an already `approved` Club.  Other than that this is gooooood to go.   Wooohooooooo.  Thanks!

mmmavis:
oh wait @toolness I still need access to clubs admin view on staging :bowtie:

sabrinang:
Some feedback I have is that we can separate the clubs that an individual has added from the public club list (so it is easier to find and manage and without the anxiety of it being mixed with existing public clubs). Matching the status label seen on the map by the name on the list will be helpful visually and I incorporated icons for you to consider. Also, instead of having text to explain it is not visible to others for each club we can put it as note at the bottom of that section.
teach-clubs-feedback

test/browser/clubs-page.test.jsx
toolness: Good idea--I'd like to wait until we have the OK from sabrina on the UX, but I will definitely add that!

toolness:
Great ideas, @sabrinang! I've just implemented them, and pushed my changes (temporarily) to http://mozteach.toolness.org/clubs/. Can you take a look?

Note that I made it so the note at the bottom of the "my clubs" section doesn't display if all the user's clubs are approved. I also made the "my clubs" section not display at all if the user doesn't have any clubs (or isn't logged in). So you'll want to test those cases.

@mmmavis, I'm holding off on adding any more tests until we've got the :+1: on the UX from sabrina. Once she gives us the go-ahead, I will add the tests, but you're welcome to review my code changes in the meantime!

toolness:
Phew... ok @mmmavis I think this is ready for review. It is a bit large at this point, though, so let me know if you want to walk through it on vidyo.

components/map.jsx
```diff
@@ -70,6 +71,40 @@ var MarkerPopup = React.createClass({
...
+  propTypes: {
+    status: React.PropTypes.string.isRequired,
+    showApproved: React.PropTypes.bool
+  },
+  render: function() {
+    if (this.props.status === 'pending') {
+      return (
+        <span className="label label-warning"
+         title="This club is pending approval and is not visible to other users.">
+          pending
```
mmmavis: Since `this.props.status` is exactly the same as the actual text of the label, could we just use that variable here?  It might be my personal preference but in case we decide to change the text of the status (e.g., denied) in the future we then will have one fewer place in the code to visit.  :grin:

components/map.jsx
```diff
@@ -70,6 +71,40 @@ var MarkerPopup = React.createClass({
...
+        </span>
+      );
+    } else if (this.props.status === 'denied') {
+      return (
+        <span className="label label-danger"
+         title="This club has been denied approval and is not visible to other users.">
+          denied
```
mmmavis: same comment as https://github.com/mozilla/teach.webmaker.org/pull/965/files#r31588708

pages/clubs.jsx
```diff
@@ -502,6 +502,64 @@ var ModalAddOrChangeYourClub = React.createClass({
...
 
+var ClubLists = React.createClass({
+  propTypes: {
+    clubs: React.PropTypes.array.isRequired,
+    username: React.PropTypes.string,
+    onDelete: React.PropTypes.func.isRequired,
+    onEdit: React.PropTypes.func.isRequired
+  },
+  componentWillMount: function() {
+    this.componentWillReceiveProps(this.props);
```
mmmavis: (I was reading https://facebook.github.io/react/docs/component-specs.html)
Is it to make sure `this.state` will be updated for the initial `render` AND all the subsequent ones?

components/map.jsx
```diff
@@ -70,6 +71,40 @@ var MarkerPopup = React.createClass({
...
+        </span>
+      );
+    }
+    if (this.props.showApproved) {
```
mmmavis: Actually, could we put this in the `else` clause above(and probably include a comment to mention status is "approved" within that clause)?  That way it's more clear that there are only 3 types of states - `pending`, `denied`, and `approved`.  

I wasn't sure if `this.props.status` could be something else and had to `console.log` to confirm.  :grin:

mmmavis:
@toolness I'm DONE reviewing.  Code is pretty straightforward and everything works as expected.

I left a few comments but free feel to tackle those in another PR.  I'll let you choose :D and feel free to merge this PR afterwards.  :dancer:  Thanks!

pages/clubs.jsx
toolness: yep exactly!!

components/map.jsx
toolness: Actually, they are the same right now, but there's a semantic difference between the two--`this.props.status` is the status code that the teach API gave us, and it'll always be `"pending"` even if we decide e.g. "pending approval" is a better phrase to show users. Also, the status code is independent of localization, so once we add l10n, we'll want to change the text we show users, but not `this.props.status`. So this looks like a DRY violation right now, but it's actually not, IMO :)

components/map.jsx
toolness: That's a good point, though I also want to make sure the code is fault-tolerant so that e.g. the renderer doesn't crash if things change....

Oooo, I know!! This looks like a good situation for [`React.PropTypes.oneOf`](https://facebook.github.io/react/docs/reusable-components.html#prop-validation)!! That way if our assumptions are ever incorrect, we'll get a console message logged in dev mode, but things won't crash in production.

PR #970, opened 2015-06-01
Tag and release v0.0.8.

toolness:
### Added
- The add club modal now requires users to mark a checkbox labeled
  "I have read the Mozilla Clubs Fact Sheet".

### Changed
- The list of recent blog posts on the homepage is no longer hard-coded,
  but dynamically pulled from the blog's RSS feed via the Google Feed API.
- The add/change club modal has better prompts so content featured
  publicly feels more in fidelity with clubs.
- Pages now load faster on devices with retina displays.


toolness:
@hannahkane does this look ok to push to production?

hannahkane:
+1

PR #972, opened 2015-06-01
Use google blog feed loader on homepage.

toolness:
Oof. In #950 we added the Google blog feed loader but forgot to actually hook it up to the homepage! This does that, and fixes #963.

@mmmavis can you review real quick?

Issue #953, opened 2015-05-28
Add unit test for the new homepage


mmmavis:
We should add unit tests for the new homepage as they were not included as part of [this homepage implementation PR](https://github.com/mozilla/teach.webmaker.org/pull/952).

PR #973, opened 2015-06-01
Use google blog feed loader on homepage.

### LINKED ISSUE #963: Blog posts tagged with #teachtheweb not appearing on teach.mozilla.org
On the homepage on teach.mozilla.org, it looks like the most recently featured post is from May 12th, however there is a more recent post with the #teachtheweb tag that is not appearing.

screen shot 2015-06-01 at 9 00 16 am

When I click the "See all blog posts link" I see the Maker Party blog post from May 18th, but it's not appearing on the teach homepage. 

cc @KevZawacki

toolness:
In #950 we added the Google blog feed loader but didn't hook it up to the homepage.

I attempted to do that in #972 but messed up.

This hopefully gets it right and fixes #963.

### DIFF
--- a/lib/blog-feed-loader.js
+++ b/lib/blog-feed-loader.js
@@ -1,2 +1,6 @@
-// TODO: Replace this with a real implementation.
-module.exports = require('../test/browser/stub-blog-feed-loader');
+if (process.env.NODE_ENV !== 'production' &&
+    require('./config').IN_TEST_SUITE) {
+  module.exports = require('../test/browser/stub-blog-feed-loader');
+} else {
+  module.exports = require('./google-blog-feed-loader');
+}
--- a/pages/home.jsx
+++ b/pages/home.jsx
@@ -109,7 +109,7 @@ var BlogSection = React.createClass({
         return;
       }
       this.setState({
-        featuredPost: data.featuredPosts,
+        featuredPost: data.featuredPost,
         latestPosts: data.latestPosts
       });
     }.bind(this));
--- a/test/browser/stub-blog-feed-loader.js
+++ b/test/browser/stub-blog-feed-loader.js
@@ -1,5 +1,5 @@
 var FAKE_POSTS = {
-  "featuredPosts": {
+  "featuredPost": {
     "title": "What’s next for Thimble?",
     "author": "Hannah Kane",
     "publishedDate": "Tue, 12 May 2015 10:47:33 -0700",


lib/blog-feed-loader.js
```diff
@@ -1,2 +1,6 @@
-// TODO: Replace this with a real implementation.
-module.exports = require('../test/browser/stub-blog-feed-loader');
+if (process.env.NODE_ENV !== 'production' &&
```
toolness: This check ensures that our stub loader is never included in the production webpack bundle ([explanation](https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin)).

lib/blog-feed-loader.js
```diff
@@ -1,2 +1,6 @@
...
+    require('./config').IN_TEST_SUITE) {
```
toolness: This ensures that tests don't error if the developer is offline and will also make it easier for @mmmavis to implement #953, as the homepage component will always load the fake blog posts in tests.

toolness: I'm not a huge fan of doing this test suite conditional code here, but it makes it a lot easier to actually test our code. Bleh. We can probably revisit it and figure out a cleaner solution later, if nothing else.

lib/blog-feed-loader.js
mmmavis: I understand why this check is needed but can't figure out how it is related to https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin.  :no_mouth:  We don't define any Google API variable [here](https://github.com/toolness/teach.webmaker.org/blob/integrate-google-feed-api-for-reals-this-time-take-2/webpack.config.js#L42) anyway?

toolness: Oh it doesn't have anything to do w/ the google feed API--it's `NODE_ENV` that the define plugin is using, so basically in production builds the above conditional will become

```javascript
if (false && require('./config').IN_TEST_SUITE) {
```

which will then get optimized out, since it's logically impossible for the conditional to ever be true. This means that `require('../test/browser/stub-blog-feed-loader')` will never be seen by webpack, so the stub blog feed loader won't be included in the JS bundle.

I just verified this on my local machine by running:

```
rm -rf dist
ORIGIN=blah NODE_ENV=production gulp
grep "Last week we announced" dist/*.js
```

The `grep` won't print anything. If, however, your remove the `process.env.NODE_ENV` from the conditional, running the same commands will print something, because it'll be including the stub loader (which contains the literal text "Last week we announced"), as the optimizer isn't certain that the conditional will always evaluate to `false`.

toolness: FWIW we're also already doing this kind of thing with the [dev ribbon](https://github.com/mozilla/teach.webmaker.org/blob/develop/components/page.jsx#L9-L12):

```js
var DevRibbon = (process.env.NODE_ENV === 'production' &&
                 process.env.SHOW_DEV_RIBBON !== 'on')
                ? null
                : require('./dev-ribbon.jsx');
```

because of the way we're writing it, we'll ensure that `dev-ribbon.jsx` isn't included in production builds (well, production builds that don't also have `SHOW_DEV_RIBBON` on).

mmmavis: Thanks @toolness for the detailed explanation above :smile: 
Sorry I got confused by the link you included in [this comment](https://github.com/mozilla/teach.webmaker.org/pull/973/files#r31464648) that's why I asked.  :smile_cat:

toolness: Ah got it--sorry for the confusion!

mmmavis:
R++++ :dancer:

PR #980, opened 2015-06-02
Fixes #862 - add Protect Your Data page

### LINKED ISSUE #862: Add Privacy module to site
The Privacy module should be the second item on the /activities page, just after the Web Literacy Basics module, and just above the Lo-fi/No-fi teaching kit.

Copy/images here: https://github.com/mozilla/learning-networks/issues/161

---

**Updates (by Mavis)**
This ticket includes 2 tasks:
- [ ] create a "Protecting Your Data" page on Teach site
- [ ] on the /activities page, add that module as the 2nd item on the list (in between Web Literacy Basics and Lo-fi/No-fi teaching Kit)

mmmavis:
This fixes #862.

[ WIP ]

What's missing?
- [x] copy for the Protect Your Data section on `/activity/` page
- [x] `web-lit-basic.jsx` should use the newly created `ActivitySection` component


components/activity-section.jsx
```diff
@@ -0,0 +1,41 @@
...
+              </Illustration>
+            </div>
+          )
+        })}
+      </section>
+    );
+  }
+});
+
+module.exports = ActivitySection;
```
mmmavis: This is basically extracted from [`web-lit-basic.jsx`](https://github.com/mozilla/teach.webmaker.org/blob/develop/pages/web-lit-basics.jsx#L46-L81) so that `web-lit-basic.jsx` and `protect-your-date.jsx` can both use this component.

less/components/activity-section.less
```diff
@@ -0,0 +1,35 @@
...
+      font-size: 1.6rem;
+      line-height: 2.2rem;
+      margin-bottom: 8px;
+      &.description {
+        font-weight: 500;
+        max-width: 60rem;
+      }
+    }
+  }
+}
```
mmmavis: extracted from https://github.com/mozilla/teach.webmaker.org/blob/develop/less/pages/web-lit-basics.less

pages/web-lit-basics.jsx
```diff
@@ -1,6 +1,6 @@
 var React = require('react');
-var ImageTag = require('../components/imagetag.jsx');
```
mmmavis: this component is not used here

mmmavis:
@toolness I now start to wonder if any of images on the Teach site is *not* decorative.  Wondering if an non-empty value for `alt` is ever needed.

toolness:
Hmm, good question. I think the Mozilla wordmark at the top-left of our site isn't decorative, in that there needs to be some kind of `alt` text there to identify where the link will take the user to... but aside from that, yeah, I'm not sure if there are many images on the site that aren't purely decorative. hmmmmmmm.

pages/activities.jsx
```diff
@@ -46,13 +49,17 @@ var ActivityKit = React.createClass({
...
     var developedByElements = developedByArray.map(function (author, key) {
       author = author.trim();
       var separator = ", ";
       if (key === 0 && hasDash) {
-        separator = " - "
-      } else if (key === developedByArray.length-1) {
+        separator = " - ";
+      } else if (key === 0 && hasInPartnershipWith) {
+        separator = " in partnership with ";
+      }
```
mmmavis: I didn't wanna have a nested if-else statements here (for `key===0` cases) so I kept this separate

mmmavis:
> Hmm, good question. I think the Mozilla wordmark at the top-left of our site isn't decorative, in that there needs to be some kind of `alt` text there to identify where the link will take the user to... but aside from that, yeah, I'm not sure if there are many images on the site that aren't purely decorative. hmmmmmmm.

Yea I will keep this in mind as I develop the site.

mmmavis:
@toolness this is ready for revieWwWwW

toolness:
This is looking great!

I used PageSpeed Insights on the page and it thinks the images are optimized well, so that's cool.

When running it through tenon.io, though, it raised only one type of warning, related to the linked illustration on the left side of these blurbs:



The image on the left has `alt=""` which *seems* fine since the image is decorative, but the problem is that the image is also *hyperlinked*, which apparently means that accessibility software has to show *something* for it. But with a null `alt` attribute and no link text, the accessibility sottware needs to come up with something to say... when I tab through the page using NVDA for Windows, for instance, it narrates the hyperlink's filename without the extension, "session01-ip", which isn't very helpful.

So understandably, for that hyperlinked image, tenon.io gives this warning:



Unfortunately clicking "Recommended Fix" on tenon gets me to a paywall, so I had to do some web sleuthing. I found an [interesting post from 2008](http://accessibilitytips.com/2008/03/04/providing-link-text/) that mentions our exact problem:

> In websites offering news, it’s common for there to be a story title and an image both linking to the actual story. The design requirement, or even a tracking requirement, may force there to be two separate links, one for the story title, and one for the story image.

And its solution:

> What we need to do here instead is to populate the alt attribute with something that can be used as link text. In the case where the only appropriate (and succinct) text is the story title, it is fine to use that.
> ...
> Its true that having two links with the same perceived link text linking to the same page is a redundancy, but this is much less of an evil than a text-less link. And much less of an accessibility barrier too.

So in the case of our image above, I guess its `alt` text should be set to "IP Tracer"?

The repetition does kinda suck, and it's 2015 now, so I was hoping that we have more usable solutions at our disposal. I tried adding [`tabIndex="-1"`](http://www.paciellogroup.com/blog/2014/08/using-the-tabindex-attribute/) to the hyperlink to make it non-focusable through tabbing and that helps a bit, but it still narrates the link when narrating the page content. Setting `aria-hidden="true"` makes NVDA ignore it when narrating the page, but then if I focus on it with my *mouse* (which I think non-blind users with low vision might use a screen-reader for) then NVDA simply says "blank". Argh! It's like there's no great solution for this.

Maybe our Accessibility Hero @rossbruniges has some advice for us? :bowtie:

components/activity-section.jsx
```diff
@@ -0,0 +1,41 @@
...
+        <h2>
+          Section {this.props.title}
+        </h2>
+        {this.props.activities.map(function (activity, i) {
+          return (
+            <div className="activity" key={i}>
+              <Illustration height={122} width={122}
+              src1x={activity.image1x}
+              src2x={activity.image2x}
+              alt=""
```
toolness: As I [mentioned elsewhere](https://github.com/mozilla/teach.webmaker.org/pull/980#issuecomment-108116906), this is actually problematic for a11y because it's a hyperlinked image... one solution may be to set `alt={activity.title}`. Redundant and annoying, but better than empty link text.

components/activity-section.jsx
```diff
@@ -0,0 +1,41 @@
...
+              link={activity.link}>
+                <div className="curriculum-description">
+                  <a href={activity.link}>
+                    <h3>{activity.title}</h3>
+                  </a>
+                  <p>
+                    <em>{activity.subtitle}</em>
+                  </p>
+                  <p className="description" dangerouslySetInnerHTML={{__html: activity.description}}></p>
```
toolness: I think an alternative to :boom: **DANGEROUSLY** :boom: setting `innerHTML` here may be to actually set `activity.description` to a [`ReactElement`](https://facebook.github.io/react/docs/glossary.html#react-elements) rather than a string of HTML, e.g. 

```jsx
var activity = {
  title: "Cool Activity",
  description: <span>This activity is <em>cooooooooool</em>.</span>
};
```

Then you could actually just change line 30 to:

```jsx
<p className="description">{activity.description}</p>
```

In fact if `description` doesn't contain any markup, you could even leave it as a string, because JSX is cool that way. :sunglasses: 

I guess there probably aren't any performance benefits to doing this, but the JSX validation could help us ensure that there isn't any malformed HTML in the descriptions.

Um, that said, it could make it harder to move this kind of content into a DB and/or localize it in the future, so there's definitely trade-offs.

Meh, let's just leave it unchanged for now I guess. :expressionless:

components/activity-section.jsx
toolness: Ah excellent, thanks for keeping us [DRY](http://en.wikipedia.org/wiki/Don%27t_repeat_yourself) :+1:

pages/activities.jsx
```diff
@@ -46,13 +49,17 @@ var ActivityKit = React.createClass({
     // Displays the last author with an "and" and an oxford comma.
     var _this = this;
     var hasDash = this.props.developedBy.indexOf("-") > -1;
-    var developedByArray = this.props.developedBy.replace("-", ",").split(",");
+    var hasInPartnershipWith = this.props.developedBy.indexOf("in partnership with") > -1;
+    var developedByArray = this.props.developedBy.replace("-", ",").replace("in partnership with", ",").split(",");
```
toolness: hmmmmmmmmmmm this will get interesting when we need to localize things... :smiling_imp: I think it's ok for now tho.

pages/protect-your-data.jsx
```diff
@@ -0,0 +1,132 @@
...
+    return (
+      <div>
+        <h1>Protect Your Data</h1>
+        <section className="intro">
+          <Illustration
+          height={204} width={204}
+          src1x="/img/pages/protect-your-data/photo-protect-your-data.png"
+          src2x="/img/pages/protect-your-data/photo-protect-your-data@2x.png"
+          alt=""
+          className="img-circle">
```
toolness: It's kinda weird to me that `Illustration` generates a link with no `href` if `link` isn't passed in as a prop, but that's probably out of the scope of this PR.

pages/protect-your-data.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+
+var LearningObjectives = React.createClass({
+  render: function () {
+    return (
+      <section className="row web-lit-basics">
+        <div className="col-sm-12">
+          <h2>Learning Objectives</h2>
+          <p>
+            These activities will help your learners understand how to take control of their privacy in a networked world. Complete the activities in sequence, or mix and match. <a href="http://discourse.webmaker.org/category/clubs">Visit our discussion forum</a> to get help and share your experience.
```
toolness: The w3c [discourages verbs in link text](http://www.w3.org/QA/Tips/noClickHere), but as you can see from this comment, I do not particularly care. That said, if we wanted to be more internet savvy, we could leave "Visit our" out of the hyperlink to our discussion forum.

components/activity-section.jsx
toolness: Alternatively, since the Web Literacy Basics page has the exact same problem, we could just leave it as-is and file a separate issue for it.

toolness:
This looks good! You're welcome to change any of the minor quibbles I had and then merge this. I think we should leave the [a11y issue](https://github.com/mozilla/teach.webmaker.org/pull/980#issuecomment-108116906) for a separate bug since the Web Literacy Basics page also has the same problem, and it's not a blocker to having this on production anyways, IMO.

toolness:
Ok, I just spun the a11y thing out into its own issue: https://github.com/mozilla/teach.webmaker.org/issues/987

Feel free to merge this!

pages/activities.jsx
mmmavis: argh right, and I haven't taken localization into account yet :eyes:

pages/protect-your-data.jsx
mmmavis: Cool, I'm not aware of this.

pages/protect-your-data.jsx
mmmavis: This fix seems pretty small - let me just do it in this PR.

components/activity-section.jsx
mmmavis: Um I'm gonna leave this unchanged for now @toolness :hushed: ... since now I don't know much about our localization plan/setup (technical-wise).  Maybe we can revisit this matter later when I know more about it? :eyes:

mmmavis:
Thanks for the suggestions and feedback @toolness !  I fixed the [issue](https://github.com/mozilla/teach.webmaker.org/pull/980#discussion_r31580335) you mentioned in this [commit](https://github.com/mozilla/teach.webmaker.org/commit/e275a64ee95da6962a8cc5effca1ee466e27a29f).  Feel free to review it if you'd like to.  I *think* it is a straightforward fix so I'm gonna merge this PR first :smiling_imp: :smiling_imp: :smiling_imp:

toolness:
The commit looks good, thanks mavis!

PR #984, opened 2015-06-02
Only emit "username:change" when the username actually changes.

toolness:
In #965 I changed the behavior of TeachAPI so it re-triggers fetching of clubs whenever the current user changes, but the `username:change` event is actually fired when the current user *doesn't* change, which ultimately results in spurious re-fetches of the club list.

This fixes things so the `username:change` event is only fired when the current username *actually* changes.

### DIFF
--- a/lib/teach-api.js
+++ b/lib/teach-api.js
@@ -38,8 +38,13 @@ util.inherits(TeachAPI, EventEmitter);
 
 _.extend(TeachAPI.prototype, {
   logout: function() {
+    var currentUsername = this.getUsername();
+
     delete this.storage[STORAGE_KEY];
-    this.emit('username:change', null);
+
+    if (currentUsername !== null) {
+      this.emit('username:change', null);
+    }
     this.emit('logout');
   },
   getLoginInfo: function() {
@@ -55,7 +60,7 @@ _.extend(TeachAPI.prototype, {
   },
   getUsername: function() {
     var info = this.getLoginInfo();
-    return info && info.username;
+    return info && (info.username || null);
   },
   checkLoginStatus: function() {
     this.emit('login:start');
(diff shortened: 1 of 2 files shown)


mmmavis:
Hi @toolness!

Some Teach API questions:  How is token generated? Is it associated with username and something else like timestamp or it's solely based on username?

Just wondering why we always check if `username` matches or exists but not `token`.  Is it because username is more human readable?

toolness:
> Some Teach API questions: How is token generated? Is it associated with username and something else like timestamp or it's solely based on username?

It's just an opaque identifier generated, ultimately, by Django REST framework's [`TokenAuthentication`](http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication) system. We only generate it once, so it's actually fairly long-lived, but we can regenerate it more frequently later if we want.

> Just wondering why we always check if `username` matches or exists but not `token`.  Is it because username is more human readable?

The main reason is because we already needed a `getUsername()` method that dealt with annoying edge cases (so that the login component could display the current user), and it was just easier to re-use that than make a new method that dealt with the same kinds of edge cases.

Also, since some code (like the login component) *only* needs that information, it was simpler to just have the `null` value communicate that nobody's currently logged in, rather than e.g. communicate a boolean indicating whether the user was logged in, and a string indicating their username if so.

Does that make sense?

lib/teach-api.js
```diff
@@ -38,8 +38,13 @@ util.inherits(TeachAPI, EventEmitter);
 
 _.extend(TeachAPI.prototype, {
   logout: function() {
+    var currentUsername = this.getUsername();
+
     delete this.storage[STORAGE_KEY];
-    this.emit('username:change', null);
+
+    if (currentUsername !== null) {
```
mmmavis: I'm kinda confused here - how could a user log out if he/she hasn't logged in yet?

mmmavis: sorry I meant - when will it be the case where `currentUsername` is `null`?

toolness: Oh, we basically call `logout` from `checkLoginStatus` if the server tells us that nobody is logged in, no matter what, just to ensure that the client-side state (session storage and anyone listening for the `logout` event) is synced with server-side state. (That said, the `logout` and `login:success` *are* potentially being emitted more than they need to be, which for our purposes is OK right now.)

mmmavis: > ... just to ensure that the client-side state (session storage and anyone listening for the logout event) is synced with server-side state

gotcha!  :smile_cat:

mmmavis:
R+ yay!

Issue #974, opened 2015-06-01
When viewing /activities/web-lit-basics/, the nav state should change


hannahkane:
When viewing https://teach.mozilla.org/activities/web-lit-basics/:
 * The Clubs Toolkit link should not appear in the nav
 * ~~The Teaching Activities item in the nav should be the darker color~~ (see discussion below)

toolness:
This shouldn't be hard to fix, and I actually have no idea how this particular part of the code works, so fixing it will be a good way for me to acquaint myself with it.

toolness:
Hmm, so making the Clubs Toolkit link not appear in the nav is super easy, but your second bullet point--making the "Teaching Activities" item in the nav appear selected--is actually non-trivial, because the code we're using to make those nav items look darker is triggered by the actual URL for the nav item being the current URL that the user is viewing.

This is reinforced by the conceptual one-to-one mapping between a nav item and its URL; if the nav item appears darkened, then it's because you're already on that nav item, and clicking on it again shouldn't do anything. So in a way, having "Teaching Activities" appear selected breaks this expectation... Though I can understand *why* you want it to appear selected, as the Web Literacy Basics page is "part of" the Teaching Activities section.

Would it be really weird to have "Web Literacy Basics" be a formal subsection of the Teaching Activities section, i.e. listed below it, just like "Clubs Curriculum" used to be listed under the Mozilla Clubs section? Or alternatively, would it be ok to not have anything on the nav bar appear selected at all, just like the About page currently works?

hannahkane:
Ah, I get it. Let's stick with having nothing in the nav bar appear selected. I don't think we need to have it be a formal sub-section. That will get out of hand quickly as we add more and more modules.

toolness:
Ah cool! Ok in that case #985 is ready for review!!

mmmavis:
https://github.com/mozilla/teach.webmaker.org/pull/985 just got merged.  Guess we can close this ticket now.

PR #985, opened 2015-06-02
Hide clubs section in nav when weblit basics is active

toolness:
This is intended as a fix for #974~~, but I'm running into some issues w/ completely fixing it as per hannah's recommendation. Discussion is ongoing in #974~~.


toolness:
@mmmavis, to clarify, all this PR needs to do is make it so the "clubs" section of the nav doesn't appear expanded when the user is on the web lit basics page... no section on the nav should appear darkened/selected when the user is on the page (just like how the About page currently works).

mmmavis:
thanks for the clarification :)
and R+

PR #988, opened 2015-06-03
Tag and release v0.0.9.

toolness:
### Added
- A new "Protect Your Data" teaching activity has been added,
  and is now available at `activities/protect-your-data/`.

### Changed
- New clubs are no longer automatically publicly visible; instead, they
  are initially set to a "pending" state, and can be approved or
  denied by staff on the back-end. A "My Clubs" section above the
  rest of the clubs now informs the user about their clubs and their
  approval statuses. Users are also informed about the approval flow
  when they add a new club.
- The list of recent blog posts on the homepage is no longer hard-coded,
  but dynamically pulled from the blog's RSS feed via the Google Feed API.
  (We thought this was in 0.0.8, but it wasn't.)
- The "Learn" CTA at the bottom of the Teach Like Mozilla page
  has been changed to "Understand", with the sub-copy changed to
  "Learn more about the Web Literacy Map".
- The Home and About pages now consistently refer to "the Mozilla
  Learning Network", rather than "Mozilla Learning Networks".
- The decorative icons at the bottom CTAs of pages now have null
  `alt` text, improving accessibility.
- When sharing a Teach site page on Facebook, the page description is
  now the blurb from the About Us page ("We want more people to see
  themselves as citizens of the web...") rather than the footer text
  ("The Hive Learning Networks, stewarded by Mozilla...").
- When the Web Literacy Basics page is shown, the Mozilla Clubs section of
  the nav is no longer expanded.


toolness:
Note that we'll have to take the site down for maintenance for a few minutes when we push this to production, as it involves deploying the back-end to production--not just the front-end.

We'll also probably want to actually test out the club approval flow on staging to verify that everything is properly integrated, and we'll likely want to double-check that it works on production once we push there, too.

toolness:
Hey @hannahkane can you take a look at this? Also, as I mentioned above, if you could double-check that the new approval flow works as expected, that'd be rad.

hannahkane:
+1. All looks good.

Issue #955, opened 2015-05-28
Timestamp for latest blog post says "invalid date" until posts are loaded


toolness:
I was visiting the homepage and it was taking a little while to load. The timestamp for the latest blog post is supposed to look like this:

![2015-05-28_19-47-40](https://cloud.githubusercontent.com/assets/124687/7873635/cb9e5820-0572-11e5-9b1e-31ede7625e60.jpg)

However, because the posts weren't yet retrieved, it looked like this:

![2015-05-28_19-47-00](https://cloud.githubusercontent.com/assets/124687/7873637/e23c1c7a-0572-11e5-8e32-4af8f0977bdb.jpg)

Note that this is also what it'll look like to users without JS enabled and search engines. For the time being, you can see it at https://teach.mofostaging.net/?safemode=on.

toolness:
Hmm, I'm also noticing in #973 that when we actually replace the stub blog feed loader with the google one, this bug is actually fairly noticeable.

mmmavis:
I think this has been fixed via https://github.com/mozilla/teach.webmaker.org/commit/0e04db2f4665e3e62ff4eae01375e94545ca5762#diff-d06502677d45fab6ff321d04b9d0f54fR66 -- we render nothing unless the feature post data has been loaded.

Gonna close this ticket. :smile:

PR #952, opened 2015-05-28
CLEANER PR for #856 the new homepage implementation

mmmavis:
My [original PR](https://github.com/mozilla/teach.webmaker.org/pull/943) is getting messier and messier and it's very painful for people to understand(sorry :grimacing: ).

This PR contains minimal changes needed for the new homepage implementation and hopefully it will be a lot easier to understand.

PR #992, opened 2015-06-03
Fixes #953 - homepage unit test

### LINKED ISSUE #953: Add unit test for the new homepage

mmmavis:
This fixes #953

### DIFF
--- a/pages/home.jsx
+++ b/pages/home.jsx
@@ -35,27 +35,36 @@ var FeaturedPost = React.createClass({
   propTypes: {
     data: React.PropTypes.object.isRequired
   },
+  componentWillUpdate: function(nextProps, nextState) {
+    this.parsedMomentDate = moment(new Date(nextProps.data.publishedDate));
+  },
+  parsedMomentDate: null,
   render: function() {
-    var parsedDate = moment(new Date(this.props.data.publishedDate));
     return(
       <div className="featured-post">
-        <div className="entry-posted-container">
-          <p className="entry-posted">
-            <time className="published" dateTime={this.props.data.publishedDate} >
-              <span className="posted-month">{parsedDate.format("MMM")}</span>
-              <span className="posted-date">{parsedDate.format("D")}</span>
-              <span className="posted-year">{parsedDate.format("YYYY")}</span>
-            </time>
-          </p>
-        </div>
-        <div className="entry-header-container">
-          <h3 className="entry-title"><a href={this.props.data.link}>{this.props.data.title}</a></h3>
-          <cite className="author">{this.props.data.author}</cite>
-        </div>
-        <p className="excerpt">
-          {this.props.data.contentSnippet}
-        </p>
-        <a className="more" href={this.props.data.link}>Continue reading</a>
+        { this.parsedMomentDate ?
+          // shows this section only when featured post data has been loaded
+          <div>
+            <div className="entry-posted-container">
+              <p className="entry-posted">
+                <time className="published" dateTime={this.props.data.publishedDate} >
+                  <span className="posted-month">{this.parsedMomentDate.format("MMM")}</span>
+                  <span className="posted-date">{this.parsedMomentDate.format("D")}</span>
+                  <span className="posted-year">{this.parsedMomentDate.format("YYYY")}</span>
+                </time>
+              </p>
+            </div>
+            <div className="entry-header-container">
+              <h3 className="entry-title"><a href={this.props.data.link}>{this.props.data.title}</a></h3>
+              <cite className="author">{this.props.data.author}</cite>
+            </div>
+            <p className="excerpt">
+              {this.props.data.contentSnippet}
+            </p>
+            <a className="more" href={this.props.data.link}>Continue reading</a>
+          </div>
+          : null
+        }
       </div>
     );
   },
--- a/pages/home.jsx
+++ b/pages/home.jsx
@@ -138,7 +138,8 @@ var BlogSection = React.createClass({
 
 var HomePage = React.createClass({
   statics: {
-    pageClassName: 'home-page'
+    pageClassName: 'home-page',
+    BlogSection: BlogSection
   },
   render: function() {
     return (
--- a/pages/home.jsx
+++ b/pages/home.jsx
@@ -79,7 +79,7 @@ var LatestPosts = React.createClass({
     return (
       <ul className="recent-posts">
         {
-          this.props.data.map(function(post, i) {
+          this.props.posts.map(function(post, i) {
             return (
               <li key={i}>
                 <a className="post-title" href={post.link}>{post.title}</a>
@@ -137,7 +137,7 @@ var BlogSection = React.createClass({
             <FeaturedPost {...this.state.featuredPost} />
           </div>
           <div className="col-sm-4 col-md-4 col-lg-4">
-            <LatestPosts data={this.state.latestPosts} />
+            <LatestPosts posts={this.state.latestPosts} />
             <a className="more" href="https://blog.webmaker.org/tag/teachtheweb/">See all blog posts</a>
           </div>
         </div>
--- a/pages/home.jsx
+++ b/pages/home.jsx
@@ -33,35 +33,36 @@ var CaseStudies = React.createClass({
 
 var FeaturedPost = React.createClass({
   propTypes: {
-    data: React.PropTypes.object.isRequired
+    title: React.PropTypes.string.isRequired,
+    author: React.PropTypes.string.isRequired,
+    publishedDate: React.PropTypes.string.isRequired,
+    contentSnippet: React.PropTypes.string.isRequired,
+    link: React.PropTypes.string.isRequired
   },
-  componentWillUpdate: function(nextProps, nextState) {
-    this.parsedMomentDate = moment(new Date(nextProps.data.publishedDate));
-  },
-  parsedMomentDate: null,
   render: function() {
+    var parsedMomentDate = this.props.publishedDate ? moment(new Date(this.props.publishedDate)) : null;
     return(
       <div className="featured-post">
-        { this.parsedMomentDate ?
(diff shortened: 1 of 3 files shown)


test/browser/home-page.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+  beforeEach(function() {
+    blogSection = stubContext.render(HomePage.BlogSection);
+  });
+
+  describe("blogSection", function() {
+    it("should display featured post", function(done) {
+      process.nextTick(function() {
+        blogSection.getDOMNode().querySelector(".featured-post .entry-title").textContent.should.eql("What’s next for Thimble?");
+        done();
+      });
```
mmmavis: @toolness in [this comment](https://github.com/mmmavis/teach.webmaker.org/pull/4#discussion_r31220849) you mentioned about using sinon to trigger the feed callback.  Could you give me more pointers?

test/browser/home-page.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+var TestUtils = React.addons.TestUtils;
+
+var stubContext = require('./stub-context.jsx');
+var HomePage = require('../../pages/home.jsx');
+
+describe("HomePage", function() {
+  var blogSection;
+
+  beforeEach(function() {
+    blogSection = stubContext.render(HomePage.BlogSection);
```
mmmavis: Not too sure if I should render the entire `homePage` here or just the `blogSection`. Since this entire file is about testing the `blogSection` I chose to render the latter.   :eyes:  :no_mouth:

toolness: Yeah, I think that's actually a really good idea--it follows the general philosophy of unit testing, where you want to test each component in isolation.

I'd just rename the `describe("HomePage"` to `describe("HomePage.BlogSection"`. That way if we ever add anything else to the homepage that needs testing, we can just put it in a separate `describe` call rather than adding it into the existing one, which complicates the testing logic and makes it more confusing to understand what's going on.

test/browser/home-page.test.jsx
toolness: Hmm, so for now I think this is ok, so long as you add a comment above line 18 that explains how the `process.nextTick` is used so the stub blog post loader loads the posts by then.

However, if you *do* want to go the extra mile/kilometer, that's totally awesome! It'd likely involve the following steps:

1. Writing a fake `loadBlogPosts` function that you pass as a prop to `HomePage.BlogSection`. You could use a [sinon spy](http://sinonjs.org/docs/#spies) for this, or you could actually just hand-write a fake function that looks something like this:

  ```javascript
  var respondWithBlogPosts;
  var fakeLoadBlogPosts = function(cb) {
    /* Hold on to the callback so we can trigger it later from our tests. */
    respondWithBlogPosts = cb;
  };
  ```

2. You might want to change [`stub-blog-feed-loader.js`](https://github.com/mozilla/teach.webmaker.org/blob/develop/test/browser/stub-blog-feed-loader.js) so it also exports `FAKE_POSTS`. You'll likely want your fake callback to pass this back to the component.

That's the basics... Writing tests is mostly just about "hacking" the inputs/dependencies of your code to make sure they provide the expected output, and preferably doing so in a way that's easy to understand and maintain. Feel free to experiment with other ideas if you don't like the ones I provided.

toolness:
Could you add at least one test that verifies how the component should behave when it's waiting for blog posts to load? You can even fix #955 real quick while you're at it if you want :grin:

test/browser/home-page.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  describe("blogSection", function() {
```
mmmavis: @toolness I have this `describe` inside of the `describe("HomePage"`, does it help solving [the issue you mentioned](https://github.com/mozilla/teach.webmaker.org/pull/992/files#r31710108)? Or I could remove this inner one and rename the outer `describe` from `HomePage` to `HomePage.BlogSection`

toolness: Yeah, I recommend just removing the inner one and renaming the outer one--that way we can keep each test suite completely isolated from one another.

The only reason we might want to nest `describe` calls is if two components or "types" of test suites need to share common infrastructure--in which case the `beforeEach`/`afterEach` of the top-level `describe` could do the setup/teardown common to both sub-suites. But until we definitely have a need for that, I think we should just keep the different suites isolated from one another as much as possible. (Unfortunately, I don't think all the test suites I've written follow this principle--sorry I'm not practicing what I'm preaching... :disappointed: )

mmmavis:
@toolness I updated this PR again (with potential fix for #955 as well).  please review and let me know if it looks sane :hamster:

mmmavis:
oh yea don't forget the `?w=1` trick :grin: https://github.com/mozilla/teach.webmaker.org/pull/992/files?w=1

pages/home.jsx
```diff
@@ -35,27 +35,36 @@ var FeaturedPost = React.createClass({
...
+            <div className="entry-header-container">
+              <h3 className="entry-title"><a href={this.props.data.link}>{this.props.data.title}</a></h3>
+              <cite className="author">{this.props.data.author}</cite>
+            </div>
+            <p className="excerpt">
+              {this.props.data.contentSnippet}
+            </p>
+            <a className="more" href={this.props.data.link}>Continue reading</a>
+          </div>
+          : null
```
toolness: For non-JS users and search engines, I wonder if it might be useful to at least link to the blog instead of not showing anything at all? Guess we can just file an issue for it--and anyways, once we move to the dynamic lightweight server in #585, we will be able to fetch the blog info on the server side and deliver this component completely pre-rendered, so not really a big deal I guess.

pages/home.jsx
```diff
@@ -35,27 +35,36 @@ var FeaturedPost = React.createClass({
   propTypes: {
     data: React.PropTypes.object.isRequired
   },
+  componentWillUpdate: function(nextProps, nextState) {
+    this.parsedMomentDate = moment(new Date(nextProps.data.publishedDate));
```
toolness: Hmm, I recommend against storing variables that are used by our rendering logic directly on `this`. I *think* it's ok to store e.g. the return value of `window.setTimeout` on `this` because those values don't affect the rendering of our component, but otherwise we really want to use `this.props` and `this.state` as much as possible, because then React takes care of automatically re-rendering things for us (it's not a *huge* deal for this particular component, but as components become more complex, it will help keep things conceptually understandable and prevent bugs).

Also, assuming that a valid published date will *always* exist on the first component update seems fragile to me; why not just check during `render` to see if `props.data.publishedDate` is truthy or non-empty? It's not actually bad to do some view-related computation during `render`, especially if we use [`PureRenderMixin`](https://facebook.github.io/react/docs/pure-render-mixin.html) to speed things up so that re-renders are performed as rarely as possible. And it keeps the behavior of our components extremely simple and easy to understand.

toolness:
Ack, I'm realizing that there are a few comments I should've made on your [original PR](https://github.com/mozilla/teach.webmaker.org/pull/952) for the blog post components:

* In `FeaturedPost`, I recommend providing the data fields (like `publishedDate`, `contentSnippet`, etc) directly as props rather than in a single nested `data` object. In fact, whenever possible, we should probably prefer to *not* pass nested data structures into `props`, as it makes our code more self-documenting (if we use `propTypes`), and it makes it really easy for us to use [`PureRenderMixin`](https://facebook.github.io/react/docs/pure-render-mixin.html) to speed up our site. (That mixin only compares top-level props and doesn't dig down into sub-objects and sub-arrays.)

  Oh, also, when you use the `FeaturedPost` component in `BlogSection`, you can easily pass `this.state.featuredPost` in as a set of props by using [JSX Spread Attributes](https://facebook.github.io/react/docs/jsx-spread.html#spread-attributes). (Or you can just pass the props in explicitly, it's up to you.)

* You mentioned a little while ago that you tend to use the word `data` too much, so I'm calling you out on it here! :stuck_out_tongue: Eliminating `props.data` for `FeaturedPost` as described above would be great. And in `LatestPosts`, you might want to consider renaming `props.data` to `props.posts`, as that's a more informative name.

Anyhow, neither of these are blockers or anything, but if you want to fix them in this PR that'd be rad, or we can file a separate issue for them.

pages/home.jsx
mmmavis: Before the data has been loaded from Google API we don't have the any info about the latest blog posts, do we? :open_mouth: On the initial render we are just passing https://github.com/mozilla/teach.webmaker.org/blob/develop/pages/home.jsx#L94-L105 as `props` to `<FeaturedPost>`. 

Both of the `<FeaturedPost>` and `<LatestPosts>` won't show but there will be [a "See all blog posts" link](https://github.com/mozilla/teach.webmaker.org/pull/992/files#diff-d06502677d45fab6ff321d04b9d0f54fR140) that links to http://blog.webmaker.org/.

pages/home.jsx
mmmavis: > why not just check during `render` to see if `props.data.publishedDate` is truthy or non-empty?

This was my very initial attempt, and that didn't work for some reason I couldn't figure out.  Guess I've probably confused myself with the nested `data` object. Heh, anyways, now that it's working I'll update this PR again!

mmmavis:
> Oh, also, when you use the `FeaturedPost` component in `BlogSection`, you can easily pass `this.state.featuredPost` in as a set of props by using [JSX Spread Attributes](https://facebook.github.io/react/docs/jsx-spread.html#spread-attributes). (Or you can just pass the props in explicitly, it's up to you.)

raaaad, now I know another thing to save my keystrokes!

review again @toolness? :grimacing:

toolness:
Woot, this looks GREAT! Thanks mavis!

Issue #975, opened 2015-06-01
Expander sections should have their own URL fragments


toolness:
We're essentially using Expanders on the Clubs Toolkit and Web Literacy Map pages to make the pages look less overwhelming and a bit more digestible. However, this only underscores the fact that users should be able to link to specific parts of such pages via fragment identifiers. For instance, what if a user wanted to specifically share a link to the "Web Mechanics" section of the Web Literacy map?

It turns out this is useful for more than just users: as @ottonomy pointed out in an email, in order for Open Badges to align to the Web Literacy Map, there needs to be a canonical URL for each term in the map.

So, by adding anchor fragments for each section of the Expander, we enhance the share-ability of our site, and also make it awesome for Open Badges.

toolness:
Note that we should take care to ensure that such links work in both JS and non-JS browsers.

toolness:
Also, this might be a bit scope creepy, but I think it'd be really rad if we put some sort of visual transition on the part of the page being linked to, like Stack Overflow does in [this example](http://stackoverflow.com/questions/12643960/internet-explorer-10-is-ignoring-xmlhttprequest-xhr-withcredentials-true/16238421#16238421) (don't worry about the actual content, just notice how the specific comment I shared fades slowly from orange to white).

I find this kind of functionality particularly useful for larger screens, which contain lots of information and might make it difficult to find the relevant content that someone's just pointed you at.

ottonomy:
@toolness -- :+1: Thanks for writing up the issue. Looking forward to being able to align a badge system to the WebLitMap. 

I'm sure you could attach an animation like Stack Overflow's to a React lifecycle method for initial page load to apply to the anchor-targeted element. Working with React is pretty fun.

Let me know if you have any questions about the canonical URLs. You could even use a https://w3id.org redirect if you're at all worried that the page URL might change in the future. (And then you'd only need to make sure all the anchors remain wherever the new page was hosted)

chrislarry33:
@hannakane @ldecoursy @anmechung and I were planning a similar ticket.  Shall we plan here?

anmechung:
FYI - Hannah responded to email to Nate after she saw this ticket from @toolness. I'm guessing Atul also saw the same email that Nate sent originally to Andrew S and myself.  And yes, let's plan.  Will follow-up with Hannah et al after our meeting this week.

hannahkane:
I forwarded the email to @toolness and asked him to write the ticket.

[...]

Issue #745, opened 2015-04-20
Can't make clubs success modal go away


hannahkane:
Firefox 37

Neither the "x" nor the "Take me to my Club" button nor clicking outside of the modal made it disappear.

toolness:
So my theory here is that this is due to #712 and React's `CSSTransitionGroup` being apparently wonky with Firefox, as I mention in https://github.com/mmmavis/teach.webmaker.org/pull/1#issuecomment-93989189. If it was wonky in such a way that the "transition complete" code never fires on modal close, it'd mean that the modal would never be removed from the DOM, I think. It's just a theory, though.

It's also possible that the wonkiness is logged as a bug in React somewhere, and it's also possible that perhaps it's fixed in React 0.13.

toolness:
Hmm, it seems there are multiple [`CSSTransitionGroup`-related issues](https://github.com/facebook/react/search?q=CSSTransitionGroup&type=Issues&utf8=%E2%9C%93) filed in GitHub. I guess I recommend, regrettably, backing out #712 and seeing if this problem goes away. :/

hannahkane:
K. let's try it.

mmmavis:
I also discovered the "Add Your Club" modal is the *only* modal that's having problem with closing.  It throws console error `"transition(): tried to perform an animation without an animationend or transitionend event after timeout (5000ms). You should either disable this transition in JS or add a CSS animation/transition."` when you try to close that modal.

Googled a bit and found a potential solution mentioned in https://github.com/facebook/react/issues/1707#issuecomment-66860858.  I'm gonna try integrating [TimeoutTransitionGroup](https://github.com/Khan/react-components/blob/master/js/timeout-transition-group.jsx) and see if it fixes the problem.

toolness:
@hannahkane, this should be fixed w/ the landing of #747 (haha get it?). Travis is taking a long time to get the latest version on teach.mofostaging.net (you can see what revision it's on through the "dev version" ribbon/modal) but once it finally does, things should work! Please re-open this ticket if they don't.

hannahkane:
Good joke. :)

Will do.

:airplane:

PR #993, opened 2015-06-03
Add permalinks for Web Literacy Map expander sections

toolness:
This is a potential fix for #975. It's a bit weird because our Expander already actually behaves somewhat inconsistently, and making it accessible is not particularly straightforward.

Right now this adds a [section sign (§)](http://en.wikipedia.org/wiki/Section_sign) at the top-right of each expander section. Clicking on it will set your browser's current URL to a permalink for the section. Here's an example permalink to the Accessibility section of the Web Literacy Map, hosted on my personal deployment branch thingy:

http://mozteach.toolness.org/teach-like-mozilla/web-literacy/#accessibility

Visiting that *should* automatically expand the Accessibility section and focus keyboard navigation on it.

(You can also see what that link looks like [with JS disabled](http://mozteach.toolness.org/teach-like-mozilla/web-literacy/?safemode=on#accessibility).)

Right now this PR throws out a lot of the focus/blur stuff in the existing code because (A) it was buggy/not behaving consistently and (B) trying to make it un-buggy resulted in lots of weird annoying edge cases, such as always collapsing the current section when switching to a different application.

### DIFF
--- a/components/expander.jsx
+++ b/components/expander.jsx
@@ -43,16 +43,20 @@ var Expander = React.createClass({
   },
   handleHashChange: function() {
     if (window.location.hash === '#' + this.props.id) {
-      this.setState({
-        expanded: true,
-        attractAttention: true
-      });
-      this.attractTimeout = window.setTimeout(this.cancelAttractAttention,
-                                              ATTRACT_ATTENTION_DURATION);
+      this.attractAttention();
       this.refs.header.getDOMNode().focus();
     } else if (this.state.attractAttention) {
       this.cancelAttractAttention();
     }
+  },
+  attractAttention: function() {
+    this.setState({
+      expanded: true,
+      attractAttention: true
+    });
+    window.clearTimeout(this.attractTimeout);
+    this.attractTimeout = window.setTimeout(this.cancelAttractAttention,
+                                            ATTRACT_ATTENTION_DURATION);
   },
   cancelAttractAttention: function() {
     window.clearTimeout(this.attractTimeout);
@@ -71,8 +75,8 @@ var Expander = React.createClass({
     if (e.which === 9) {
       // We've just been focused via the keyboard. Toggling the content
       // is annoying to fiddle with via pure keyboard navigation, so just
-      // expand our content.
-      this.expand();
+      // expand our content and attract attention to it.
+      this.attractAttention();
     }
   },
   render: function() {
--- a/components/expander.jsx
+++ b/components/expander.jsx
@@ -6,7 +6,13 @@ var Expander = React.createClass({
       expanded: false
     };
   },
-  collapse: function() {
+  componentDidMount: function() {
+    if (this.props.id && window.location.hash === '#' + this.props.id) {
+      this.expand();
+      this.refs.header.getDOMNode().focus();
+    }
+  },
+  collapse: function(e) {
     this.setState({
       expanded: false
     });
@@ -22,7 +28,12 @@ var Expander = React.createClass({
     } else {
       this.expand();
     }
-    e.preventDefault();
+  },
+  handleKeyUp: function() {
+    // We've just been focused via the keyboard. Toggling the content
+    // is annoying to fiddle with via pure keyboard navigation, so just
+    // expand our content.
+    this.expand();
   },
   render: function() {
     var className = "expand-div";
@@ -31,12 +42,19 @@ var Expander = React.createClass({
     }
     return (
       <div className="expander-container">
-        <div className={className} tabIndex="0" onBlur={this.collapse} onFocus={this.expand}>
-          <h4 onMouseDown={this.handleMouseDown} className="expander-header">
+        <div className={className}>
+          <h4 ref="header" className="expander-header" id={this.props.id}
+           tabIndex="0" onKeyUp={this.handleKeyUp}
+           onMouseDown={this.handleMouseDown}>
             {this.props.head}
             <span className="ion"></span>
           </h4>
-          <div className="expander-items-container">
+          <div className="expander-items-container" onFocus={this.expand}>
+            {this.props.id
+             ? <a className="expander-permalink"
+                href={"#" + this.props.id}
+                title="Permalink to this section">&sect;</a>
+             : null}
             <div className="items-margin">
               {this.props.children}
             </div>
--- a/components/expander.jsx
+++ b/components/expander.jsx
@@ -50,11 +50,13 @@ var Expander = React.createClass({
       this.expand();
     }
   },
-  handleKeyUp: function() {
-    // We've just been focused via the keyboard. Toggling the content
-    // is annoying to fiddle with via pure keyboard navigation, so just
-    // expand our content.
-    this.expand();
+  handleKeyUp: function(e) {
+    if (e.which === 9) {
+      // We've just been focused via the keyboard. Toggling the content
+      // is annoying to fiddle with via pure keyboard navigation, so just
+      // expand our content.
+      this.expand();
+    }
   },
   render: function() {
     var className = "expand-div";
--- a/components/expander.jsx
+++ b/components/expander.jsx
@@ -31,7 +31,7 @@ var Expander = React.createClass({
     }
   },
   handleKeyUp: function(e) {
-    if (e.which === 9 && this.props.anchorId) {
+    if (e.which === 9) {
       // We've just been focused via the keyboard. Toggling the content
       // is annoying to fiddle with via pure keyboard navigation, so just
       // expand our content and attract attention to it.
--- a/components/expander.jsx
+++ b/components/expander.jsx
@@ -21,6 +21,7 @@ var Expander = React.createClass({
   },
   handleAttractAttentionToAnchor: function() {
     this.expand();
+    this.refs.header.getDOMNode().focus();
   },
   handleMouseDown: function(e) {
     if (this.state.expanded) {
@@ -30,7 +31,7 @@ var Expander = React.createClass({
     }
   },
   handleKeyUp: function(e) {
-    if (e.which === 9) {
+    if (e.which === 9 && this.props.anchorId) {
       // We've just been focused via the keyboard. Toggling the content
       // is annoying to fiddle with via pure keyboard navigation, so just
       // expand our content and attract attention to it.
--- a/components/expander.jsx
+++ b/components/expander.jsx
@@ -1,35 +1,13 @@
 var React = require('react');
 
-var ATTRACT_ATTENTION_DURATION = 4000;
+var AnchorManagerMixin = require('../mixins/anchor-manager');
 
 var Expander = React.createClass({
-  propTypes: {
-    id: React.PropTypes.string
-  },
+  mixins: [AnchorManagerMixin],
   getInitialState: function() {
     return {
-      expanded: false,
-      attractAttention: false
+      expanded: false
     };
-  },
-  componentDidMount: function() {
-    if (this.props.id) {
-      window.addEventListener('hashchange', this.handleHashChange);
-      this.handleHashChange();
-    }
-  },
-  componentWillUnmount: function() {
-    if (this.props.id) {
-      window.removeEventListener('hashchange', this.handleHashChange);
-    }
-    this.cancelAttractAttention();
-  },
-  componentDidUpdate: function(prevProps) {
-    if (process.env.NODE_ENV !== 'production') {
-      if (prevProps.id !== this.props.id) {
-        console.warn('"id" prop is expected to be constant, but changed.');
-      }
-    }
   },
   collapse: function(e) {
     this.setState({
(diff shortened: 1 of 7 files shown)


ottonomy:
Functionally, :+1: . The only thing I noticed it didn't do is if I entered a new anchor in the URL bar, it didn't expand the new section. That's an edge case though, and I don't know if it's really worth adding a handler to catch it.

Visually, one pattern that you might be interested in is like GitHub does for wiki pages, where the permalink icon is only visible when hovering/focusing the target. You can see it if you hover over the headings on [this page](https://github.com/mozilla/openbadges-specification/blob/master/Assertion/latest.md) for instance.

Thanks for getting on this so quickly! I'm excited to badge the web lit map. :)

toolness:
Thanks for the feedback @ottonomy!

> The only thing I noticed it didn't do is if I entered a new anchor in the URL bar, it didn't expand the new section. That's an edge case though, and I don't know if it's really worth adding a handler to catch it.

Yeah, I actually originally had event listeners attached to `hashchange` during component mount/unmount, but I felt a bit weird about attaching so many listeners... I also want to add exhaustive tests for all this, and doing so for `hashchange` will take time--when I realized that this is an edge case that is likely only triggered by people like us fiddling with the URL bar, I opted to simplify the code and remove it. :grin: 

That said, I'm now realizing that technically it *can* happen if someone expands a section, clicks its permalink, collapses the section, navigates back in their browser history (so the URL no longer contains the hash), and then navigates forward in their browser history. That seems kind of unlikely, but I can re-add the hashchange code if we think it's important. 

> Visually, one pattern that you might be interested in is like GitHub does for wiki pages, where the permalink icon is only visible when hovering/focusing the target.

Yeah, I actually like this pattern in general, but I wasn't sure how the UX would work here because our headers are already used as interactive elements (to expand/collapse their content), so I was worried that complicating the interactivity of that area could make things confusing.

I'm also very wary of hover states because they tend to not work well on touchscreens--I'm actually not sure how/if GitHub wiki's permalink functionality works on tablets, for instance. They also don't usually work very well for keyboard navigation users--for instance, when headings are focused via the tab key, on Firefox there isn't any visual indication that the heading is even focused, let alone that it leads to a permalink. (That said, I might be over-rotating on this accessibility stuff too, lol!)

toolness:
Oh! @ottonomy, do the individual strands (explore, build, connect) of the map need their own permalinks too? or just the competencies?

toolness:
Ahh, I ran into more funky legit edge cases where hash changes did occur. So I added that hash change code back in!

I also added the funkalicious stack-overflow-esque "make the background a bright color and slowly fade it to the normal color when the user followed a permalink here" behavior. Check it OUT!

http://mozteach.toolness.org/teach-like-mozilla/web-literacy/#credibility

ottonomy:
The hashchange feature you added works well. And you're right that it would probably occasionally be triggered by use of the back button. When JS apps can really handle history states, they feel so much more professional. The background color animation works pretty well.

On whether or not the strands need their own permalink too: I could imagine badges being aligned to completion of an entire strand, so if these URLs existed, they might be happily used. I don't know if our project in Oregon would use that feature or not. I think it comes down to a question of the meaning of the lit map. Do you think the strand itself is a node that should be able to be referenced from elsewhere, including badges? Probably. 


I agree with all your reasoning on the visual design and its accessibility of the section icon.

:clap: :clap: :clap: @toolness

toolness:
Awesome, thanks @ottonomy!! I think we can make the strands permalinkable, we just have to figure out a good place to put the actual permalink. Perhaps we'll just do the Github Wiki-style permalink-icon-on-hover thing for those?

@mmmavis, while this PR still doesn't have any tests and the design needs to be finalized, I'm interested in your feedback on the actual code--particularly the [`AnchorManagerMixin`](https://github.com/mozilla/teach.webmaker.org/pull/993/files#diff-c8271e258eb867fc3e97253c5cd10bc2) I just added.

toolness:
Phew, I think this is ready for review @mmmavis. It ended up being a bit more code than I expected, but I wanted to make sure everything was unit tested. Let me know if you want to walk through this together over vidyo!

mixins/anchor-manager.js
```diff
@@ -0,0 +1,169 @@
...
+  },
+  handleNavigateFromAnchor: function() {
+    this.cancelAttractAttentionToAnchor();
+  },
+  attractAttentionToAnchor: function() {
+    this.setState({
+      attractAttentionToAnchor: true
+    });
+    window.clearTimeout(this.attractTimeout);
+    this.attractTimeout = window.setTimeout(
```
toolness: I'm using timeouts here instead of relying on CSS transitions for the same reason we decided to use `TimeoutTransitionGroup` in #745. Basically, I don't trust CSS transition events and I think we should just use timeouts to be safe. :stuck_out_tongue:

pages/web-literacy.jsx
```diff
@@ -49,7 +49,7 @@ var ActivitiesPage = React.createClass({
         <section>
           <WebMaps>
             <WebMap head="Explore" subhead="Reading the Web">
-              <Expander head="navigation">
+              <Expander head="navigation" anchorId="navigation">
```
toolness: Note that `anchorId` is like a slug-ified version of `head`, so it seems like we might be able to determine it algorithmically to keep things DRY, but [it's a trap](https://www.youtube.com/watch?v=4F4qzPbcFiA) because we will eventually need to localize `head`, while we don't need to localize `anchorId`.

mixins/anchor-manager.js
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+};
+
+var defaultAnchorManager = new AnchorManager();
+
+var AnchorManagerMixin = {
+  propTypes: {
+    anchorId: React.PropTypes.string,
+    anchorManager: React.PropTypes.object
```
toolness: The `anchorManager` prop is only used for testing. I considered making it a [React context](http://www.tildedave.com/2014/11/15/introduction-to-contexts-in-react-js.html) but those are such a pain to test and I kinda want to hold off on using contexts more until their API is stabilized and actually documented in React. We can always migrate to using a context for this later, too.

mmmavis:
This is on my todo list today.  A Vidyo call invite might be sent from me on Monday if I'm unable to finish reviewing this today :wink:

components/expander.jsx
```diff
@@ -1,12 +1,15 @@
...
+
 var Expander = React.createClass({
+  mixins: [AnchorManagerMixin],
   getInitialState: function() {
     return {
       expanded: false
     };
   },
-  collapse: function() {
+  collapse: function(e) {
```
mmmavis: is there a reason why we need this change? :open_mouth:

components/expander.jsx
```diff
@@ -1,12 +1,15 @@
 var React = require('react');
 
+var AnchorManagerMixin = require('../mixins/anchor-manager');
+
 var Expander = React.createClass({
```
mmmavis: @toolness could you add `propTypes` here? :sunglasses:

mixins/anchor-manager.js
```diff
@@ -0,0 +1,169 @@
...
+  this.anchorMap = {};
+  this.initialized = false;
+  this.currentAnchor = null;
+  this.handleHashChange = this.handleHashChange.bind(this);
+};
+
+AnchorManager.prototype = {
+  register: function(component) {
+    var anchorId = component.props.anchorId;
+    if (process.env.NODE_ENV !== 'production' &&
```
mmmavis: question - why do we have to check `process.env.NODE_ENV`?

mixins/anchor-manager.js
```diff
@@ -0,0 +1,169 @@
+// AnchorManagerMixin is a React mixin that can be used to make the
+// component support permalinkable anchors (URL fragments) that
+// attract the user's attention to them when navigated to.
+//
+// Any component that uses this mixin supports an `anchorId` prop,
+// which is the URL fragment (without the leading hash) of the anchor.
+// This is expected to be constant throughout the lifetime of the
+// component. (If `anchorId` is undefined, the mixin will be largely
+// disabled.)
```
mmmavis: Can we make `anchorId` prop a required prop?  Or is there a reason we are making it optional?

mixins/anchor-manager.js
mmmavis: yea... React context IS confusing :stuck_out_tongue_winking_eye:

mixins/anchor-manager.js
mmmavis: Nice! And thanks for bringing up that `TimeoutTransitionGroup` issue again :+1:

components/expander.jsx
toolness: Oh, no, I think at some point in coding, I had `collapse` double as an event handler, but then i undid that and forgot to remove the args. Will fix.

mixins/anchor-manager.js
toolness: Oh, because it's just a "sanity check" intended for development, and we don't want it to negatively impact production code. React's own [prop validation](http://facebook.github.io/react/docs/reusable-components.html#prop-validation) has similar behavior:

> Note that for performance reasons `propTypes` is only checked in development mode.

mixins/anchor-manager.js
toolness: Yeah, it's because the Clubs Toolkit--which uses the Expander component, which uses this AnchorManagerMixin--isn't using it... while we could (and probably should) add permalinks to the Clubs Toolkit sections, it didn't seem like something that the Expander component should absoutely *require* in order to use, so I made it optional.

components/expander.jsx
toolness: Done! Note that `anchorId` isn't in it, because React automatically merges the `propTypes` of mixins and complains if there are any conflicts.

mmmavis: Ah, there's also `children` @toolness :wink:

test/browser/anchor-manager.test.jsx
```diff
@@ -0,0 +1,261 @@
...
+    var anchor = React.addons.TestUtils.renderIntoDocument(
+      <SampleAnchorClass anchorManager={manager} {...props}/>
+    );
+    anchors.push(anchor);
+    return anchor;
+  };
+
+  var unmountAnchor = function(anchor) {
+    React.unmountComponentAtNode(anchor.getDOMNode().parentNode);
+    anchors.splice(anchors.indexOf(anchor, 1));
```
mmmavis: ummm... why do we need to specifically search from `1`? :no_mouth:

toolness: Oh shoot, this might be a typo--I just wanted to remove the anchor from the list... I think I meant to write this:

```js
anchors.splice(anchors.indexOf(anchor), 1);
```

eep

test/browser/anchor-manager.test.jsx
```diff
@@ -0,0 +1,261 @@
...
+  };
+
+  beforeEach(function() {
+    clock = sinon.useFakeTimers();
+    manager = StubAnchorManager();
+    anchors = [];
+  });
+
+  afterEach(function() {
+    anchors.slice().forEach(unmountAnchor);
```
mmmavis: I'm curious why we need to create a copy of `anchors` here.  Would `anchors.forEach(unmountAnchor);` be sufficient?

test/browser/anchor-manager.test.jsx
```diff
@@ -0,0 +1,261 @@
...
+    var a = renderAnchor({anchorId: 'foo'});
+    manager.simulateHashChange('foo');
+    clock.tick(AnchorManagerMixin.DEFAULT_ATTRACT_DURATION + 1);
+    a.state.attractAttentionToAnchor.should.be.false;
+  });
+
+  it('waits ATTRACT_ATTENTION_TO_ANCHOR_DURATION if defined', function() {
+    var a = renderAnchor({anchorId: 'foo'});
+    a.ATTRACT_ATTENTION_TO_ANCHOR_DURATION = 10;
+    a.attractAttentionToAnchor();
```
mmmavis: What do you think about adding a `a.state.attractAttentionToAnchor.should.be.true;` check here? @toolness

test/browser/anchor-manager.test.jsx
toolness: Oh, so that's because we're iterating over an array while deleting items from it. Technically it *might* work depending on how `forEach` is implemented and/or how we're iterating over it, but I wanted to make sure that we e.g. don't skip any items, so I added the `slice` call.

test/browser/anchor-manager.test.jsx
toolness: Sure, just added it!

components/expander.jsx
toolness: Hmm, so I have concerns with how helpful that actually is, since `children` is already a documented React property and the component won't actually crash or anything if `children` isn't defined...  We've actually had react warnings pop up in other components because `children` was declared to be `propTypes.array` instead of `propTypes.node`.

Aside from that, though, because we're using some stub context helper code to create the expander component in our tests, we can't easily add children to the expanders in our tests, and I don't want warnings logged to in our tests complaining that our expanders don't have children...

mmmavis: alright, gotcha!

mmmavis:
donnnnne reviewing!  awesome work @toolness :dancer: :dancer: :dancer:

Issue #638, opened 2015-04-09
Make the site print-friendly


toolness:
Based on the conversation in https://github.com/mozilla/teach.webmaker.org/pull/582#discussion_r28011035 and https://github.com/mozilla/webmaker-curriculum/issues/99 I thought it might be useful to make the Teach site reasonably print-friendly. This shouldn't be too hard, I'm guessing, because we already support browsers that don't have JS, and many of our solutions for supporting those use cases can also apply to print.

That said, we might want to consult the teach team to see if it's even a priority, or if it's only a priority for certain pages, etc., so I'm filing this issue to deal with it at a later time. Perhaps during GROOMING.

<img src="http://media.giphy.com/media/4WO68upjIpDPO/giphy.gif">

LauraHilliger:
relates to #94 (the clubs curriculum could use your hands, stylesheet is very rudimentary IMHO)

hannahkane:
Grooming notes:
* P2: make it possible to print 
* P3: make it ideal
* Effort for P2: small to medium

PR #995, opened 2015-06-04
Make the expander and tabulator print-friendly

toolness:
This helps with the very basics of #638 by at least making expander and tabulator content completely visible (rather than collapsed) when printed. This is primarily done just by reusing our CSS for javascript-disabled browsers.

The easiest way to manually test this is with Chrome, which has a handy print preview. This is what the Web Literacy map looks like without this PR:

As you can see, this is pretty useless because the actual content is collapsed.

With this PR, it looks like this:

### DIFF
--- a/less/components/expander.less
+++ b/less/components/expander.less
@@ -1,4 +1,4 @@
-html.no-js {
+.expander-hide-nothing() {
   .expander-container {
     .expand-div {
       outline: 0;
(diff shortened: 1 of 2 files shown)


toolness:
Hey @ScottDowne, do you want to take a look at this? I think you wrote the original CSS and were confused about how to make it print-friendly, so maybe this could help. If you don't have time or aren't interested, no worries--though please let me know so I can reassign this to @mmmavis!

toolness:
Hmm I guess @ScottDowne must be busy. @mmmavis can you take a look at this one?

ScottDowne:
Hey, sorry, just saw this today.

Yeah, it looks good and simple to me.

Issue #968, opened 2015-06-01
Map popup disappears if it's too tall


toolness:
I've noticed that if a city has lots of clubs in it--or even just one club with a really long description--the map popup can become so tall that the marker cluster plugin causes the pin to collapse into a cluster when the user is viewing the top of the popup, which then causes the popup to disappear. Because mapbox immediately scrolls to the top of the popup when it's opened, this results in the popup seeming to show for a split-second and then immediately disappear.

We may want to specify a maximum height for the popup, and have its contents scroll if they overflow.

mmmavis:
I haven't been able to reproduce this, @toolness .  But yea specifying a max. height could also help with solving https://github.com/mozilla/teach.webmaker.org/issues/976 (enable vertical scrollbar for tall popup)

toolness:
Er yeah, I believe this bug will be addressed by the vertical scrollbar that will be added #976, so we might want to just mark this one as a dupe?

toolness:
Woot this just got fixed by #1001, so I'm closing it.

PR #1001, opened 2015-06-06
Fix map popup styling

### LINKED ISSUE #976: Popup info panel for Clubs Map needs better spacing and vertical scrollbar
screen shot 2015-06-01 at 3 26 31 pm
(screencap taken from https://github.com/mozilla/teach.webmaker.org/pull/965's branch)

- [ ] We should probably introduce divider of some sort to visually separate one club from another.
- [ ] a vertical scrollbar should be presented when the popup gets too long

mmmavis:
This fixes #976 and #968.

### DIFF
--- a/less/components/map.less
+++ b/less/components/map.less
@@ -4,7 +4,7 @@
   width: 100%;
   .mapDivChild {
     position: relative;
-    height: 500px;
+    height: 50rem;
   }
 }
 
@@ -19,4 +19,29 @@
 .map > div {
   position: relative;
   height: 100%;
+}
+
+.popup-clubs-list {
+  padding-left: 0;
+  margin-bottom: 0;
+  list-style-type: none;
+  li {
+    h4 {
+      margin-bottom: .5rem;
+    }
+    p {
+      font-size: 1.2rem;
+      line-height: 1.2;
+    }
+  }
+  .action-panel {
+    margin-top: 1rem;
+    text-align: right;
+  }
+}
+
+.leaflet-popup-content {
+  max-height: 48.5rem;
+  margin-top: 2rem;
+  overflow-y: auto;
 }
--- a/components/map.jsx
+++ b/components/map.jsx
@@ -59,14 +59,14 @@ var MarkerPopup = React.createClass({
   },
   render: function() {
     return (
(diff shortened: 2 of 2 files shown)


toolness:
Thanks for taking the initiative on this! This is a big styling improvement. :yum:

less/components/map.less
```diff
@@ -20,3 +20,25 @@
...
   height: 100%;
 }
+
+.marker-popup {
+  padding-left: 0;
+  margin-bottom: 0;
+  list-style-type: none;
+  li {
+    &:first-child {
+      border-top: none;
```
toolness: This doesn't seem to have any effect, I think because the `listItemBorderColorizer` used to style lists on the page has higher [CSS specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity). One gross way to fix this is by adding `!important`, but I think a better way would be by fixing #1002, which I just filed.

One option is for you to fix #1002 in this PR, and another is to simply remove the above rule and we can fix #1002 in a separate PR.

Also, personally I think it would be cool to also have `border-bottom: none` on the `:last-child`, so that the orange lines only appear *between* clubs... Applying the same styling for the clubs list below the map would be awesome too (that's actually how I originally wanted to style the clubs list, but at the time I was in a hurry and didn't want to deal with #1002, which was a prerequisite). But like you said, it's just a temporary fix--albeit one that will likely be around for a while because our priorities for the site will be elsewhere for the next several weeks at least--so it's up to you.

toolness: Oh LOL I didn't realize you marked this as an in-progress WIP! My bad!

mmmavis: Oh yea, I noticed this as well.  I think that day I jumped to another branch and just committed whatever code changes I made. I knew some more tweaks would be made! . ...so yea smile_cat:

mmmavis:
Hey @toolness I'm leaving the top/bottom border tweaks to https://github.com/mozilla/teach.webmaker.org/issues/1002 (I added a 3rd todo there).  Let's tackle the tweaks in that ticket instead :dancer:

toolness:
This looks great, thanks mavis!

PR #1018, opened 2015-06-10
Add zoom-to-location link in club list view

### LINKED ISSUE #982: Add a "show this club on the map" button to clubs in the club list view
I just realized that since we're only showing the name of the city in the club list view, it might be confusing/ambiguous to figure out exactly where some clubs are when looking at the list. Adding some kind of affordance that takes the user to the club's pin on the map, similar to the "take me to my club" button on the add/change club modal, wouldn't be hard to add, and could be quite useful.

toolness:
This fixes #982 by adding a map marker icon in front of the city name in the club list view:

> 

When a user hovers over the icon or the city name, the line is highlighted in blue (the standard link color). When they click on it, the page immediately scrolls up to the map and focuses in on the club (the behavior is identical to what happens when the user clicks "take me to my club" at the end of the club submission process).

Note that this isn't an accessible solution, as the map marker only responds to click events and isn't focusable via the keyboard, but that's because the map itself isn't particularly accessible. I guess it might be a good idea to add support for sighted but keyboard-only users, since they might still want to be able to see where in the world the club is located, but I'm not sure if there's any good solution for non-sighted users. Other than perhaps adding some [`sr-only`](http://getbootstrap.com/css/#conveying-meaning-to-assistive-technologies) text after the city name that contains the region/country information too, so it's not ambiguous to screen reader users. I guess I recommend filing separate issues for these rather than addressing them in this PR.

Also, the code here is a lot bigger than one would expect, because we have to pass the new `onZoomToLocation` prop through a few levels of components. And because our test suite sorta violated DRY, I decided to refactor things a bit while making tests still pass without raising any React warnings.

### DIFF
--- a/less/pages/clubs.less
+++ b/less/pages/clubs.less
@@ -5,3 +5,12 @@
     background-image: url(/img/pages/clubs/hero-clubs@2x.png);
   });
 }
+
+.club-location {
+  font-size: 1.5rem;
+  cursor: pointer;
+}
+
+.club-location:hover {
+  color: @link-color;
+}
--- a/pages/clubs.jsx
+++ b/pages/clubs.jsx
@@ -24,7 +24,11 @@ var ClubListItem = React.createClass({
     club: React.PropTypes.object.isRequired,
     username: React.PropTypes.string,
     onDelete: React.PropTypes.func,
-    onEdit: React.PropTypes.func
+    onEdit: React.PropTypes.func,
+    onZoomToLocation: React.PropTypes.func.isRequired
+  },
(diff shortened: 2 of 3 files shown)


mmmavis:
Awesome!  Thanks @toolness !

One UX feedback - I wonder if the `focusOnClub` method should open the popup as well.  If there are multiple pins on the map viewport, it's not very easy to spot the pin the Club is located at.

Um, I think I'll create a separate ticket for the team to discuss.  Gonna merge this PR nowww :+1:

Issue #991, opened 2015-06-03
Changes to nav items


hannahkane:
Let's not yet change the placement of the navigation on the page just yet, but we do need to add two items:
 * Tools (will eventually go to the new /tools page, designed in #887)
 * Community (will link to Discourse)

We'll also need to remove the "Maker Party" item.

The order of items should be:
 * Teaching Activities
 * Mozilla Clubs
 * Tools
 * Community
 * Teach Like Mozilla

hannahkane:
Blocked on #990

hannahkane:
OK, thinking on this more, I don't think we can afford to lose the "Maker Party" nav item until after MP has passed. That means we'll have six top-level nav items until early August, but I think it's necessary for people to be able to easily navigate to the MP page.

That means the order should be:
    Teaching Activities
    Mozilla Clubs
    Maker Party
    Tools
    Community
    Teach Like Mozilla

mmmavis:
I'm adding the "blocked" label back since this is currently blocked on [Build Tools page](https://github.com/mozilla/teach.webmaker.org/issues/990)

mmmavis:
We will also need illustrations for "Tools" and "Community" nav items before we can start making changes to the sidebar.

cc @sabrinang

hannahkane:
Moving this to the next heartbeat

mmmavis:
@hannahkane http://discourse.webmaker.org/ is the link to Discourse, correct? Just wanted to double check.

hannahkane:
Yup. Thanks for checking.

mmmavis:
![screen shot 2015-07-02 at 2 43 50 pm](https://cloud.githubusercontent.com/assets/2896608/8488623/cbd3cc50-20c8-11e5-990b-9274d0c346d4.png)

@hannahkane should we include subheads? (see "fix me" in screencap)

hannahkane:
Let's go with this:

Tools: Open source software to teach and learn the Web
Community: Connect with others on topics you care about

(Can make edits later if others would like to)

PR #1025, opened 2015-06-12
Add Tools Page

### LINKED ISSUE #990: Build Tools page
Design is in #887

mmmavis:
This fixes #990

I was gonna work on https://github.com/mozilla/teach.webmaker.org/issues/991 here as well but the illustration files are still missing.  I'm gonna open another PR once the image files are ready.


pages/tools.jsx
```diff
@@ -0,0 +1,151 @@
...
+        <ToolsSection/>
+        <section>
+          <div className="row">
+            <div className="col-sm-offset-2 col-sm-8 col-md-offset-2 col-md-8 col-lg-offset-2 col-lg-8">
+              <p className="callout-heading">Visit our <Link to="activities">Teaching Activities page</Link> for more examples of how to use these tools in your practice.</p>
+            </div>
+          </div>
+        </section>
+        <img src="/img/pages/tools/svg/icon-horizontal-divider.svg" width={292} alt="" className="center-block horizontal-divider"/>
+        <section>
```
mmmavis: This footer CTA `<section>` is exactly the same as what we have on `teach-like-mozilla.jsx`.  I wonder if we should extract the code and make it a component?  I just can't think of a good file name for it.  Thoughts? @toolness :grimacing:

toolness:
I'm noticing that in the `ToolsColumn` you're using a prop name called `desctiption`, was that intentional? Or should it be called `description`?

pages/tools.jsx
```diff
@@ -0,0 +1,151 @@
...
+        <div className="sample-activity">
+          <div className="label-container"><div className="label label-success">Sample Activity</div></div>
+          <a href={this.props.activityLink}>{this.props.activityTitle}</a>
+        </div>;
+    }
+    return (
+      <div className="col-sm-4 col-md-4 col-lg-4 tools-col">
+        <Illustration
+          link={this.props.link}
+          height={150} width={200}
```
toolness: Hmm, so this is giving the illustration an aspect ratio of 4:3, but the `webmaker@2x.jpg` is 400x336, which has an aspect ratio closer to a square, so it appears "squished" on my retina display. The non-retina logo does have the correct aspect ratio, so I suspect it looks fine on retina displays.

Hmm, it seems like there should be an easy way to simulate retina/non-retina from the browser...

pages/tools.jsx
toolness: Hmm, I guess I'm curious if it's just coincidentally the same--if we're likely to change it to diverge from Teach Like Mozilla's footer CTA section anytime soon?  If that's the case, then it might be a good idea to just leave it as is, rather than refactoring things now and then un-refactoring when we need to change things again.

If we're unsure, I think it's probably OK to just leave this as-is for now, since this is just content and not functionality.

toolness:
Aside from `desctiption` and the `webmaker@2x` aspect ratio problem, I think this looks good! Feel free to change those things and merge.

pages/tools.jsx
mmmavis: +1 on keeping it as is

mmmavis:
> I'm noticing that in the `ToolsColumn` you're using a prop name called `desctiption`, was that intentional? Or should it be called `description`?
Haha, sorry that was definitely a typo.

mmmavis:
Thanks for spotting that `webmaker@2x` ratio issue @toolness .  The original logo assets provided came with too much paddings around the actual illustrations so I had to do some image processing work with PS myself.  Apparently I failed to properly fix one of them :wink:

Issue #845, opened 2015-04-30
404 for poster image on /events/resources/


alicoding:
https://teach.mozilla.org/events/resources/

![](https://dl.dropboxusercontent.com/spa/n5v3bx9nnjkdpzf/j8mcs3b4.png)

hannahkane:
I can't reproduce this

toolness:
Hmmmmm, I can't either...

alicoding:
Weird... I still can't get the image to display here...
https://teach.mozilla.org/img/event-resources-page/resource-thumbnails-07@2x.png

toolness:
Ahhh you're right! I'm looking at it in on my retina laptop now and it's indeed not working.

This has actually been fixed in #840 and will be included in our next push to production (#847). You can see the analogous link does exist on staging: https://teach.mofostaging.net/img/event-resources-page/resource-thumbnails-07@2x.png.

So, closing this bug as it'll be fixed as soon as #847 is resolved!

PR #1026, opened 2015-06-12
Optimize hero unit images at build time

### LINKED ISSUE #946: Optimize hero unit images
As mentioned in https://github.com/mozilla/teach.webmaker.org/pull/943#issuecomment-105556993

toolness:
This fixes #946 by modifying our `copy-images` gulp task to optimize the hero unit images while copying.

This is done via a new ad-hoc gulp plugin at `lib/gulp-image-converter.js` which uses [`lwip`](https://www.npmjs.com/lwip) to convert images.

Specifically, we convert the hero unit images to JPEG files (instead of their current PNG), as [Google's Image Optimization Guide](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization?hl=en#selecting-the-right-image-format) recommends using it for photos.

We currently save the JPEGs at a quality of 70 (max is 100), but this can easily be tweaked in our gulpfile if we want to experiment with the best image-quality-to-file-size tradeoff. We can even specify a custom optimization level for different images if we want.

Currently running `gulp copy-images` by itself results in the following output:

screen shot 2015-06-12 at 11 41 13 am

So converting all the hero unit images currently takes about 1 second on my 2012 Macbook Pro. Right now, that is the main downside of doing this at build time, but perhaps when we're able to use [Git Large File Storage](https://git-lfs.github.com/), we can simply cache the converted images in the repository without bloating it.

In the future, I'd like to use this gulp plugin to also automatically generate non-retina images from our retina images, which should make it easier for us to create new content while also reducing the repository size and preventing errors like https://github.com/mozilla/teach.webmaker.org/pull/1025#discussion_r32309272 and #845.


lib/gulp-image-converter.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+    if (file.isStream()) {
+      return cb(new Error('Unable to process stream-based files'));
+    }
+    convertedFiles.push(file.relative);
+    return convert(convertOptions, file, cb);
+  }).on('end', function() {
+    var difference = _.difference(Object.keys(fileMap), convertedFiles);
+    if (difference.length) {
+      this.emit('error', new Error('Unconverted files: ' +
```
toolness: I wanted to make sure that we don't have any junk entries in the file map that we pass to this plugin, so this code verifies that all the entries in the file map were actually converted when the gulp task is done.

lib/gulp-image-converter.js
```diff
@@ -0,0 +1,62 @@
...
+      targetFormat,
+      options.formatParams || {},
+      function(err, buffer) {
+        if (err) {
+          return cb(err);
+        }
+        var newFile = new File({
+          cwd: file.cwd,
+          base: file.base,
+          path: file.path.replace(format, targetFormat),
```
toolness: Technically this will cause weird things to happen if the file path contains e.g. `.jpg` somewhere in the middle of it, rather than at the very end, but that seems quite unlikely.

gulpfile.js
```diff
@@ -121,9 +122,28 @@ gulp.task('copy-test-dirs', function() {
...
+      'img/pages/home/hero-unit.png': TO_JPEG,
+      'img/pages/activities/hero-teaching-activities@2x.png': TO_JPEG,
+      'img/pages/activities/hero-teaching-activities.png': TO_JPEG,
+      'img/pages/clubs/hero-clubs@2x.png': TO_JPEG,
+      'img/pages/clubs/hero-clubs.png': TO_JPEG,
+      'img/pages/events/hero-events@2x.png': TO_JPEG,
+      'img/pages/events/hero-events.png': TO_JPEG,
+      'img/pages/teach-like-mozilla/hero-teach-like-mozilla@2x.png': TO_JPEG,
+      'img/pages/teach-like-mozilla/hero-teach-like-mozilla.png': TO_JPEG
+    }
```
mmmavis: I wonder if renaming all hero-unit images to `hero-unit.png` and `hero-unit@2x.png` would be an option... so we can just do something like `img/pages/**/hero-unit.png`?  That way we don't have to specify hero-unit images one by one here (I'm worried we might forget to list files here when we add new hero unit images to the site)

lib/gulp-image-converter.js
```diff
@@ -0,0 +1,65 @@
+var path = require('path');
+var _ = require('underscore');
+var through = require('through2');
+var lwip = require('lwip');
+var File = require('vinyl');
```
mmmavis: `File` isn't used anywhere in this file? :eyes:

gulpfile.js
toolness: OOOoooooooooo good idea! 

Hmm do you mind if we file it as a separate issue though, and just land this as-is? I'm worried adding support for filename globbing could take a while and this PR is blocking some other optimizations I'd like to add, like #1030.

lib/gulp-image-converter.js
toolness: Ah good catch, it was being used but I stopped using it in https://github.com/toolness/teach.webmaker.org/commit/2fa210f2cd94614eac0a03be242cb87784eae427 and forgot to remove the `require`.

gulpfile.js
toolness: Actually um lemme try something

lib/gulp-image-converter.js
```diff
@@ -0,0 +1,64 @@
...
+      function(err, buffer) {
+        if (err) {
+          return cb(err);
+        }
+
+        // Note that vinyl 0.5 supports file.extname for manipulations
+        // like this, but we're not yet using 0.5 at the time
+        // of writing this, so we'll need to replace the extension
+        // ourselves.
+        file.path = require('replace-ext')(file.path, '.' + targetFormat);
```
mmmavis: @toolness do you mind assigning a var to `require('replace-ext')` and moving this to the top of the file?  That way it's easier for people to tell what modules are "required".

lib/gulp-image-converter.js
mmmavis: Yeah, double checking FTW!  :dancer: 

But since we are [pushing files to `convertedFiles`](https://github.com/mozilla/teach.webmaker.org/pull/1026/files#diff-ab99f4d93870cd7d62acc5faa2e6192cR55) before [they are actually converted](https://github.com/mozilla/teach.webmaker.org/pull/1026/files#diff-ab99f4d93870cd7d62acc5faa2e6192cR56), how do we handle the case if errors are thrown during the conversion phase?  :no_mouth:

lib/gulp-image-converter.js
```diff
@@ -0,0 +1,64 @@
...
+  var fileMap = options.files;
+  var convertedFiles = [];
+
+  return through.obj(function(file, enc, cb) {
+    var posixRelative = file.relative.replace(/\\/g, '/');
+    var convertOptions = fileMap[posixRelative];
+    if (!convertOptions) {
+      return cb(null, file);
+    }
+    if (file.isStream()) {
```
mmmavis: @toolness when will it be the case that `file` happen to be "stream-based"?  Mind coming up with an examples for me?

lib/gulp-image-converter.js
toolness: Ah ok... I wanted to keep it close to the one place we use it just b/c eventually I'd like to remove it in favor of `file.extname`, and that way I won't forget to remove the `require`, but I do like the convention of putting all the `require`s at the top, so I'll do that.

mmmavis: Cool thanks Atul!  Whichever works better for you I guess.  It's just I tend to miss this kind of inline todo notes myself so yea... as long as you remember :grin:

toolness:
Ok, I've changed some stuff! Inspired by [your glob idea](https://github.com/mozilla/teach.webmaker.org/pull/1026#discussion_r32451258), I've moved all the file patterns into a new reloadable module called `image-convert.config.js`. (I was thinking of having it just be a JSON file, but I suspect we are going to want to be able to have constants and stuff in there to be DRY, so I made it a really simple reloadable module instead.)

You should be able to edit `image-convert.config.js` on-the-fly during `gulp watch` and it should automatically re-convert the images. So e.g. you can play around with different values for `quality` and then reload the page to see what the new settings look like! Neato.

Other things to try:
* During `gulp watch`, add a syntax error to `image-convert.config.js` and see what happens.
* During `gulp watch`, add a pattern that doesn't match any files in the `img` directory and see what happens.

toolness:
If the general structure of `image-convert.config.js` looks okay to you, I should probably go ahead and document the structure of its `patterns` export so that it's easy for anyone to add a new entry there and stuff.

lib/gulp-image-converter.js
toolness: Oh, yeah, basically gulp's vinyl file objects can have their contents be either buffers or streams... by default they're buffers, but sometimes, if you're dealing with super massive files that can't fit into memory, or if your plugin uses libraries that work with streams instead of buffers, it may be better to use streams instead. That can be done from the gulpfile by writing e.g. `gulp.src('*.js', { buffer: false })`.

In our case, we're using buffers because that's what `lwip` operates on, so it's just easiest to write a plugin that only works with buffers (especially since we don't currently have any plans to make this plugin reusable by other folks).

Anyways, since we don't pass `{ buffer: false }` into the stream that we give to this plugin in our gulpfile, the error will never get raised, but it's useful for documentation purposes to keep that assertion in there, like the [gulp docs on using buffers](https://github.com/gulpjs/gulp/blob/master/docs/writing-a-plugin/using-buffers.md) do.

mmmavis:
> If the general structure of `image-convert.config.js` looks okay to you, I should probably go ahead and document the structure of its `patterns` export so that it's easy for anyone to add a new entry there and stuff.
Yea, it looks great!  Please go ahead and work on the documentation.

I'm going through the updated `gulp-image-converter.js` and will finish reviewing this PR soon :bowtie:

lib/gulp-image-converter.js
```diff
@@ -0,0 +1,74 @@
...
+  var matchers = Object.keys(options).map(function(pattern) {
+    return {
+      pattern: pattern,
+      convertOptions: options[pattern],
+      minimatch: new Minimatch(pattern),
+      matches: 0
+    };
+  });
+
+  return through.obj(function(file, enc, cb) {
```
mmmavis: what does `through.obj(function(file, enc, cb)` do?  Can't find the docs on https://github.com/dominictarr/through :no_mouth:

mmmavis: ah, nvm, should be looking for [`through2`](https://github.com/rvagg/through2) not `through` :p

mmmavis:
@toolness other than the question I have in [this comment](https://github.com/mozilla/teach.webmaker.org/pull/1026#discussion_r32464803) this looks awesome and is ready to be merged!

lib/gulp-image-converter.js
toolness: Oh, if any errors occur, we just pass them as errors to our callback in `convert`, which ultimately gets converted to an `error` event in the gulp stream and makes gulp complain/explode. :)  Hopefully that should never happen, though!

toolness:
Woot thanks mavis, I just answered your question [here](https://github.com/mozilla/teach.webmaker.org/pull/1026#discussion_r32474785) and will merge!

mmmavis:
okie dokie thanks @toolness :dancers: :dancers: :dancers:

Issue #151, opened 2015-03-05
Determine browser support


toolness:
It's unclear what browsers we're trying to support.

As noted in #143, I'm assuming at the moment that we'll essentially not progressively enhance anything below IE9 with JavaScript, meaning that we can use modern JS libraries for our client-side code.

However, this also implies that browser-side JS will *only* be used for progressive enhancement, and not core functionality. Looking at the current [roadmap](https://wiki.mozilla.org/Learning/Networks/Product-Roadmap) and mock-ups, I'm not sure how well we'll be able to keep this promise, at least not without migrating from our current static site generation model to one of dynamic server-side page generation.

We should talk about this soon!

mmmavis:
im also wondering if flexbox is safe to use ([ref](http://caniuse.com/#search=flexbox))

alicoding:
Can we get the data from our user on webmaker.org at least for /teach to see what is their browser? @adamlofting can we get specific details for that? @hannahkane is this something we can come up with minimum browser support? This will really help us through development if we know which browser we want to support.

hannahkane:
I don't have a lot of data to inform this decision. I would default to whatever browsers we support for Webmaker. Would welcome feedback from others, though.

gesa:
In general, Webmaker supports current minus one, which translates to, for example:
![](http://snaps.akibraun.com/4kesc.png)
(i just grabbed a random screenshot from caniuse to show which versions are current)

gesa:
So this kind of wandered back in to my mind and think it's as good a time as ever to talk about whether we support Android 4.2 in our general Webmaker properties. (we **do** target it on Webmaker App)

alicoding:
I don't think we need to worry about Android Web view? Do we? Since we are not bundling into an app? Or are we planning to?

gesa:
oh right that's just webview that is evil, not chrom(e | ium)

davidascher:
We will hopefully get visitors from emerging markets who are likely to run oldish androids and may very well have stock browsers (not Chrome). If that becomes a real blocker in terms of our css feature set for example, we could UA-detect and suggest they upgrade to either Firefox or Chrome, assuming those actually work on 4.2.

Alternatively, a UA detection early in the process that renders a specially-made-for-that audience set of pages, but treat it like an entirely different responsive view.

In general, I suspect we need to file a separate issue on what interactions we want people to be able to have on this site when coming from a mobile browser.

hannahkane:
[...]

PR #1034, opened 2015-06-16
[WIP] Some accessibility fixes for modals

toolness:
This makes some a11y fixes for our modals (some specific to club modals):

* The close box for our modals was just being announced as "button"; now it's "Close button".
* The "a network error occurred" and "your submission has some problems" alerts on our modals aren't announced by screen readers because they don't have `role="alert"` on them, which is now fixed. Also, NVDA doesn't seem to announce any `<ul>`'s within alerts, so I had to convert the lists to paragraphs. Which seems dumb, but I guess it's better than not announcing them at all?
* The "Add Your Club" button on the Clubs page was a link instead of a `<button>`, and because it didn't have a `href` attribute, it wasn't focusable with the keyboard. I changed it to a `<button>`.
* The club location autocomplete field now has a `<label>` associated with it.
* Changes between individual steps of the modal, e.g. going from the "add your club" form step to the "thanks for your submission" step, weren't announced by screen readers. We now define [ARIA Live Regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) so this is fixed.
* The way input focus was being handled between opening and closing the modal, and navigating between the individual steps in a modal, was terrible. I had no idea where my input focus was during these transitions, and often it was in an invalid place, e.g. still on the "Add My Club" button when the "Add Your Club" modal is in focus. I've fixed this so that, at the very least, the modal itself becomes focused when shown, and the DOM element that had focus before it was shown is re-focused when the modal closes. Also, a `StepView` component for multi-step modals has been added, which manages the auto-focusing of elements as users transition into a new step.

Other notes:

* The club location autocomplete field isn't very accessible to screen-readers, though at least it's fully usable via keyboard. Not really sure if this can be fixed without altering [`react-select`](https://github.com/JedWatson/react-select) and there don't seem to be any issues in that project about accessibility, aside from the delete button being pronounced "times". 

Things still left to do:

* I should add unit tests for some of this stuff, but I'm holding off on it until we've done some QA and verified that this is the behavior we want.


toolness:
@mmmavis can you take a look at this using VoiceOver and verify that these changes work OK? I've only tested on NVDA so far and could use a second opinion.

mmmavis:
yup yup! should i check out this PR or the `develop` branch?

toolness:
oh this pr please. you can compare how it works on screenreaders/with keyboard against staging if you like.

toolness:
Hey any chance you could get around to testing this on VoiceOver soon? If you think the fixes are usable then I will bust up unit tests!

mmmavis:
@toolness oh yea sorry, checking it out now

mmmavis:
> This makes some a11y fixes for our modals (some specific to club modals):
* The close box for our modals was just being announced as "button"; now it's "Close button".

:+1: with VoiceOver

> * The "a network error occurred" and "your submission has some problems" alerts on our modals aren't announced by screen readers because they don't have `role="alert"` on them, which is now fixed. Also, NVDA doesn't seem to announce any `<ul>`'s within alerts, so I had to convert the lists to paragraphs. Which seems dumb, but I guess it's better than not announcing them at all?

VoiceOver reads it as "Window" and I can't get inside of the div :disappointed:  It's the same if I change them back to list form. 

> * The "Add Your Club" button on the Clubs page was a link instead of a `<button>`, and because it didn't have a `href` attribute, it wasn't focusable with the keyboard. I changed it to a `<button>`.
* The club location autocomplete field now has a `<label>` associated with it.
* Changes between individual steps of the modal, e.g. going from the "add your club" form step to the "thanks for your submission" step, weren't announced by screen readers. We now define [ARIA Live Regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) so this is fixed.

After the form has been successfully submitted, VoiceOver reads the button text "Take Me To My Club" and skips everything else.  :confused: 

> * The way input focus was being handled between opening and closing the modal, and navigating between the individual steps in a modal, was terrible. I had no idea where my input focus was during these transitions, and often it was in an invalid place, e.g. still on the "Add My Club" button when the "Add Your Club" modal is in focus. I've fixed this so that, at the very least, the modal itself becomes focused when shown, and the DOM element that had focus before it was shown is re-focused when the modal closes. Also, a `StepView` component for multi-step modals has been added, which manages the auto-focusing of elements as users transition into a new step.

Works :+1: with VoiceOver and keyboard.

toolness:
Ok, so [`role="alert"` is extremely frustrating](http://www.paciellogroup.com/blog/2012/06/html5-accessibility-chops-aria-rolealert-browser-support/) because it works *so differently* across different browser/screenreader combinations that getting it to work in one makes it break in another. In https://github.com/mozilla/teach.webmaker.org/commit/8720213ff825d00a293803fe1b66323fff5bc3a4 I implemented a weird, kludgy work-around that got things to work on OS X Safari+Chrome, but that broke things in NVDA on Windows, so I reverted the changes in https://github.com/mozilla/teach.webmaker.org/commit/13142d4509a1d61f434a8a05369b8244bb047bae, as it's easier to just keep the code as simple as possible and rely on UA vendors to make `role="alert"` work in a consistent way that doesn't drive developers insane.

Since the current PR doesn't *degrade* VoiceOver performance but improves NVDA performance, I think we should leave the behavior as-is and just add unit tests for what we've got here and land these changes. What do you think @mmmavis ?

toolness:
Hey @mmmavis, any idea when you might be able to get to this?

mmmavis:
Sorry @toolness I missed [your ping earlier](https://github.com/mozilla/teach.webmaker.org/pull/1034#issuecomment-116667642).  I will get to this today :bowtie: (btw, is this still WIP? :grin: )

toolness:
Yeah, it's still WIP, I just wanted to know what your opinion on [my last comment](https://github.com/mozilla/teach.webmaker.org/pull/1034#issuecomment-116667642) was. If you're ok w/ the current behavior then I will start writing tests!

mmmavis:
cool, i'll check this out after the standups.

mmmavis:
Hey @toolness , thanks for sharing links to those useful posts. :wink:  It's very challenging to make a11y work across browsers/platform/screen readers so I'm fine with your solution here. (thanks for working on this!)

I know we haven't decided on [our browser support](https://github.com/mozilla/teach.webmaker.org/issues/151) but a11y-wise I'm wondering if we should focus on just a few platform combos(e.g., FF on Mac OSX with VoiceOver, Chrome on Windows with NVDA... something like that).  It's :disappointed: to say this but if we are too committed to making the site widely accessible I'm worried it will lead to dev overhead. :no_mouth:

Pomax:
closing, as the codebase has passed this PR by significantly.

Issue #908, opened 2015-05-14
Design low bar CTA for "Pledge to teach"


hannahkane:
Context: right now, the only incentive for someone to sign-up is to add their Club, which is a high bar ask.. In the future, we'll have many more reasons to sign up (including low-bar options).

In the meantime, we'd like to capture site visitors who need something low bar to entice them. We're thinking a "Pledge to Teach" that requires them to provide their email address would be a good option.

Deliverable:
UX/UI for a pledge, that integrates into the new homepage.

@ldecoursy to provide copy for pledge and @amirad to provide copy for transactional email that follows

amirad:
Timeline for email to be completed?
I think we need their name too for communication or data purposes?
@ldecoursy thoughts on "pledge to teach" versus "become a web champion" - we talked about this in our working group yesterday.

hannahkane:
Would be good to have all copy finalized by the start of the next heartbeat (Monday, May 25).

re: getting their name—is it absolutely necessary? not sure the value of being able to use their name in an email outweighs the value of having a dead simple form. having said that, it could be kind of cute in the context of a pledge. "I, [fill in your name], pledge to..."

hannahkane:
Brainstorm just now: perhaps the three main CTAs should actually be Pledge, Teach, and Start a Club, since soon (after #736), we'll be promoting Maker Party in a special way anyway.

chrislarry33:
I LOVE this

hannahkane:
Note: will need an icon in the style of the other CTAs

ldecoursy:
@sabrinang to mock up form for when someone clicks Pledge

ldecoursy:
Data to go into BSD list - @amirad do we want to add to existing list or create new one?

sabrinang:
An idea for the Pledge to Teach icon for low bar CTA in a similar style with the other icons:
![screen shot 2015-05-25 at 3 30 56 pm](https://cloud.githubusercontent.com/assets/2492510/7802815/19f17dd6-02f3-11e5-97db-e80414fc9c95.png)
and when we promote Maker Party an idea is that we can have some banner below:
![teach_homepage_makerparty_edition-03](https://cloud.githubusercontent.com/assets/2492510/7802813/fe4d3d90-02f2-11e5-85f8-4207aace5274.png)

sabrinang:
@ldecoursy - let me know when you have copy for the pledge form and I can mockup the page

amirad:
[...]

Issue #1016, opened 2015-06-10
[ Pledge Workflow Part 3 ] Form submission and BSD list


mmmavis:
Depends on [[ Pledge Workflow Part 2 ] Build modals](https://github.com/mozilla/teach.webmaker.org/issues/1015)

Once https://github.com/mozilla/teach.webmaker.org/pull/1028 is landed on staging, verify the following
 - [x] Upon submission, the user's email address should go into a BSD list
 - [x] Should trigger transactional email (see copy in https://github.com/mozilla/teach.webmaker.org/issues/1016#issuecomment-111628846 and https://github.com/mozilla/teach.webmaker.org/issues/1016#issuecomment-113293385)

---

If all looks good,
- [x] push to PROD
- [x] Edit the existing [Mentor Signup Form](https://sendto.mozilla.org/page/s/mentor-signup) on BSD so that it sends proper transactional email and its redirect URL goes to https://teach.mozilla.org/?pledge=thanks

---

meta bug for Pledge Workflow: https://github.com/mozilla/teach.webmaker.org/issues/978

mmmavis:
The following are the required fields we need for the transaction email sent by BSD.

![screen shot 2015-06-12 at 2 52 17 pm](https://cloud.githubusercontent.com/assets/2896608/8140714/f7e49682-1112-11e5-9d65-535ce52d8863.png)


From https://teach.etherpad.mozilla.org/clubtransactionalemails

**Subject line**
```
Thanks for pledging to Teach the Web!
```

**Email body**
```
Welcome! We're thrilled that you've made a commitment to the open Web and to sharing your knowledge with others. Thanks for taking the first step to keep the Web an open public resource for all. 

Now, we'd love to learn a little more about you so we can help you teach the Web! 

Please fill out this short survey: http://bit.ly/1F4jomG

Mozilla Learning Networks offers:
- Free and open source activities and lesson plans for teaching web literacy skills
- Opportunities to develop your own leadership and mentoring skills
- A forum to connect with others around the globe who care about teaching digital skills
- Best practices for leading hands-on, production-centered events and programs

Tell us more about yourself so we can share the most relevant opportunities and resources: http://bit.ly/1F4jomG. Of course, we do respect your privacy* and honor the information you're sharing with us.

If you have questions, we’re here to help. Follow us @Mozteach and use the #teachtheweb hashtag to find us. 

Michelle Thorne
on behalf of the Mozilla Learning team
```
[...]

Issue #900, opened 2015-05-13
Implement "pre/during MP" state for /events 


hannahkane:
Pending #843

toolness:
This will be interesting to implement from a technical standpoint... At first I was thinking that we could control it entirely through an environment variable, which would involve steps similar to what mavis did for #894.

But at the same time I think it'd be great for QA folks to be able to switch between the different states on the staging site, perhaps via a rudimentary UI on the dev ribbon modal. So we could have the "default" state be set via environment variable, but allow the state to be changed through developer-only UI hooks in the site, so that the different configurations can easily be verified during development.

hannahkane:
This is no longer blocked by design.

sabrinang:
maker-party_logo.png & maker-party_illustration_globe.png found here: https://drive.google.com/a/mozilla.com/folderview?id=0B4Q8pzCpDS_efkJQcEJseDlNOUlMUGNhdWwzMVFkcWp1anh6NEVZX0hMT1hiWTk5TExVWnM&usp=sharing
![maker-party_logo](https://cloud.githubusercontent.com/assets/2492510/8242341/868105b0-15c3-11e5-9342-fc400976834b.png)
![maker-party_illustration_globe](https://cloud.githubusercontent.com/assets/2492510/8242345/8f23ba14-15c3-11e5-9178-f730863975c5.png)

repeating blue MP border pattern: https://drive.google.com/a/mozilla.com/file/d/0B4Q8pzCpDS_eUENsWU5VekM2SlE/view?usp=sharing
![maker-party_border-pattern](https://cloud.githubusercontent.com/assets/2492510/8242352/99cc1222-15c3-11e5-91d0-7386d6143175.png)

Maker Party photo thumbnails found here: https://drive.google.com/a/mozilla.com/folderview?id=0B4Q8pzCpDS_efnhlekE4NElUSEtWWU9qUzF3UGRNa3VaYUJrUHVDaUZ4QnV0YlkxTDVUU2c&usp=sharing
![lofi-maker-party](https://cloud.githubusercontent.com/assets/2492510/8242353/9e2d1eb0-15c3-11e5-89a5-7369afb9943c.png)
![make-remake-hackathon](https://cloud.githubusercontent.com/assets/2492510/8242354/9f8516a0-15c3-11e5-83b0-e734769d4739.png)
[...]

PR #1039, opened 2015-06-17
Pledge modals for the pledge workflow on Homepage

### LINKED ISSUE #1015: [ Pledge Workflow Part 2 ] Build modals
Please create a branch against the **[`pledge-workflow` branch](https://github.com/mozilla/teach.webmaker.org/tree/pledge-workflow)**, not `develop`.  :+1: ヽ( ^ω^ ゞ )

---
Depends on [[ Pledge Workflow Part 1 ] Rearrange CTA buttons on Homepage](https://github.com/mozilla/teach.webmaker.org/issues/1014)
 - [ ] Implement modal pledge (design see #908)
 - [ ] Implement Success modal including Twitter and Facebook sharing functionality (design see #908)
 -  [ ] ~~Load FontAwesome (pending @sabrinang 's decision on https://github.com/mozilla/teach.webmaker.org/issues/939)~~ (created a separate issue for this task https://github.com/mozilla/teach.webmaker.org/issues/1035)

---

meta bug for Pledge Workflow: https://github.com/mozilla/teach.webmaker.org/issues/978

mmmavis:
This fixes #1015 
 - [ ] Implement modal pledge (design see #908)
 - [ ] Implement Success modal including Twitter and Facebook sharing functionality (design see #908)

### DIFF
--- a/components/icon-button.jsx
+++ b/components/icon-button.jsx
@@ -2,25 +2,25 @@ var React = require('react');
 var Router = require('react-router');
 var Link = Router.Link;
 var ImageTag = require('./imagetag.jsx');
+var _ = require('underscore');
 
 var LinkSwap = React.createClass({
   propTypes: {
     linkTo: React.PropTypes.string,
-    href: React.PropTypes.string,
-    onClick: React.PropTypes.func
+    href: React.PropTypes.string
   },
   render: function() {
-    var classes =  "btn btn-awsm";
+    var otherProps = _.extend(_.omit(this.props, "linkTo", "href"), { className: "btn btn-awsm" });
     // Swap out Link or a simple anchor depending on the props we have.
     if (this.props.linkTo) {
       return (
-        <Link to={this.props.linkTo} className={classes} onClick={this.props.onClick}>
+        <Link to={this.props.linkTo} {...otherProps}>
           {this.props.children}
         </Link>
       )
     }
     return (
-      <a href={this.props.href} className={classes} onClick={this.props.onClick}>
+      <a href={this.props.href} {...otherProps}>
         {this.props.children}
       </a>
     )
--- a/components/icon-button.jsx
+++ b/components/icon-button.jsx
@@ -4,18 +4,23 @@ var Link = Router.Link;
 var ImageTag = require('./imagetag.jsx');
 
 var LinkSwap = React.createClass({
+  propTypes: {
+    linkTo: React.PropTypes.string,
+    href: React.PropTypes.string,
+    onClick: React.PropTypes.func
+  },
   render: function() {
     var classes =  "btn btn-awsm";
     // Swap out Link or a simple anchor depending on the props we have.
     if (this.props.linkTo) {
       return (
-        <Link to={this.props.linkTo} className={classes}>
+        <Link to={this.props.linkTo} className={classes} onClick={this.props.onClick}>
           {this.props.children}
         </Link>
       )
     }
     return (
-      <a href={this.props.href} className={classes}>
+      <a href={this.props.href} className={classes} onClick={this.props.onClick}>
         {this.props.children}
       </a>
     )
@@ -29,12 +34,13 @@ var IconButton = React.createClass({
     imgSrc: React.PropTypes.string.isRequired,
     imgSrc2X: React.PropTypes.string,
     head: React.PropTypes.string.isRequired,
-    subhead: React.PropTypes.string
+    subhead: React.PropTypes.string,
+    onClick: React.PropTypes.func
   },
   render: function() {
     return (
       <div className="icon-button">
-        <LinkSwap linkTo={this.props.linkTo} href={this.props.href}>
+        <LinkSwap linkTo={this.props.linkTo} href={this.props.href} onClick={this.props.onClick}>
           <figure>
             <ImageTag className="image" src1x={this.props.imgSrc} src2x={this.props.imgSrc2x}
             alt="" height={50} />
--- a/components/modal.jsx
+++ b/components/modal.jsx
@@ -5,7 +5,9 @@ var ModalManagerMixin = require('../mixins/modal-manager');
 var Modal = React.createClass({
   mixins: [React.addons.PureRenderMixin, ModalManagerMixin],
   propTypes: {
-    modalTitle: React.PropTypes.string
+    modalTitle: React.PropTypes.string,
+    foldedStyle: React.PropTypes.bool,
+    className: React.PropTypes.string
   },
   componentDidMount: function() {
     document.addEventListener('keydown', this.handleKeyDown);
@@ -24,12 +26,16 @@ var Modal = React.createClass({
     }
   },
   render: function() {
+    var classes = this.props.foldedStyle ? "modal-dialog folded" : "modal-dialog";
+    if (this.props.className) {
+      classes += " " + this.props.className;
+    }
     return (
       <div className="modal show"
        role="dialog"
        aria-labelledby="modal-label"
        onClick={this.handleOutsideOfModalClick}>
-        <div className="modal-dialog">
+        <div className={classes}>
           <div className="modal-header">
             <button type="button" className="close"
              onClick={this.hideModal}>&times;</button>
--- a/components/modal.jsx
+++ b/components/modal.jsx
@@ -5,9 +5,7 @@ var ModalManagerMixin = require('../mixins/modal-manager');
 var Modal = React.createClass({
   mixins: [React.addons.PureRenderMixin, ModalManagerMixin],
   propTypes: {
-    modalTitle: React.PropTypes.string,
-    foldedStyle: React.PropTypes.bool,
-    className: React.PropTypes.string
+    modalTitle: React.PropTypes.string
   },
   componentDidMount: function() {
     document.addEventListener('keydown', this.handleKeyDown);
@@ -26,16 +24,12 @@ var Modal = React.createClass({
     }
   },
   render: function() {
-    var classes = this.props.foldedStyle ? "modal-dialog folded" : "modal-dialog";
-    if (this.props.className) {
-      classes += " " + this.props.className;
-    }
     return (
       <div className="modal show"
        role="dialog"
        aria-labelledby="modal-label"
        onClick={this.handleOutsideOfModalClick}>
-        <div className={classes}>
+        <div className={"modal-dialog " + this.props.className}>
           <div className="modal-header">
             <button type="button" className="close"
              onClick={this.hideModal}>&times;</button>
--- a/less/components/modal.less
+++ b/less/components/modal.less
@@ -11,6 +11,10 @@
     width: 50rem;
     margin: 10rem auto;
   }
+  p {
+    font-size: inherit;
+    line-height: inherit;
+  }
   .modal-header {
     background-color: #EDEDED;
     border-bottom: .1rem solid @modal-inner-border-color;
@@ -88,6 +92,45 @@
         background: saturate(@orangeSidebarColor, 10%);
       }
     }
+  }
+}
+
+.modal-dialog.folded {
+  // folded corner effect on non-solid background
+  // based on http://stackoverflow.com/a/26241310
+  border: none;
+  position: relative;
+  overflow: hidden;
+  background: transparent;
+  .modal-header {
+    background-color: transparent;
+    border-bottom: none;
+    &:before {
+      position: absolute;
+      content: "";
+      top: 0;
+      right: 0;
+      border-top: 2rem solid transparent;
+      border-left: 2rem solid @lightGrey;
+      border-bottom: 2rem solid @lightGrey;
+      border-right: 2rem solid transparent;
+    }
+    button {
+      position: relative;
+      top: .5rem;
+    }
+  }
+  &:after {
+    content: "";
+    position: absolute;
+    width: 1000%;
+    height: 1000%;
+    top: 4rem;
+    background: @white none repeat scroll 0% 0%;
+    transform-origin: 50% 0 0;
+    transform: rotate(45deg);
+    z-index: -1;
+    right: -500%;
   }
 }
 
--- a/less/components/modal.less
+++ b/less/components/modal.less
@@ -163,6 +163,9 @@
     z-index: -1;
     right: -500%;
   }
+  .modal-body {
+    padding-bottom: 3.5rem;
+  }
 }
 
 .modal-backdrop {
--- a/less/components/modal.less
+++ b/less/components/modal.less
@@ -53,8 +53,7 @@
     input, textarea {
       width: 100%;
       padding: 1rem;
-      border: 1px solid #D2D2D2;
-      box-shadow: inset 0 1px 0 rgba(0, 0, 0, 0.1);
+      border: .1rem solid @lightGrey;
     }
     .Select.is-disabled .Select-control {
       background: rgb(235, 235, 228);
@@ -90,6 +89,38 @@
       border: 1px solid darken(@orangeSidebarColor, 15%);
       &:hover {
         background: saturate(@orangeSidebarColor, 10%);
+      }
+    }
+    .btn.btn-awsm {
+      // FIXME / TODO
+      // some (redundant) rules to override .modal-body .btn stylings
+      border-radius: @form-field-border-radius;
+      width: auto;
+      margin: 1rem auto;
+      background-color: @blueSidebarColor;
+      box-shadow: 0px 0.4rem 0px @darkBlueSidebarColor,
+                  0px 0.1rem 0px rgba(255, 255, 255, 0.5) inset;
+      color: @white;
+      // the following rules are copied directly from the .btn-awsm ruleset in btn.less,
+      // once we have come to a conclusion on the button stylings for modal (Issue #1045)
+      // we should clean this file up by separating rules into different rulesets
+      position: relative;
+      padding: 1rem 3.5rem;
+      vertical-align: baseline;
+      line-height: 1.42857;
+      font-size: 1.5rem;
+      text-transform: uppercase;
+      font-weight: 700;
+      letter-spacing: 0.01rem;
+      border: medium none transparent;
+      &:hover {
+        top: 0.2rem;
+        box-shadow: 0px 0.2rem 0px @darkBlueSidebarColor,
+                    0px 0.1rem 0px rgba(255, 255, 255, 0.5) inset;
+      }
+      &:active {
+        top: 0.4rem;
+        box-shadow: 0px 0.1rem 0.1rem rgba(0, 0, 0, 0.2) inset;
       }
     }
   }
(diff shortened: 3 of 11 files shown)


less/components/modal.less
```diff
@@ -88,6 +91,77 @@
         background: saturate(@orangeSidebarColor, 10%);
       }
     }
+    .btn.btn-awsm {
```
mmmavis: I don't really like this entire chunk of code as it's redundant and it overcomplicates our CSS button rulesets.  However, I think we should keep the mixed button styling issue separated from this PR since it involves decision from the designer as well as changes to many other existing buttons.  I have created https://github.com/mozilla/teach.webmaker.org/issues/1045 for this matter.

pages/events.jsx
```diff
@@ -98,9 +98,6 @@ var EventsPage = React.createClass({
   contextTypes: {
     router: React.PropTypes.func.isRequired
   },
-  contextTypes: {
```
mmmavis: duplicate of the above code

lib/travis.js
```diff
@@ -5,7 +5,9 @@ var TRAVIS_BRANCH_CONFIGS = {
     ORIGIN: 'https://teach.mofostaging.net',
     SHOW_DEV_RIBBON: 'on',
     MAILINGLIST_URL: 'https://sendto.mozilla.org/page/s/maker-party-signup-for-teach-site-staging-testing',
-    MAILINGLIST_PRIVACY_NAME: 'custom-3508'
+    MAILINGLIST_PRIVACY_NAME: 'custom-3508',
+    // PLEDGE_MAILINGLIST_URL: '',
+    // PLEDGE_MAILINGLIST_PRIVACY_NAME: ''
```
mmmavis: I will plug in the right values and uncomment these lines once I know which BSD form to use

lib/travis.js
```diff
@@ -6,6 +6,8 @@ var TRAVIS_BRANCH_CONFIGS = {
     SHOW_DEV_RIBBON: 'on',
     MAILINGLIST_URL: 'https://sendto.mozilla.org/page/s/maker-party-signup-for-teach-site-staging-testing',
     MAILINGLIST_PRIVACY_NAME: 'custom-3508'
+    // PLEDGE_MAILINGLIST_URL: '',
+    // PLEDGE_MAILINGLIST_PRIVACY_NAME: ''
```
mmmavis: These serve as a reminder for me.  I'll plug in the corresponding values and uncomment these lines once I know which BSD form to use.

mmmavis:
Hey @toolness ,

Except adding proper tests and hooking the form on our site with the BSD form(pending decision on which BSD form to use), this PR is pretty much done.  Free feel to start reviewing it once you have a chance :grin:

components/modal.jsx
```diff
@@ -5,7 +5,9 @@ var ModalManagerMixin = require('../mixins/modal-manager');
 var Modal = React.createClass({
   mixins: [React.addons.PureRenderMixin, ModalManagerMixin],
   propTypes: {
-    modalTitle: React.PropTypes.string
+    modalTitle: React.PropTypes.string,
+    foldedStyle: React.PropTypes.bool,
```
toolness: Should we just assume people will use `className` to pass in `"folded"` themselves? I'm a bit wary of adding a separate `propType` for each different style of modal that we might have, but I'm up for changing my mind...

components/icon-button.jsx
```diff
@@ -4,18 +4,23 @@ var Link = Router.Link;
 var ImageTag = require('./imagetag.jsx');
 
 var LinkSwap = React.createClass({
+  propTypes: {
+    linkTo: React.PropTypes.string,
+    href: React.PropTypes.string,
+    onClick: React.PropTypes.func
```
toolness: Perhaps it might be useful to just pass `{...this.props}`, or `{..._.omit(this.props, 'linkTo', 'href')}` into the ultimate `Link`/``a` we render? That way we can pass any props that are valid for `<a>` tags (I believe `<Link>` already passes through props itself to the underlying `<a>` it renders).

less/components/modal.less
toolness: Good idea, I agree!

pages/events.jsx
toolness: Oops hehe thanks for fixing that

pages/home.jsx
```diff
@@ -146,11 +149,121 @@ var BlogSection = React.createClass({
   }
 });
 
+var validateSignupForm = function(signUpFormState) {
+  var errors = [];
+  // regex copied from http://stackoverflow.com/a/46181
+  var regexEmail = /^([\w-]+(?:\.[\w-]+)*)@((?:[\w-]+\.)*\w[\w-]{0,66})\.([a-z]{2,6}(?:\.[a-z]{2})?)$/i;
```
toolness: Is this regex copied from [`pages/events.jsx`](https://github.com/mozilla/teach.webmaker.org/blob/develop/pages/events.jsx)? Maybe we should add a new `lib/util.js` that helps us keep things DRY? That way it'd be really easy to add tests for the regex too, if we want.

pages/home.jsx
```diff
@@ -146,11 +149,121 @@ var BlogSection = React.createClass({
...
+      e.preventDefault();
+      alert("PLEDGE_MAILINGLIST_URL is not defined. Simulating " +
+            "a successful pledge signup now.");
+      window.location = "?pledge=thanks";
+    }
+  },
+  renderValidationErrors: function() {
+    if (this.state.validationErrors.length) {
+      return (
+        <div className="alert alert-danger">
```
toolness: We should add a `role="alert"` like I did in #1034 so this is announced to screen readers!

pages/home.jsx
```diff
@@ -146,11 +149,121 @@ var BlogSection = React.createClass({
...
+          <p className="error-msg">Please enter an email address.</p>
```
toolness: Hmm, this string is hard-coded here and duplicated in your `validateSignupForm` function too. We should probably only have it appear once in our codebase, right?

If we don't anticipate more validation errors, we could just have `state.validationErrors` be a boolean, maybe even rename it to `state.invalidEmail` so that the code is clearer... if not we should probably do the whole `.map()` thing to show all the errors.

I guess it might actually be useful to factor out a `ValidationErrors` React component that does all this for us--that way we wouln't have to remember to add [`role="alert"`](https://github.com/mozilla/teach.webmaker.org/pull/1039/files#r32724485) in every new dialog we make, too. Doing that would affect #1034, though, so we should be smart about how we land that change, if we decide to do it.

pages/home.jsx
```diff
@@ -146,11 +149,121 @@ var BlogSection = React.createClass({
...
+        <p><strong>Pledge now, and we'll follow up with with details about how you can teach the web!</strong></p>
+        <form className="mailinglist-signup" action={process.env.PLEDGE_MAILINGLIST_URL} method="POST" onSubmit={this.handleSubmit}>
+          <fieldset>
+            <label htmlFor="pledge-email" className="sr-only">email</label>
+            <div className="icon-field-container">
+              <i className="fa fa-envelope"></i>
+              <input id="pledge-email" name="email" type="email" size="30" placeholder="email@example.com" valueLink={this.linkState("email")} required />
+            </div>
+            <label htmlFor="pledge-pp-note" className="sr-only">I'm okay with you handling this info as you explain in your <a href="https://www.mozilla.org/en-US/privacy/websites/">privacy policy</a></label>
+            <input id="pledge-pp-note" name={process.env.PLEDGE_MAILINGLIST_PRIVACY_NAME} type="checkbox" checked readOnly required hidden />
```
toolness: Hmm, because of the `hidden` attribute on this checkbox, I'm not sure if screen readers will actually be able to announce its label to users. I wonder if it might be better to give it a `sr-only` class instead of `hidden`?

Also, might want to put the label text in a `<span>` variable to avoid reduplication but it's up to you. Another alternative might be to remove the `<label>`, give the `.pp-note` below an `id`, and map that to the [`aria-labelledby` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) of the checkbox, though it might not work on really old screen readers...

pages/home.jsx
```diff
@@ -146,11 +149,121 @@ var BlogSection = React.createClass({
...
+});
+
+var ThankYouModal = React.createClass({
+  render: function() {
+    return (
+      <Modal modalTitle="Thanks for your pledge!" className="modal-pledge">
+        <p>We appreciate your commitment to keeping the web open, accessible and ours.</p>
+        <p><strong>Share and tell your friends</strong></p>
+        <div className="social-share">
+          <a href="https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fteach.mozilla.org" className="facebook">
```
toolness: To be more robust and DRY-friendly, I'd use `encodeURIComponent(require('../lib/config').ORIGIN)` here and in the twitter link below instead of hard-coding `https%3A%2F%2Fteach.mozilla.org`, but I understand that this code is likely fairly temporary for the duration of our pledgey thing, so it's up to you.

toolness:
This is looking good so far!

Another alternative to dealing with the accessibility issues is just to land this and file a separate issue for improving the accessibility of it. That way we can land #1034 separately and then land a separate PR that refactors some of the common code between all our modals, which should improve the accessibility of these pledge modals.

pages/home.jsx
mmmavis: yup yup was thinking about that but just wasn't sure where that regex should go :smile:  `pages/events.jsx` it is!

components/icon-button.jsx
mmmavis: @toolness I'm not sure if `<Link>` takes any valid `<a>` attributes.  It seems like it doesn't... ? Mind double checking for me?

- https://github.com/rackt/react-router/blob/master/modules/Link.js#L77-L80
- https://github.com/rackt/react-router/blob/57543eb41ce45b994a29792d77c86cc10b51eac9/build/lib/createRouter.js#L188-L214

components/modal.jsx
mmmavis: Awesome!  I tried to do the same for another ticket but it didn't work(not sure why :eyes:). That's why I didn't bother using the approach you suggested this time.  Anyways, thanks for flagging this - all good now!  I'll update the code.

pages/home.jsx
mmmavis: Right! And I need to check out that PR!

pages/home.jsx
mmmavis: Personally I would downvote this "fake" checkbox treatment.  It feels unconventional and complicates things for screen readers.  If I recall correctly... we adapted this style from https://id.webmaker.org/signup.  Not sure if the design was based on researches/analytics(fewer fields higher conversion rate?) or it was for prettifying UI's sake.  If it was the latter maybe we could drop this for the standard, non-hidden privacy policy checkbox?

pages/home.jsx
mmmavis: > If we don't anticipate more validation errors, we could just have `state.validationErrors` be a boolean, maybe even rename it to `state.invalidEmail` so that the code is clearer... if not we should probably do the whole `.map()` thing to show all the errors.

I'm hoping to turn that fake privacy policy checkbox into a standard one.  See [my comment](https://github.com/mozilla/teach.webmaker.org/pull/1039/files#r32792452).  After we have decided what to do with it, I'll clean up this chunk of code.

And yea, I thought about refactoring the code when I copied the form validation code from `events.jsx`.  But I *felt* this pledge CTA is more or less temporary that's why I got lazy about the factoring work... :smiling_imp: :grimacing:   If we do decide to factor out a `ValidationErrors` React component, do we want to land the changes in this PR or a separate one?

pages/home.jsx
mmmavis: I like the idea of using `encodeURIComponent`! :+1: However I think it makes sense to hard code the URL "https://teach.mozilla.org" here instead of using `require('../lib/config').ORIGIN` as it might be confusing to people who are QAing on the staging site.

How about passing this `twitterShare` var  to `href` to make the code more readable?
```javascript
var twitterShare = "https://twitter.com/home?status=" +
                   encodeURIComponent("I just pledged to #TeachTheWeb! https://teach.mozilla.org");
```

components/icon-button.jsx
toolness: So your first link is to this code:

```js
    var props = Object.assign({}, this.props, {
      href: router.makeHref(to, query),
      onClick: this.handleClick
    });
```

The `this.props` in the first line basically means it's copying all the properties passed to it, but then overriding the `href` and `onClick` properties with its own values.

pages/home.jsx
toolness: Ah, ok, if we're adapting this from the webmaker signup then I guess we should just assume they've done usability testing on it and keep it as-is... I agree that it's kinda weird but we can revisit it later!

pages/home.jsx
toolness: Yep no worries, I get lazy about temporary-ish code too...

As for `ValidationErrors`, good question--I think it impacts the unit tests mostly... Maybe land this but file two issues, one for factoring out `ValidationErrors`, and another for adding unit tests to the code introduced in this PR?

components/icon-button.jsx
mmmavis: yea and I think this means we can't just pass a URL as `href` to `<Link>`? :confused: dislike

mmmavis: ohhhhhh I think we can utilize [`{...other}`](https://facebook.github.io/react/docs/transferring-props.html) :smile_cat:

pages/home.jsx
mmmavis: Cool, I'm tempted to factor out a `<EmailSignupForm>` component actually... but since my work for #900 is in progress already and it involves copying a signup form to hero unit section, I would wanna wait until both 2 PRs are landed first to save me from fixing merge conflicts.  Ticket created: https://github.com/mozilla/teach.webmaker.org/issues/1050


Unit tests ticket created: https://github.com/mozilla/teach.webmaker.org/issues/1049

mmmavis:
@toolness I think this is ready for review again!

Note that I created https://github.com/mozilla/teach.webmaker.org/issues/1049 and https://github.com/mozilla/teach.webmaker.org/issues/1050 for work left in this PR.

Once this PR gets landed on staging, we can then move onto the last step - https://github.com/mozilla/teach.webmaker.org/issues/1016 !  :turtle: :smile:

toolness:
Isn't the title "pledge to teach" supposed to appear at the top of the modal?

This is what it looks like now:



But this is @sabrinang's mockup from https://github.com/mozilla/teach.webmaker.org/issues/908#issuecomment-110122037:



Hmm, also, the submit button in sabrina's mockup says "pledge to teach", while yours says "submit email". I think I like sabrina's better because it feels more like a pledge and less like signing up for a mailing list. :stuck_out_tongue: 

I also wonder if it makes sense to push the top image up a bit using a negative top margin or somesuch, so there's less whitespace at the top of the modal and it looks a bit more like the mock-up... That's up to you though. I know it's a bit tricky because of the DOM structure of that modal, so it's ok to just skip it IMO.

lib/util.js
```diff
@@ -0,0 +1,4 @@
+exports.REGEX = {
+  // copied from http://stackoverflow.com/a/46181
+  email: /^([\w-]+(?:\.[\w-]+)*)@((?:[\w-]+\.)*\w[\w-]{0,66})\.([a-z]{2,6}(?:\.[a-z]{2})?)$/i
```
toolness: Hmm, instead of exporting the email regex, maybe it's better to just export a function called `isValidEmail()`? That seems more readable and more extensible, since it allows us to add more non-regex-based code to the function in the future if we want. It's up to you though, this is ok as-is.

pages/events.jsx
```diff
@@ -64,7 +64,7 @@ var FormMailingListSignup = React.createClass({
           </fieldset>
           <fieldset>
             <label htmlFor="mailinglist-pp-note" className="sr-only">I'm okay with you handling this info as you explain in your <a href="https://www.mozilla.org/en-US/privacy/websites/">privacy policy</a></label>
-            <input id="mailinglist-pp-note" name={process.env.MAILINGLIST_PRIVACY_NAME} type="checkbox" checked readOnly required hidden />
+            <input id="mailinglist-pp-note" name={process.env.MAILINGLIST_PRIVACY_NAME} type="checkbox" className="sr-only" checked readOnly required />
```
toolness: Nice, I like this solution, it actually shows up in NVDA (my screen reader) and is announced as read-only and required. Much more obvious than the current sign-up page on id.webmaker.org, where it's basically just treated as loose text by NVDA and could easily go unnoticed.

pages/home.jsx
```diff
@@ -146,11 +150,123 @@ var BlogSection = React.createClass({
...
+                      encodeURIComponent("https://teach.mozilla.org");
+  var twitterShare = "https://twitter.com/home?status=" +
+                     encodeURIComponent("I just pledged to #TeachTheWeb! https://teach.mozilla.org");
+    return (
+      <Modal modalTitle="Thanks for your pledge!" className="modal-pledge">
+        <p>We appreciate your commitment to keeping the web open, accessible and ours.</p>
+        <p><strong>Share and tell your friends</strong></p>
+        <div className="social-share">
+          <a href={facebookShare} className="facebook">
+            <i className="fa fa-facebook"></i>
```
toolness: I noticed the `:focus`'d but not `:hover`'d state of this link (and the twitter link too) is underlined:



That looks a little bit funky to me, and I think the standard user agent `outline` CSS for the focused element is enough, but it's up to you really.

toolness:
Purely code-wise, this looks good!

I think that @sabrinang should design review the modal though, at least the comments I made at https://github.com/mozilla/teach.webmaker.org/pull/1039#issuecomment-113487323. Once she gives the :+1: I think this is good to land.

mmmavis:
Agree that the modal needs some tweaks.  Being looking at it for too long so I missed the details :grimacing:

sabrinang:
Looking good guys!

For this pledge modal, If possible I'd prefer if the image was above the title so it isn't sandwiched between the text so that it visually gives the illustration more hierarchy and makes it look more distinct than our regular modals.  Also +1 to "Pledge Now" button label.

Looking at the screenshot it feels that the "I pledge to:" & "Pledge now, and we'll follow up..." text is a bit too bold that my eyes skip through the points to look at the "Pledge now, and we'll follow up..." line cause of the weight contrast - I'd suggest try using try semi bold?

mmmavis:
screen shot 2015-06-19 at 11 14 46 am

@sabrinang how's this?

sabrinang:
Only thing is I'd just add a bit more margin beneath the "Pledge Now" button so the white space balances out with the top a bit better and doesn't feel as tight.

lib/util.js
mmmavis: oh yea I prefer your solution too!

mmmavis:
@toolness I fixed more things as what you and @sabrinang suggested.  Gonna merge this PR once I see the OK from Travis.  Thanksssss!

PR #1051, opened 2015-06-19
set og:image metadata and spider it.

### LINKED ISSUE #1047: Set a constant og:image meta tag for all pages
As mentioned in https://github.com/mozilla/teach.webmaker.org/issues/798#issuecomment-113153133, the image Facebook uses when people share a Teach page is a bit unpredictable, particularly since our hero unit images are in the CSS of the page, which Facebook's scraper doesn't look at.

Ideally we should include the page-specific hero unit image, if any, as the `og:image` metadata, but for now we can just hard-code the homepage's hero unit image for all pages.

toolness:
fixes #1047.

this sets the `og:image` for all pages to the hero unit image for the homepage, so sharing on facebook now looks like this:

note that if you want to test this locally, you should use ngrok.com *and* set your `ORIGIN` to whatever your ngrok origin is. otherwise the [absolute URL in the `og:image`](https://github.com/mozilla/teach.webmaker.org/pull/1051#discussion_r32827149) won't be findable by facebook's [open graph debugger](https://developers.facebook.com/tools/debug/).

i also noticed that the open graph debugger is *extremely* flaky right now; i had to re-scrape my pages multiple times for it to succeed, for some reason.

### DIFF
--- a/lib/index-static.jsx
+++ b/lib/index-static.jsx
@@ -1,6 +1,7 @@
 var _  = require('underscore');
 var React = require('react');
 
+var config = require('./config');
 var routes = require('./routes.jsx');
 
 // This isn't actually called in node, it's stringified and plopped in
@@ -35,6 +36,7 @@ function generateWithPageHTML(url, options, pageHTML) {
         <meta charSet="utf-8"/>
         <meta name="viewport" content="width=device-width, initial-scale=1"/>
         <meta property="og:description" content="We want more people to see themselves as citizens of the web. The Mozilla Learning Network offers programs and a global community dedicated to helping people learn the most important skills of our age: the ability to read, write and participate in the digital world."/>
+        <meta property="og:image" content={config.ORIGIN + "/img/pages/home/hero-unit.jpg"}/>
         {Object.keys(options.meta).map(function(name, i) {
           return <meta key={i} name={name} content={options.meta[name]}/>;
         })}
--- a/test/browser/spider.js
+++ b/test/browser/spider.js
@@ -3,12 +3,32 @@ var urlResolve = require('url').resolve;
 var chalk = require('chalk');
 var Crawler = require('simplecrawler');
 
+var config = require('../../lib/config');
 var server = require('./server').create();
+
+var URL_FINDERS = [
+  findSrcSetURLs,
+  findOpenGraphURLs
+];
 
 function findSrcSetURLs(resourceText, baseURL) {
   var urls = [];
 
   resourceText.replace(/\ssrcset\s?=\s?['"]([^"']+)/ig, function() {
+    arguments[1].split(',').forEach(function(resource) {
+      var url = resource.trim().split(' ')[0];
+
+      urls.push(urlResolve(baseURL || '', url));
+    });
+  });
+
+  return urls;
+}
+
+function findOpenGraphURLs(resourceText, baseURL) {
+  var urls = [];
+
+  resourceText.replace(/\sproperty="og:image"\s?content\s?=\s?['"]([^"']+)/ig, function() {
     arguments[1].split(',').forEach(function(resource) {
       var url = resource.trim().split(' ')[0];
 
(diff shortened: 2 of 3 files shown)


lib/index-static.jsx
```diff
@@ -35,6 +36,7 @@ function generateWithPageHTML(url, options, pageHTML) {
         <meta charSet="utf-8"/>
         <meta name="viewport" content="width=device-width, initial-scale=1"/>
         <meta property="og:description" content="We want more people to see themselves as citizens of the web. The Mozilla Learning Network offers programs and a global community dedicated to helping people learn the most important skills of our age: the ability to read, write and participate in the digital world."/>
+        <meta property="og:image" content={config.ORIGIN + "/img/pages/home/hero-unit.jpg"}/>
```
toolness: This is really annoying, facebook doesn't rebase the URL here based on the current origin, it just expects absolute URLs, so we have to absolute-ify it ourselves. Unfortunately our link-checker wasn't built for this so I had to modify it to deal with this crap.

test/browser/spider.js
```diff
@@ -27,13 +47,26 @@ function crawl() {
...
+    });
 
-    return resources;
+    return resources.map(function(url) {
+      // It's possible that some of our URLs might be absolute URLs
+      // based on config.ORIGIN. However, because the server we've
+      // started for crawling is based at a dynamic origin on localhost,
+      // we need to rebase such URLs to be at our dynamic origin, so
+      // that they're actually spidered.
+      if (url.indexOf(config.ORIGIN) === 0) {
```
toolness: see rationale for this [above](https://github.com/mozilla/teach.webmaker.org/pull/1051/files#r32827149).

test/browser/spider.js
```diff
@@ -19,6 +25,20 @@ function findSrcSetURLs(resourceText, baseURL) {
   return urls;
 }
 
+function findOpenGraphURLs(resourceText, baseURL) {
+  var urls = [];
+
+  resourceText.replace(/\sproperty="og:image"\s?content\s?=\s?['"]([^"']+)/ig, function() {
+    arguments[1].split(',').forEach(function(resource) {
```
mmmavis: Is there a way to specify each argument(by giving it a more descriptive name) in the function's signature so we can grab the argument we want directly instead of using `arguments[1]`?  I don't use the `replace` method that often and after reading [MDN doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_function_as_a_parameter) it's still not immediately clear to me what `arguments[1]` refers to.  :sob:

mmmavis:
Awesome work @toolness :+1:  I tested it on Facebook's debugger and verified it's working as expected.  Feel free to decide what to do with my [only comment](https://github.com/mozilla/teach.webmaker.org/pull/1051#discussion_r32862725) :grin: I think once I'm more used to the `.replace` method, `arguments` will look more friendly to me!

test/browser/spider.js
toolness: Oh, sorry, all I wanted to do was find all the occurrences of a RegExp in a string and get an array of all their [capturing groups](http://www.regular-expressions.info/brackets.html)--so that I could find what the value of the `content` attribute is--but there appears to be no way to easily do that aside from calling `replace` and passing it a function. The first argument passed to the function is the whole string, i think, while the following arguments are the capturing groups, or something. It's all ridiculously complicated and not well-documented. :sob:

toolness:
Ok, I actually did a fairly significant refactoring to make that `replace` nonsense a bit easier to understand. Can you take a look?

mmmavis:
will get to this PR today

test/browser/spider.js
```diff
@@ -3,14 +3,38 @@ var urlResolve = require('url').resolve;
...
 
+var URL_FINDERS = [
+  findSrcSetURLs,
+  findOpenGraphURLs
+];
+
+// Return an array of all the first submatches (the parts of a
+// regular expression in parentheses) of a given regular expression
+// in a body of text. The regular expression is assumed to have the
+// global match flag.
```
mmmavis: Awesome! Thanks for putting this comment.

test/browser/spider.js
```diff
@@ -27,13 +60,26 @@ function crawl() {
...
 
-    resources.push.apply(resources, findSrcSetURLs(
-      resourceData.toString("utf8"),
-      queueItem.url
-    ));
+    URL_FINDERS.forEach(function(findURLs) {
+      resources.push.apply(resources, findURLs(
+        unicodeResourceData,
+        queueItem.url
+      ));
```
mmmavis: @toolness do we need to call `apply` here? or the following would work as well

```javascript
resources.push(resources, findURLs(
  unicodeResourceData,
  queueItem.url
));
```

mmmavis: Ah nvm, I get it now :grin:  
We are calling `apply` so we can use `push` in a way similar to how we use `concat`, since we can just pass an array as an argument instead of looping through that array and call `resources.push` multiple times

mmmavis:
R+ :+1: :dancer:

PR #1048, opened 2015-06-18
Add Maker Party 2015 page

mmmavis:
This fixes #1000.

PR #1060, opened 2015-06-20
Refresh Events page with pre/during Maker Party design

### LINKED ISSUE #900: Implement "pre/during MP" state for /events

mmmavis:
This fixes #900


less/components/page-end-cta.less
```diff
@@ -3,26 +3,4 @@
...
-    input[type="email"] {
-      width: 100%;
-    }
-    input[type="checkbox"] + div {
-      text-align: left;
-    }
-    input[type="submit"] {
-      margin-top: 0;
-    }
-  }
```
mmmavis: I realized the signup form on Events page is the **only** form on the site that's embedded within `.page-end-cta`.  So I think it makes sense to move these rules to `less/pages/events.less`.  That way if we ever have to have another form in `<PageEndCTA>`, we won't have to override stylings.

less/pages/events.less
```diff
@@ -13,6 +20,40 @@
...
+  input {
+    padding: 1rem 1.5rem;
+    border: .1rem solid @grey;
+  }
+  input[type="email"] {
+    width: 100%;
+  }
+  input[type="submit"] {
+    margin-top: 0;
+  }
```
mmmavis: we should find time to unifying our form stylings across the site though... :eyes:

pages/clubs-toolkit.jsx
```diff
@@ -2,7 +2,6 @@ var React = require('react');
 var Router = require('react-router');
 var Link = Router.Link;
 var Illustration = require('../components/illustration.jsx');
-var PageEndCTA = require('../components/page-end-cta.jsx');
```
mmmavis: not used

pages/clubs.jsx
```diff
@@ -10,7 +10,6 @@ var Map = require('../components/map.jsx');
 var Blockquote = require('../components/blockquote.jsx');
 var IconLinks = require('../components/icon-links.jsx');
 var IconLink = require('../components/icon-link.jsx');
-var PageEndCTA = require('../components/page-end-cta.jsx');
```
mmmavis: not used

pages/events.jsx
```diff
@@ -8,12 +8,11 @@ var IconLinks = require('../components/icon-links.jsx');
 var IconLink = require('../components/icon-link.jsx');
 var Illustration = require('../components/illustration.jsx');
 var PageEndCTA = require('../components/page-end-cta.jsx');
+var util = require('../lib/util');
```
mmmavis: Wondering if we should give this var a different name?  We've been assigning `require('util')` to `util`.
```javascript
var util = require('util');
```
found 3 occurrences

components/illustration.jsx
```diff
@@ -3,34 +3,29 @@ var ImageTag = require('./imagetag.jsx');
...
-            </a> :
-            image
-          }
-        </div>
-        <div className="content-container">{this.props.children}</div>
+        {imageContainer}
+        {contentContainer}
       </div>
     );
   }
```
mmmavis: I cleaned up this file a bit so we can now pass `classesName` directly to the wrapper of this `<Illustration>` component.

components/illustration.jsx
```diff
@@ -3,34 +3,29 @@ var ImageTag = require('./imagetag.jsx');
...
     src2x: React.PropTypes.string,
     alt: React.PropTypes.string.isRequired,
     link: React.PropTypes.string
   },
   render: function() {
-    var classes = this.props.verticalLayout ? "illustration vertical-layout" : "illustration";
+    var classes = this.props.className ? ("illustration " + this.props.className) : "illustration";
     var image = <ImageTag width={this.props.width}
                           height={this.props.height}
-                          className={this.props.className}
```
mmmavis: We shouldn't be doing this.  `className` should go to the component's wrapper instead.

less/components/illustration.less
```diff
@@ -1,15 +1,28 @@
 .illustration {
+  overflow: auto;
+  padding: 0 0 2rem 0;
   &.vertical-layout {
     flex-direction: column;
     .image-container {
       padding-right: 0;
     }
   }
-  .img-circle {
```
mmmavis: `.img-circle` is a class defined in [Bootstrap](https://github.com/twbs/bootstrap/blob/3487898a6af8c1d5b12f7b040d30e71b19e34923/less/scaffolding.less#L106).  Since we have changed how classes get consume in the `<Illustration>` component I had to rename `.img-circle` into something else (ie. `.circular-image`) to avoid tons of styling override.

mmmavis:
I think this is almost done.  Gonna head out now and will review my code again later tonight.

The design ticket is kinda huge and I'm worried I might have missed some details.  If @hannahkane and @sabrinang get a chance, please help reviewing design and copy. Thanks!

events-mp

pages/events.jsx
```diff
@@ -108,70 +192,99 @@ var EventsPage = React.createClass({
...
-              </div>
-              <div className="col-xs-4 col-sm-4 col-md-4 col-lg-4">
-                <a href="https://www.flickr.com/photos/newyouthcity/14715964794/in/faves-mozilladrumbeat/"><img alt="2014 Hive NYC Summer Quest Maker Party" src="//c4.staticflickr.com/4/3905/14715964794_a9fa75002c_c.jpg" className="flickr-thumbnail" /></a>
-              </div>
+          </section>
+        </div>
+        <div className="row mp-activities-banner">
+          <section>
+            <div className="btn-container">
+              <a className="btn btn-awsm" href="">Get the 2015 Maker Party Activities</a>
```
mmmavis: waiting for https://github.com/mozilla/teach.webmaker.org/pull/1048 to land
and I'll change this to `<Link to="maker-party-2015">`

mmmavis:
This is ready for review @toolness @sabrinang 

@sabrinang I would encourage you to check out this branch to verify stylings on this page match what you expected.  Please pay attention to font sizes especially as I had to guess them based on the [flat mockup](https://cloud.githubusercontent.com/assets/2492510/8172494/ed49f02c-1375-11e5-844d-4f8f9c3fabbd.png).  

Note that styling-wise the Sample Maker Parties section doesn't look identical to the Activities section on https://teach.mozilla.org/activities/ - not sure if it was intentional.
- font size for header
- width of thumbnail
- vertical alignment for the text block (top align on mockup vs center align on Activities page)

Let me know if there's anything I should tweak!

:shipit: You can check out this branch by :shipit:

1. open up Terminal
2. `cd` to where your Teach repo locates
3. `git checkout develop`
4. `git pull origin develop`
5. `git checkout -b mmmavis-issue-900-pre/during-MP-version develop`
6. `git pull https://github.com/mmmavis/teach.webmaker.org.git issue-900-pre/during-MP-version`
7. `npm install`
8. `npm start`
9. visit `http://localhost:8008/events/`

Ping me if you have trouble running it locally.

pages/events.jsx
```diff
@@ -7,13 +7,16 @@ var ModalManagerMixin = require('../mixins/modal-manager');
 var IconLinks = require('../components/icon-links.jsx');
 var IconLink = require('../components/icon-link.jsx');
 var Illustration = require('../components/illustration.jsx');
+var ImageTag = require('../components/imagetag.jsx');
 var PageEndCTA = require('../components/page-end-cta.jsx');
+var config = require('../lib/config');
+var util = require('../lib/util');
+
+var UniqeIdMixin = require('unique-id-mixin');
```
toolness: There's a minor typo here, `UniqeIdMixin` should be `UniqueIdMixin`.

toolness:
Hmm, so I have some concerns about the [`unique-id-mixin`](https://www.npmjs.com/package/unique-id-mixin) module. After looking at its [source code](https://github.com/DelvarWorld/Unique-Id-Mixin/blob/master/index.js) it appears that it's storing module-level state (an `id` variable) that gets incremented every time a new ID is requested. Because this can't be configured on a per-page basis, the `id` of the pre-rendered page (generated by our static site build process or our dynamic lightweight server) may be different from the `id` of the user's browser, which will result in the following kinds of errors being logged:

```
"Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) nglist-signup-email-1" class="sr-only" d
 (server) nglist-signup-email-5" class="sr-only" d"
```

That's literally what's logged when I run `npm start` and visit `/events/`. If you run `npm run app` (the dynamic lightweight server) instead, you'll see that the `id`s the server is generating increase by 4 every time you reload the `/events/` page.

So rather than using the `unique-id-mixin`, could you just pass an `idPrefix` prop into the `FormMailingListSignup` class the same way that our `ModalAddOrChangeYourClub` in `clubs.jsx` does? And then just pass unique `idPrefix` props to the component when you use it in `EventsPage.render()`, e.g. `idPrefix="hero_unit_"` for the top one and `idPrefix="page_end_cta_"` for the bottom one? I know it's a bit more of a hassle, but it'll result in consistent markup between server and client, which helps speed up page loads.

toolness:
The hyperlinked "privacy policy" in the hero unit has a really bad contrast ratio because of the photographic background:



I think one alternative is to style the hyperlink so it's white and underlined, but I am also a little worried in general about the smallness of that type against such a varied background. I wonder if adding a text shadow might help increase the contrast more?

less/components/illustration.less
toolness: O blug. Maybe we should call it `illustration-img-circle` instead of `circular-image` so that it doesn't become too confusing? And mention in a comment above the rule definition that we're calling it that so we don't collide w/ Bootstrap's namespace?

sabrinang:
+1 to @toolness suggestion to white and underline that hyperlink since it is hard to predict what will photo would be behind this text.

Some other adjustments to make:
screen shot 2015-06-30 at 1 15 22 pm
- Join the Global Movement text section and Maker Party logo looking weird on mobile - Have the text take up the whole column and have the image go below it

- To differentiate these image thumbnails from the clickable activity thumbnails we can mask the images as circles like this...
screen shot 2015-06-30 at 1 14 57 pm
and shrink them a bit and align text from the top of the circle to balance out the white space.

mmmavis:
> Because this can't be configured on a per-page basis, the id of the pre-rendered page (generated by our static site build process or our dynamic lightweight server) may be different from the id of the user's browser, which will result in the following kinds of errors being logged:

@toolness would you mind explaining more about why the ids would differ?

mmmavis:
https://github.com/mmmavis/teach.webmaker.org/commit/2c1cfe75059fc1691aa06760ac879a2855cc264a has the tweaks based on @sabrinang 's https://github.com/mozilla/teach.webmaker.org/pull/1060#issuecomment-117376726 (one exception though, Sabrina and I agree that the text block of the sample Maker Party section should stay vertically center aligned)

toolness:
> @toolness would you mind explaining more about why the ids would differ?

One way of thinking about it is that when you have a module-level variable like `unique-id-mixin`'s [`id`](https://github.com/DelvarWorld/Unique-Id-Mixin/blob/master/index.js#L1) that is incremented every time a function in the module is called, its value is necessarily tied to the **lifetime** of the module. On the server-side, that is a very long lifetime--the module will come into existence when the server starts, and every time the server needs to generate the Maker Party page for *any* user, it will call code in that module which increments its `id`. So if the server has been around a long time and served the Maker Party page to 30 different users, `id` will have a high value.

On the other hand, when the website is loaded in the user's browser, the `unique-id-mixin` module exists there too, but it's lifetime is much shorter. When the user first loads the Maker Party page, `id` will be at 1, so when the page first loads and React generates its initial markup, it will find discrepancies between it and the markup the server generated because the `id`s of the two instances of the `unique-id-mixin` module (one on the server, one on the client) were totally different and led to different attributes in the HTML.

Does that help?

mmmavis:
@toolness (sorry I still need to read your comment above but) did you get a chance to review my updated PR?  Trying to see if we can push this out today :smiley:

mmmavis:
Hey @toolness ,

@hannahkane and I discussed and we thought it would be better to push a new release today instead of Friday or next week(I'll be on PTO Mon and Tue) since Maker Party is getting really close.  I'm gonna merge this PR first and hopefully you can still review my code changes when you are back.  :cold_sweat:  :cold_sweat:  :cold_sweat:

PR #1072, opened 2015-07-04
Add /healthcheck page

### LINKED ISSUE #884: Have a way to tell what version of site we are running
I'm not sure if there's an existing method already but I can't tell what version of the Teach site we are running on PROD (I guess this could apply on STAGING as well?).

Maybe something like https://webmaker.org/healthcheck where we can simply return the version number specified in `package.json` would help?

Updates:
Oh I forgot we have the "dev ribbon" on staging, maybe we can incorporate the `commit` link in `healthcheck` for PROD as well?

mmmavis:
This fixes #884 and is basically a simplified version of the dev ribbon modal.

### Option 1 ###
site-health-check

### Option 2 ###
site-health-check-2

@toolness any feedback or suggestion? I prefer option #2 since it's way more obvious for visitors to distinguish this page from the rest of the site(the real content).

### DIFF
--- a/less/pages/healthcheck.less
+++ b/less/pages/healthcheck.less
@@ -7,10 +7,22 @@
     width: 100%;
   }
   .site-meta {
+    font-size: 1.8rem;
     text-align: center;
     margin: 5rem auto;
     @media screen and (min-width: @screen-sm-min) {
       max-width: 50rem;
+    }
+  }
+  .go-back-home {
+    margin-top: 4rem;
+    a {
+      &:hover {
+        text-decoration: none;
+        i.fa {
+          transform: scale(1.2);
+        }
+      }
     }
   }
 }
--- a/less/pages/healthcheck.less
+++ b/less/pages/healthcheck.less
@@ -1,0 +1,9 @@
+.page.healthcheck {
+  .site-meta {
+    text-align: center;
+    margin: 5rem auto;
+    @media screen and (min-width: @screen-sm-min) {
+      max-width: 50rem;
+    }
+  }
+}
--- a/less/pages/healthcheck.less
+++ b/less/pages/healthcheck.less
@@ -1,4 +1,11 @@
 .page.healthcheck {
+  .sidebar,
+  footer {
+    display: none !important;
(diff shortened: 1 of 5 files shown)


less/pages/healthcheck.less
```diff
@@ -0,0 +1,16 @@
+.page.healthcheck {
+  .sidebar,
+  footer {
+    display: none !important;
+  }
+  main {
+    width: 100%;
+  }
```
mmmavis:
```css
.sidebar,
  footer {
    display: none !important;
  }
  main {
    width: 100%;
  }
```
is just a lazy way to implement [option 2](https://github.com/mozilla/teach.webmaker.org/pull/1072#issue-92952486) design.  :eyes:

toolness:
Overall this looks good and I agree that option 2 is ideal too.

For `healthcheck.jsx` do you think you could have it reuse code from the dev mode ribbon instead of duplicating it? Since this is still marked as WIP I assume you're already going to do this, but I figured I'd mention it just in case.

Or perhaps a better way might be to move all the shared logic into `healthcheck.jsx`, actually, since we don't want the dev-mode ribbon to be included in the JS bundle when `process.env.NODE_ENV === 'production'`, but we *do* want the health check to always be included in the build.

mmmavis:
> For `healthcheck.jsx` do you think you could have it reuse code from the dev mode ribbon instead of duplicating it? Since this is still marked as WIP I assume you're already going to do this, but I figured I'd mention it just in case.

> Or perhaps a better way might be to move all the shared logic into `healthcheck.jsx`, actually, since we don't want the dev-mode ribbon to be included in the JS bundle when `process.env.NODE_ENV === 'production'`, but we *do* want the health check to always be included in the build.

Yup yup for sure and that was my plan too.  Thanks for mentioning it anyway, I sometimes do forget to factor things out.  :smile_cat:

mmmavis:
@toolness PR updated again, R? :smile:

toolness:
Looks great to me, thanks @mmmavis!

PR #1089, opened 2015-07-13
A few grammar and CSS tweaks to /healthcheck

toolness:
Just a few minor changes:

* JSX can be a bit annoying with newlines because it doesn't currently [treat newlines as whitespace text](https://github.com/facebook/jsx/issues/19) like HTML does. Because it also [trims the trailing whitespace at the end of lines](https://facebook.github.io/react/blog/2014/02/20/react-v0.9.html#jsx-whitespace) we need to wrap any such whitespace in `{' '}`, which is weird, but whatever. I did this so that there's now a space between "commit" and the commit hash, and between "version" and the version number.
* I added a CSS transition to the home icon, neato.
* The word "based" was repeated twice in a row.
* The sentence didn't have a period at the end of it.

### DIFF
--- a/less/pages/healthcheck.less
+++ b/less/pages/healthcheck.less
@@ -16,6 +16,9 @@
   }
   .go-back-home {
     margin-top: 4rem;
+    a i.fa {
+      transition: transform 0.25s;
+    }
     a {
       &:hover {
         text-decoration: none;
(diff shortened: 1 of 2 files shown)


mmmavis:
Neat!  Thanks for the patch @toolness !  I should've proofread the stuff one last time before I committed last commit.

Issue #1083, opened 2015-07-08
Visual conventions across Teaching Activities and Maker Party case studies is inconsistent


toolness:
When I saw the cool new Maker Party case studies, I tried clicking on the title/image to see where it took me, and was confused for a bit because they weren't hyperlinked:

![2015-07-08_19-01-29](https://cloud.githubusercontent.com/assets/124687/8584400/1f201104-25a4-11e5-9d4b-6ac7732c3646.jpg)

Upon further reflection I realized that I had *assumed* they were hyperlinked because we use very similar visual conventions on the Teaching Activities page:

![2015-07-08_19-01-58](https://cloud.githubusercontent.com/assets/124687/8584404/35b36ac4-25a4-11e5-9b24-2c8c992224e0.jpg)

*Those* titles/images are hyperlinked. It seems like we might want to be more consistent with our visual conventions, maybe? I'm not sure about this one--I've actually done similar kinds of things on my own projects, and have similarly avoided underlining hyperlinked headings because it just looks way too gaudy... But on the other hand, when the only indication you have of whether something is hyperlinked is by hovering over it or tabbing to it--gestures that don't even exist on touch devices--that could be bad. Maybe we could put some sort of icon next to each link instead of underlining it? I dunno.

mmmavis:
I agree with you @toolness. I have a tendency to interpret text in body colour(black, grey) as non-hyperlinks as well - I never tried to hover over them. :eyes:  Maybe adding icon or changing the font colour would help?

I was thinking if we could add some flavours to the hyperlinked image... maybe some sort of tiny animation?  This is not touch devices friendly but I guess users can always check out the hyperlinked headers.

Btw, I just noticed there's inconsistency in letter casing in headers.

mmmavis:
@sabrinang this is not urgent, just pinging you in case this gets buried in your inbox :wink: 
Would love your thoughts on ^^

PR #1092, opened 2015-07-13
Italicize emphasized words instead of making them all caps

### LINKED ISSUE #1084: Use of all-caps to emphasize text is weird
On the Events page, the mini case study for the "Make and Remake Hackathon" says:

> At the Make and Remake Hackathon Day attendees spent the first half of the day using Webmaker tools to MAKE something awesome. In the second half of the Hackathon individuals then turned their completed makes over to a friend so they could then REMAKE their make to create something new.

The use of all-caps for MAKE and REMAKE reads really weird to me, perhaps because we don't use such a typographic convention anywhere else on the site. I think it'd look more consistent (and probably be more accessible) if these words were italicized via `<em>` instead, but it's up to you... Feel free to just close this issue if you disagree.

mmmavis:
This fixes #1084

### DIFF
--- a/pages/events.jsx
+++ b/pages/events.jsx
@@ -121,7 +121,7 @@ var MakerPartyExample = React.createClass({
               </div> : null
             }
             <div>
-              <p className="description" dangerouslySetInnerHTML={{__html: this.props.description}}></p>
+              <p className="description" dangerouslySetInnerHTML={{__html: this.props.descriptionHTML}}></p>
             </div>
           </div>
         </Illustration>
@@ -141,7 +141,7 @@ var MakerPartyExamples = React.createClass({
       },
       location: "A private home in Barcelona, Spain",
       participants: "David and his family",
-      description: "Participants learned about the importance of Net Neutrality, and considered how to take action. They used Thimble to create Net Neutrality-themed memes.",
+      descriptionHTML: "Participants learned about the importance of Net Neutrality, and considered how to take action. They used Thimble to create Net Neutrality-themed memes.",
       src1x: "/img/pages/events/nn-maker-party.png",
       src2x: "/img/pages/events/nn-maker-party@2x.png",
     },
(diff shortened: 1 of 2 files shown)


pages/events.jsx
```diff
@@ -163,7 +165,7 @@ var MakerPartyExamples = React.createClass({
         link: "https://www.mozilla.org/contact/communities/indonesia/"
       },
       location: "Park in Jakarta, Indonesia",
-      description: " At this Lo-Fi Maker Party, participants used paper, Post-it's and tennis balls to learn basic HTML and website structure, simple programming commands and app design.",
+      description: "At this Lo-Fi Maker Party, participants used paper, Post-it's and tennis balls to learn basic HTML and website structure, simple programming commands and app design.",
```
mmmavis: removed extra space

pages/events.jsx
```diff
@@ -120,7 +120,9 @@ var MakerPartyExample = React.createClass({
                 <span className="span-content label-tag">Participants</span><span className="span-content">{this.props.participants}</span>
               </div> : null
             }
-            <div className="description">{this.props.description}</div>
+            <div>
+              <p className="description" dangerouslySetInnerHTML={{__html: this.props.description}}></p>
+            </div>
```
mmmavis: https://github.com/mozilla/teach.webmaker.org/blob/develop/pages/activities.jsx#L98 should probably be wrapped in `<p>` as well but we should probably wait till we decide on https://github.com/mozilla/teach.webmaker.org/issues/1083 before we make the change.

toolness: That sounds good.

Because `description` is HTML, I think that ideally it might be good to actually call the property `descriptionHTML`, just so it's obvious to readers that it's HTML... But since we're not really following this convention anywhere else in the code yet (afaik) it's up to you if you want to make this change.

toolness:
This looks good! Feel free to implement [my one suggestion](https://github.com/mozilla/teach.webmaker.org/pull/1092#discussion_r34590222) if you want, and merge.

pages/events.jsx
mmmavis: I like this suggestion and yay we should start this naming convention from here! :smile:

PR #1133, opened 2015-07-27
Tag and release v0.0.16

mmmavis:
### Added
- Link to a sample Webmaker activity has been added to the Tools page.
- We record a GA event when any of the three Homepage CTA buttons is clicked.
- The three Made with Code activities pages that are currently on webmaker.org
  have been migrated here.
- The Private Eye activity page has been migrated from webmaker.org.

### Changed
- Our semvar string for phantomjs now matches mocha-phantomjs's(1.9.1 - 1.9.7-15).
  This is to fix the NPM error rasied from peerDependencies conflicts since the phantomjs
  we required before was at 1.9.17 but it was not a version mocha-phantomjs would take.
- Configs in .travis.yml have been updated.


hannahkane:
Webmaker activity looks good, as do the Made With Code pages. :)

mmmavis:
yayyyy thanks for reviewing @hannahkane ! gonna push it to PROD now! :tada:

PR #1207, opened 2015-08-31
Replaced MozFest banner with Thimble promo banner

### LINKED ISSUE #1166: AUGUST 31: Thimble promotion: homepage banner
@sabrinang - can you add assets?

mmmavis:
This fixes #1166.

Another PR for you to review @Pomax :dancers:


less/variables.less
```diff
@@ -18,7 +18,7 @@
 @redSidebarColor: #BB3C33;
 @pinkSidebarColor: #D64262;
 @footerBackgroundColor: #f1f1f1;
-@mozFestPromoRed: #E12526;
+@thimblePromoGreen: #279761;
```
mmmavis: Wondering if we should just give this a more general name instead (i.e., `promoBannerBackground`?) This way we don't have to update the variable name in both `less/variables.less` and `less/pages/home.less` every time we refresh the banner

Pomax: yeah that seems a better idea. That we way we won't have to rename it in the future

pages/home.jsx
```diff
@@ -314,15 +314,15 @@ var HomePage = React.createClass({
...
           </IconButtons>
         </HeroUnit>
-        <div className="row full-row mozfest-banner">
+        <div className="row full-row promo-banner">
           <div className="inner-container">
             <section>
               <Illustration
-                height={200} width={384}
-                src1x="/img/pages/home/Mozilla-Festival_Logo-Lockup_White.png" src2x="/img/pages/home/Mozilla-Festival_Logo-Lockup_White@2x.png"
+                height={190} width={190}
```
Pomax: are we basing this off of the image itself, to make sure it's sized correctly? If so, we can *probably* get that off of the file by using a node module for checking metadata on png/jpg

mmmavis: > are we basing this off of the image itself, to make sure it's sized correctly?

yup yup

> If so, we can probably get that off of the file by using a node module for checking metadata on png/jpg

Cool! Does that module automatically add `width` and `height` attributes to the underlying `` tag? (mind creating an issue for this?)

Pomax: filed as https://github.com/mozilla/teach.webmaker.org/issues/1213

Pomax:
this looks R+ to me, with a followup on possibly using a node module (or two) to find image dimensions so we don't have to hardcode in the future

pages/home.jsx
mmmavis: lovely!

PR #1275, opened 2015-09-18
update references to old repo URL

cadecairos:


cadecairos:
@jbuck R?

jbuck:
@cadecairos r+ when travis turns green

jbuck:
@cadecairos also, we should test to make sure that deploys still work on Travis

PR #1297, opened 2015-09-30
Me page

### LINKED ISSUE #1276: Implement /me page
This page will be a place for webmaker users to view their makes, once we shut down the legacy Webmaker site.

Blocked on design in #1265 

I'm not too familiar with the teach tech stack, but it sounds like we need to load and display makes client side.

cadecairos:
Fixes #1276


cadecairos:
cc @mmmavis

mmmavis:
merginggg :tada:

PR #1328, opened 2015-10-09
Tidy up colorizer mixins

### LINKED ISSUE #1002: Refactor Clubs CSS to make it more maintainable and extensible
One of the things I noticed when adding the clubs list, and am now reminded of while reviewing #1001, is that `page.less` uses the `listItemBorderColorizer` mixin to add orange borders to every `<ul>` on the Clubs page's main content, regardless of its context. This makes it *extremely* hard to have lists that are styled differently.

So, two things should be done:

- [x] Move the clubs-specific rules from `page.less` to `clubs.less`.
- [x] Scope the rule containing the `listItemBorderColorizer` to require a specific class of some sort (not sure what its name should be), and then change the existing `<ul>`'s on the clubs page to use this class.
- [x] For Club list items on `\clubs`, remove `border-top` from the `first-child` and `border-bottom` from `last-child` so that borders only appear *in between* list items.

mmmavis:
Fixed #1002 

- [x] Clean up colorizer related mixins by merging them into one mixin called `.pageColorizer()`
- [x] Borders for `.colored-list` should only appear in between the list items `<li>`

### DIFF
--- a/components/map.jsx
+++ b/components/map.jsx
@@ -59,7 +59,7 @@ var MarkerPopup = React.createClass({
   },
   render: function() {
     return (
-      <ul className="popup-clubs-list">
+      <ul className="popup-clubs-list colored-list">
         {this.props.clubs.map(function(club, i) {
           return React.createElement(MarkerPopupClub, _.extend({
             key: i,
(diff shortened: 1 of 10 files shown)


Pomax:
looks pretty solid, R+

mmmavis:
:tada:

PR #1334, opened 2015-10-13
added copy and loading animation for /me

mmmavis:
related to #1276

### DIFF
--- a/less/pages/makes.less
+++ b/less/pages/makes.less
@@ -63,3 +63,31 @@
   }
 }
 
+@keyframes loading {
+  0% {
+    width: 0;
+  }
+  33% {
+    width: .6rem;
+  }
+  66% {
+    width: 1.2rem;
+  }
+  100% {
+    width: 1.8rem;
+  }
+}
+
+.loading-message {
+  &:after {
+    content: "...";
+    display: inline-block;
+    overflow: hidden;
+    width: 0;
+    margin-left: .5rem;
+    font-size: 2.5rem;
+    vertical-align: bottom;
+    animation: loading 1.2s infinite;
+  }
+}
+
--- a/less/pages/makes.less
+++ b/less/pages/makes.less
@@ -2,6 +2,9 @@
   h1 {
     font-weight: 500;
   }
+  .context {
+    margin-top: 2rem;
+  }
 }
 
 .makes-list {
--- a/less/pages/makes.less
+++ b/less/pages/makes.less
@@ -70,12 +70,6 @@
   0% {
     width: 0;
   }
-  33% {
-    width: .6rem;
-  }
-  66% {
-    width: 1.2rem;
-  }
   100% {
     width: 1.8rem;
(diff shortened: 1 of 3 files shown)


less/pages/makes.less
```diff
@@ -63,3 +66,31 @@
...
+  }
+  33% {
+    width: .6rem;
+  }
+  66% {
+    width: 1.2rem;
+  }
+  100% {
+    width: 1.8rem;
+  }
```
Pomax: this looks like a linear animation, can we omit the 33% and 66% steps?

less/pages/makes.less
```diff
@@ -63,3 +66,31 @@
...
+.loading-message {
+  &:after {
+    content: "...";
+    display: inline-block;
+    overflow: hidden;
+    width: 0;
+    margin-left: .5rem;
+    font-size: 2.5rem;
+    vertical-align: bottom;
+    animation: loading 1.2s infinite;
```
Pomax: with a `linear` in there?

pages/makes.jsx
```diff
@@ -106,8 +106,9 @@ var MePage = React.createClass({
     if (!this.state.username) {
       pageContent = <span>Please sign in.</span>;
     } else if (this.state.loadingMakes) {
-      pageContent = <span>Loading Makes...</span>;
-    } else {
+      pageContent = <div className="loading-message">Loading makes</div>;
```
Pomax: What shall we call these things... "previous makes"? "older makes"? something else? /cc  @hannahkane?

mmmavis: do we still use the term "makes" to refer to things published using the **new tools**? if not, I think calling these "makes" is fine :eyes:

pages/makes.jsx
```diff
@@ -120,6 +121,7 @@ var MePage = React.createClass({
       });
       pageContent = (
         <div>
+          <p className="context">In the fall of 2015, we retired Popcorn Maker and Appmaker, as well as older versions of Thimble and X-Ray Goggles. Any makes you created with these tools are still accessible below. Projects created with the new X-Ray Goggles, Thimble, or Webmaker are accessible through those respective platforms.</p>
```
Pomax: this isn't technically true yet...

hannahkane: What's not true about it?

pages/makes.jsx
hannahkane: We're using the term "project" now. Let's say "Loading Projects..."

pages/makes.jsx
hannahkane: Let's also replace "makes" with "projects" in this copy.

mmmavis: i'm going to replace "makes" with "projects" and push this live today.  we can work on tweaks later so they dont act like blocker.  :+1: :smile:

mmmavis:
merging this now :tada:

PR #1339, opened 2015-10-14
Tag and release v0.0.24

mmmavis:
### Added
- Added 'Your Projects' page. This is a page for users to see the projects
  they created using tools that we retired, such as Popcorn Maker and Appmaker, 
  as well as older versions of Thimble and X-Ray Goggles.

### Changed
- Removed links to the old Webmaker search.


mmmavis:
i had @cadecairos review the /me page on staging :smile: pushing this live now

/cc @hannahkane

PR #1398, opened 2015-10-23
Privacy basics activities

Pomax:
fixes https://github.com/mozilla/teach.mozilla.org/issues/1353 - I've left the wikimedia images as hotlinks rather than local files, mostly because I'm not sure how we want to deal with the crediting /cc @hkane

### DIFF
--- a/pages/activities.jsx
+++ b/pages/activities.jsx
@@ -25,7 +25,7 @@ var ActivityKit = React.createClass({
     "Dee Salgal": "",
     "Leonie Van Der Linde": "",
     "Alan Levine": "http://cogdog.info",
-    "Stacy Martin": "https://www.linkedin.com/in/stacycmartin",
+    "Stacy Martin": "https://mozillians.org/en-US/u/stacy",
     "Mozilla": "https://webmaker.org",
     "Educator Inovator": "http://educatorinnovator.org/",
     "Jen Dick": "https://twitter.com/jennifer_dick",
@@ -78,12 +78,14 @@ var ActivityKit = React.createClass({
     return (
       <div className="activity-kit">
         <Illustration
-        height={165} width={225}
-        src1x={this.props.src1x}
-        src2x={this.props.src2x}
-        alt={this.props.title}
-        link={this.props.link}
-        externalLink={true}>
+          height={165} width={225}
+          src1x={this.props.src1x}
+          src2x={this.props.src2x}
+          caption={this.props.caption}
+          alt={this.props.title}
+          link={this.props.link}
+          externalLink={true}>
+
           <div className="activity-kit-content">
             <h3><OutboundLink to={this.props.link} eventLabel={this.props.link}>{this.props.title}</OutboundLink></h3>
             <div>
@@ -96,6 +98,7 @@ var ActivityKit = React.createClass({
             </div>
             <div className="description">{this.props.description}</div>
           </div>
+
         </Illustration>
       </div>
     );
@@ -142,8 +145,7 @@ var ActivitiesPage = React.createClass({
               level="Beginner"
               link="/activities/web-lit-basics/"
               developedBy="Our seasoned educator community"
-              description="This six-part series helps learners become familiar with reading, writing and participating on the web. They'll discover the foundations of the web through production and collaboration. This is an ideal kit for Mozilla Clubs."
-            />
+              description="This six-part series helps learners become familiar with reading, writing and participating on the web. They'll discover the foundations of the web through production and collaboration. This is an ideal kit for Mozilla Clubs."/>
             <ActivityKit
               src1x="/img/pages/protect-your-data/protect-your-data.png"
               src2x="/img/pages/protect-your-data/protect-your-data@2x.png"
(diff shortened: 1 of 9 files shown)


mmmavis:
> I've left the wikimedia images as hotlinks rather than local files, mostly because I'm not sure how we want to deal with the crediting

@Pomax and maybe adding a line just below the image to note the image source too?

pages/activities.jsx
```diff
@@ -154,6 +154,15 @@ var ActivitiesPage = React.createClass({
               description="These six hands-on activities engage learners in thinking critically about online privacy by creating secure passwords, understanding how and where their data is being collected, and more. This is an ideal kit for Mozilla Clubs."
             />
             <ActivityKit
+              src1x="https://upload.wikimedia.org/wikipedia/commons/9/98/SafeWalletLogo.png"
+              src2x="https://upload.wikimedia.org/wikipedia/commons/9/98/SafeWalletLogo.png"
```
mmmavis: if we don't have a retina version of image file than we don't have to explicitly pass in `src2x` :bowtie:

pages/activities.jsx
```diff
@@ -154,6 +154,15 @@ var ActivitiesPage = React.createClass({
...
+              title="Privacy Basics: Passwords, Tracking, and Data Retention"
+              level="Intermediate"
+              link="/activities/privacy-basics/"
+              developedBy="Stacy Martin, Senior Data Privacy Manager at Mozilla and the Mozilla Learning Network team"
```
mmmavis: this shows up as

screen shot 2015-10-26 at 2 02 38 pm

(notice the "and" after the comma) Which doesn't read right - Stacy *is* the Senior Data Privacy Manager at Mozilla. 

We probably need to clean up the `render` function of `ActivityKit`.  We used to **only** have comma separated string for the "DEVELOPED BY" field that's how the `authorLinks` map came in handy.  I guess that approach needs to change now before nested if-else blocks get more conplicated?  :grin:

pages/privacy-basics.jsx
```diff
@@ -0,0 +1,194 @@
...
+var Intro = React.createClass({
+  render: function () {
+    return (
+      <div>
+        <h1>Privacy Basics: Passwords, Tracking, and Data Retention</h1>
+        <section className="intro">
+          <Illustration
+          height={204} width={204}
+          src1x=" https://upload.wikimedia.org/wikipedia/commons/9/98/SafeWalletLogo.png"
+          src2x=" https://upload.wikimedia.org/wikipedia/commons/9/98/SafeWalletLogo.png"
```
mmmavis: no need to have this line :)

pages/privacy-basics.jsx
```diff
@@ -0,0 +1,194 @@
...
+    pageClassName: 'privacy-basics'
+  },
+  sections: [
+    {
+      title: "Passwords",
+      activities: [
+        {
+          title: "Passwords Agenda",
+          image1x: " https://upload.wikimedia.org/wikipedia/commons/9/98/SafeWalletLogo.png",
+          image2x: " https://upload.wikimedia.org/wikipedia/commons/9/98/SafeWalletLogo.png",
```
mmmavis: same, no need to have `image2x` if we don't have a retina version of the image :p this applies to other sections below

also it'll be rad if you can remove the extra space at the beginning :hamster:

Pomax:
@mmmavis but what does the "and" mean? Do we leave them hotlinks?

pages/privacy-basics.jsx
```diff
@@ -0,0 +1,194 @@
...
+          description: "Learners will use several different ways to create strong passwords, learning about security and privacy.",
+          link: "https://chadsansing.github.io/curriculum-testing/expanded-privacy-curriculum/creating-strong-passwords.html"
+        },
+        {
+          title: "Avoiding Password Reuse",
+          image1x: " https://upload.wikimedia.org/wikipedia/commons/9/98/SafeWalletLogo.png",
+          image2x: " https://upload.wikimedia.org/wikipedia/commons/9/98/SafeWalletLogo.png",
+          subtitle: "Understanding Security",
+          description: "Learners will see why re-using passwords is a bad idea and pick up strategies for avoiding that habit, learning about security.",
+          link: "https://d157rqmxrxj6ey.cloudfront.net/mozillalearning/11704/"
```
mmmavis: should point to http://chadsansing.github.io/curriculum-testing/expanded-privacy-curriculum/avoiding-password-reuse.html

pages/activities.jsx
Pomax: given what "erase all kittesn" does, I'm swapping that `,` for a ` - ` instead, and then it won't inject the "and" bit

mmmavis:
@Pomax tests failed https://travis-ci.org/mozilla/teach.mozilla.org/builds/87570198 (｡ŏ_ŏ) don't forget to run `npm test` before you submit a PR.  (although I often forgot about that step as well, heh)

mmmavis:
@Pomax looks awesome!

There's one thing that needs to be adjusted but let's just land this and file a follow-up ticket for that -

screen shot 2015-10-26 at 10 02 49 pm

The nested `<a>`'s makes the child link not reachable, and therefore users can't simply check out the source of the original file :grimacing:

Pomax:
how is that even autohappening... but yeah, filing as immediate followup.

Pomax:
wait @mavis your screenshot doesn't have a nested anchor? The image is wrapped by one `<a>`, and the figcaption has an empty `<a...></a>` for ... reasons? I don't know why it's there, but it's then followed by a regular `<a href="correct url">correct text</a>`.

PR #1402, opened 2015-10-27
put link wrapping in the right place

Pomax:
fixes https://github.com/mozilla/teach.mozilla.org/issues/1401

This moves the illustration image link wrapping into the ImageTag component instead of the Illustration component, so that captions don't get wrapped by links.


mmmavis:
:tada:

PR #1427, opened 2015-11-18
Souktel Development Progress

AbuAlKarmi:
It's Souktel development branch. [Not ready to merge]


simonwex:
Hey @AbuAlKarmi, please don't commit your editor-specific files. May I suggest adding .idea to the `.gitignore` file.

.gitignore
```diff
@@ -3,6 +3,8 @@ build
 node_modules
 .env
 .DS_Store
+.idea
 *~
 tag-message-*.txt
 release-*.html
+npm-debug.log
```
Pomax: small nit: this shouldn't be in the .gitignore: if there is an npm-debug.log in your dir, that npm error needs to be addressed. Pushing with an npm-debug.log file implies you've pushed code that has problems running.

components/badge.jsx
```diff
@@ -0,0 +1,62 @@
...
+         */
+        function status (statusVal) {
+            var classNames = 'label label-default ' + statusVal;
+            var statusIcon = "";
+
+            if (statusVal === undefined) {
+                return "";
+            }
+
+            switch (statusVal) {
```
Pomax: we're in React, so there's no need to build HTML using strings: we can use JS and JSX to get the job done:

```
var classnames = require("classnames");
...
var statusIcons = {
  pending: 'fa-clock-o',
  achieved: 'fa-check',
  eligible: 'fa-pencil'
};
var statusClasses = classnames('fa', 'fa-fw', statusIcons[statusVal]);
var statusIcon = <i className={statusClasses} />;

return (
  <div className="status">
    <div className={classNames}>
      <span>{ statusIcon }</span>{ statusVal }
    </div>
  </div>
);
```

components/badge.jsx
```diff
@@ -0,0 +1,62 @@
...
+        icon2x: React.PropTypes.string.isRequired
+    },
+    render: function (){
+
+        /**
+         * returning the Status HTML
+         * @param statusVal
+         * @returns ReactElement
+         */
+        function status (statusVal) {
```
Pomax: try not to declare functions inside of `render`, because that rebuilds them on every render call. Instead, we can define them as part of the Badge class prototype:
```
var Badge = React.createClass({
  proptTypes: ...,
  status: function(statusVal) {
    ...
  },
  render: function() {
    ...
      statusElement  = this.status(...);
    ...
  }
...
});
```

components/badge.jsx
```diff
@@ -0,0 +1,62 @@
...
+                    break;
+            }
+            return (
+                <div className="status">
+                    <div className={classNames}><span dangerouslySetInnerHTML={{__html: statusIcon}}></span>{statusVal}</div>
+                </div>
+            );
+        }
+
+        var statusElement = "";
```
Pomax: see note above. We can then do this as:
```
var statusElement = this.status();
```
where `this.status()` is responsible for checking `this.props.status`, return `false` if not set, and calling a `.toLowerCase()` when it is.

components/requirement-list.jsx
```diff
@@ -0,0 +1,47 @@
...
+        className   : React.PropTypes.string,
+        icon        : React.PropTypes.string
+    },
+    render: function () {
+        var className = this.props.className ? 'list-unstyled requirements-list' + this.props.className : 'list-unstyled requirements-list';
+        var icon = this.props.icon ? this.props.icon : 'fa fa-check';
+        var i = 0;
+        var listItems = this.props.list.map(function(item){
+            i++;
+            return <RequirementRow key={i} icon={icon} text={item} />
```
Pomax: `key` should not be dependent on the list position, but should be based on the item itself. It's used to determine whether the dom diff needs to delete/insert, or simply move things (much faster).

That said, `map` and `forEach` etc can take a second argument so you don't need an external counter:
```
['a','b','c','d'].map(function(letter, position) {
  console.log(position + ": " + letter);
});
```
generates
```
0: a
1: b
2: c
3: d
```

components/social-share.jsx
```diff
@@ -0,0 +1,46 @@
+var React = require('react');
+
+var SocialShareButton = React.createClass({
+    propTypes : {
+        networkType: React.PropTypes.string.isRequired,
+        iconClassName: React.PropTypes.string.isRequired
+    },
+    render : function () {
+        return (
+            <a href="#" className={"btn icon-only btn-social " + this.props.networkType }>
```
Pomax: `href="#"` means "navigate to the top of this page" under HTML5 rules, so if this is a button, it should probably be a `<button>` with styling to make it look like how we want it to look, rather than be a link. If there is no url, it's not a link =)

package.json
```diff
@@ -24,6 +24,7 @@
     "gulp-util": "^3.0.4",
     "gulp-webpack": "^1.2.0",
     "imports-loader": "^0.6.3",
+    "jquery": "^2.1.4",
```
Pomax: since we're using React, we shouldn't need jQuery.

pages/badge-single.jsx
```diff
@@ -0,0 +1,217 @@
+var React = require('react'),
+    HeroUnit = require('../components/hero-unit.jsx'),
+    SocialShare = require('../components/social-share.jsx'),
+    BadgeHorizontalIcon = require('../components/badge-horizontal-icon.jsx'),
+    RequirementsList = require('../components/requirement-list.jsx'),
+    Badge = require('../components/badge.jsx'),
+    Link = require('react-router').Link,
+    $ = require('jquery'),
```
Pomax: React and ES2015 remove the need for jQuery and underscore, so it's worth taking these back out.

pages/badge-single.jsx
```diff
@@ -0,0 +1,217 @@
...
+ * should take a badgeId
+ */
+var BadgePage = React.createClass({
+    statics: {
+        pageTitle: 'Badges',
+        pageClassName: 'badges single-badge'
+    },
+    onQualificationsSubmit : function ( event ){
+        event.preventDefault();
+        console.log( $( event.target ).serializeArray() );
```
Pomax: if we remove jQuery, a `console.log(event.nativeEvent.target)` should work for debugging purposes

pages/badge-single.jsx
```diff
@@ -0,0 +1,217 @@
...
+    },
+    handleFileSelect : function (event) {
+        $( this.refs.optionalFile.getDOMNode() ).click();
+    },
+    getInitialState: function () {
+        return {
+            badge: {
+                key: 1,
```
Pomax: better to use the badge title as key

pages/badge-single.jsx
```diff
@@ -0,0 +1,217 @@
...
+            <div>
+
+                <div className="row top-back-navigation">
+                    <div className="col-md-12">
+                        <Link className="text-uppercase btn btn-link" to="/badges/"><span className="fa fa-long-arrow-left"></span> Back to credentials</Link>
+                    </div>
+                </div>
+
+                <BadgeHorizontalIcon
+                    key={this.state.badge.key}
```
Pomax: `.title`

pages/badge-single.jsx
```diff
@@ -0,0 +1,217 @@
...
+    statics: {
+        pageTitle: 'Badges',
+        pageClassName: 'badges single-badge'
+    },
+    onQualificationsSubmit : function ( event ){
+        event.preventDefault();
+        console.log( $( event.target ).serializeArray() );
+    },
+    handleFileSelect : function (event) {
+        $( this.refs.optionalFile.getDOMNode() ).click();
```
Pomax: As the `click` function exists on the This should work fine without jQuery: `this.refs.optionalFile.getDOMNode().click()`

pages/badge-single.jsx
```diff
@@ -0,0 +1,217 @@
+var React = require('react'),
+    HeroUnit = require('../components/hero-unit.jsx'),
+    SocialShare = require('../components/social-share.jsx'),
+    BadgeHorizontalIcon = require('../components/badge-horizontal-icon.jsx'),
+    RequirementsList = require('../components/requirement-list.jsx'),
+    Badge = require('../components/badge.jsx'),
+    Link = require('react-router').Link,
+    $ = require('jquery'),
+    _ = require('underscore');
```
Pomax: `_` does not appear to be used in this file.

pages/badges.jsx
```diff
@@ -0,0 +1,192 @@
+var React = require('react'),
+    ImageTag = require('../components/imagetag.jsx'),
+    HeroUnit = require('../components/hero-unit.jsx'),
+    IconLinks = require('../components/icon-links.jsx'),
+    IconLink = require('../components/icon-link.jsx'),
+    BadgeVerticalIcon = require('../components/badge-vertical-icon.jsx'),
+    Link = require('react-router').Link,
+    _ = require('underscore');
```
Pomax: with ES2015 we shouldn't need underscore for anything anymore.

pages/badge-single.jsx
```diff
@@ -0,0 +1,217 @@
...
+                        nextNavigationDetails={this.state.navigation.next}
+                        prevNavigationDetails={this.state.navigation.prev} />
+                </div>
+
+            </div>
+        );
+    }
+});
+
+module.exports = BadgePage;
```
Pomax: this file can probably do with a bit of whitespace cleanup (but with a newline at the end of the file so github doesn't flag it as problematic)

pages/badges.jsx
```diff
@@ -0,0 +1,192 @@
...
+
+var BadgesPage = React.createClass({
+    statics: {
+        pageTitle: 'Badges',
+        pageClassName: 'badges'
+    },
+    getInitialState: function () {
+        var dummyBadges = [
+            {
+                key : 1,
```
Pomax: better not to use plain numbers as keys. Using `title` for the key is a safer approach.

pages/badges.jsx
```diff
@@ -0,0 +1,192 @@
...
+    Link = require('react-router').Link,
+    _ = require('underscore');
+
+var BadgesPage = React.createClass({
+    statics: {
+        pageTitle: 'Badges',
+        pageClassName: 'badges'
+    },
+    getInitialState: function () {
+        var dummyBadges = [
```
Pomax: Probably better to put this outside the React class definition, so it doesn't get rebuilt every time an instance is created.

pages/badges.jsx
```diff
@@ -0,0 +1,192 @@
...
+                icon: '/img/components/badge-icon/ProblemSolving.png',
+                icon2x: '/img/components/badge-icon/ProblemSolving@2x.png'
+            }
+        ];
+
+        return {
+            badges: dummyBadges
+        };
+    },
+    componentDidMount: function () {},
```
Pomax: better to leave this out if it doesn't have a function body.

pages/badges.jsx
```diff
@@ -0,0 +1,192 @@
...
+        this.setState({
+            badges: parsed
+        });
+    },
+    parseBadges: function (credlyResponse) {
+        // do parsing here
+        var data = credlyResponse;
+
+        if (data.status === '200' && data.body && data.body.data && data.body.data.length) {
+            data = _.map(data.body.data, function (badgeData) {
```
Pomax: this shouldn't need underscore. If the `data` isn't a true array we can use `Array.from` to make it one:
```
data = Array.from(response.body.data).map(function(badgeData) { 
  ...
});
```

pages/badges.jsx
```diff
@@ -0,0 +1,192 @@
...
+    componentDidMount: function () {},
+    setBadgesData: function (error, data) {
+        var parsed = this.parseBadges(data);
+        this.setState({
+            badges: parsed
+        });
+    },
+    parseBadges: function (credlyResponse) {
+        // do parsing here
+        var data = credlyResponse;
```
Pomax: better to say `var data = []` here, and then use `credlyResponse` in the conditional, so we don't need an `else` later:

pages/badges.jsx
```diff
@@ -0,0 +1,192 @@
...
+
+        if (data.status === '200' && data.body && data.body.data && data.body.data.length) {
```
Pomax: `if (response.status === ... && response.body && ...)`

pages/badges.jsx
```diff
@@ -0,0 +1,192 @@
...
+            data = _.map(data.body.data, function (badgeData) {
+                return {
+                    'title' : badgeData.title || '',
+                    'status' : 'Achieved',
+                    'description': badgeData.criteria || '',
+                    'icon': badgeData.image || '',
+                    icon2x: badgeData.image || ''
+                };
+            });
+        } else {
```
Pomax: with that small rewrite, we can get rid of this else

pages/badges.jsx
```diff
@@ -0,0 +1,192 @@
...
+        return data;
+    },
+    render: function () {
+        var badges = this.state.badges.map(function (badge) {
+            var link = '/badge/' + badge.key;
+            return (
+                <div key={badge.key} className="col-md-4">
+                    <Link to={link} className={'badge-icon-link'}>
+                        <BadgeVerticalIcon
+                            key={badge.key}
```
Pomax: `title`

pages/badges.jsx
```diff
@@ -0,0 +1,192 @@
...
+                            { badges }
+                        </div>
+                    </section>
+                </div>
+            </div>
+        );
+    }
+});
+
+module.exports = BadgesPage;
```
Pomax: Github likes to have newlines at the end of files. It's a bit silly, but easy to add in, so we tend to make sure files always end on a newline.

Pomax:
Love how compact you've kept things - that makes it really easy to review, as well as maintain =) I've left some comments mostly around React patterns and some JS features and patterns that are available in modern JS that make things a little less code heavy.

Pomax:
Would it be possible to change the badge links so that instead of pointing to `http://localhost:8008/badge/4`, it points to `http://localhost:8008/badge/collaboration`? The more human friendly content URLs are, the easier it'll be for people to navigate their way around them.

components/badge.jsx
```diff
@@ -0,0 +1,66 @@
...
+                </div>
+            </div>
+        );
+
+    },
+    render: function (){
+
+        var statusElement = "";
+        if( this.props.status ){
+            statusElement = this.status( this.props.status );
```
Pomax: because of how `this.status` now returns the empty string if there is no `this.props.status` value top work with, we don't need the `if` anymore. We can use a one line `var statusElement = this.status(this.props.status)`, although since the `this.status` function is responsible for simply saying "what the status is", we can even say this:
```
var statusElement = this.status();
```
and then rather than taking an argument, have `this.status` responsible for checking:
```
status: function() {
  if (!this.props.status) {
    return '';
  }
  var statusVal = this.props.status.toLowerCase();
  ...
}
```

components/requirement-list.jsx
```diff
@@ -0,0 +1,45 @@
...
+    propTypes: {
+        list        : React.PropTypes.array.isRequired,
+        className   : React.PropTypes.string,
+        icon        : React.PropTypes.string
+    },
+    render: function () {
+        var className = this.props.className ? 'list-unstyled requirements-list' + this.props.className : 'list-unstyled requirements-list';
+        var icon = this.props.icon ? this.props.icon : 'fa fa-check';
+        var listItems = this.props.list.map(function(item,position){
+            return <RequirementRow key={position + '-' + item} icon={icon} text={item} />
```
Pomax: This is a bad idea for the key: If we reorder the items in this array, using a `sort()` for instance, then React will now think that every item in the array is new, and that all the old ones are gone, so it will delete all the DOM nodes, and then rebuild them, instead of simply swapping the ordering without deleting/creating DOM nodes. If the text in the items is uniquely identifying, `key={item}` is the right key.

npm-debug.log
```diff
@@ -0,0 +1,38 @@
+0 info it worked if it ends with ok
```
Pomax: you'll want to remove this file from your root dir.

pages/badges.jsx
```diff
@@ -0,0 +1,190 @@
...
+                };
+            });
+        }
+
+        return data;
+    },
+    render: function () {
+        var linkUrl = "";
+        var badges = TESTING_DATA.map(function (badge) {
+            linkUrl = '/badge/' + urlize( badge.id + '-' + badge.title );
```
Pomax: while not necessary to change here I think, it's good to know that ES6 has a new string format that reduces the need for string concatenation using `+`:
```
linkUrl = '/badge/' + urlize(`${badge.id}-${badge.title}`);
```
Here, the backticks set up a template string, and the `${...}` are variable references

Pomax:
marking as "needs work" for the last few small things.

Pomax:
closing this out, as the remaining work is being done in https://github.com/mozilla/teach.mozilla.org/tree/souktel-improvements

Issue #1436, opened 2015-11-21
Add platform-specific instructions to the README


simonwex:
Continue the work that was done in this PR: https://github.com/mozilla/teach.mozilla.org/pull/1430

(See PR for discussion)

PaulaPaul:
Hello @simonwex - I just tidied up the README to separate instructions for Mac - Linux - Windows/other (and tested this on Mac); I just issued a pull request and realized I did not include this issue number... let me know if you want me to re-issue.  Thanks!

simonwex:
Hey @PaulaPaul your changes are great, thank you!

Not mentioning the issue number is no biggie. The only issue is that it seems like your PR might have rolled changes in from another branch, too? You can see a bunch of other commits I don't think you intend:
http://downloads.simonwex.com/grabs/Clementine.2015-11-24_20-47-43.png

Do you mind trying to resubmit without the other changes? If you want a hand with this, let me know I'm happy to walk you through it.

You can find me as "wex" on the mozilla irc in the #teach channel (https://wiki.mozilla.org/IRC) or you can find me on skype as `simon.wex` or here!

PaulaPaul:
Sorry for the trouble @simonwex!  I'm working through what it means to sync a fork and contribute to the up-to-date repo and see where I went wrong... thanks for your help and patience.
I've closed my initial pull request for this and have re-issued (including the issue number this time).  I will look you up on irc to get tips on the process.  Would like to tackle some simple code issues next!

simonwex:
@PaulaPaul, no trouble at all.

Hey @Pomax & @mmmavis  now that you're back, I just wanted you to "meet" @PaulaPaul. She's added some real value to the README. Do you have any good "first bugs" she could start on?

@PaulaPaul, @Pomax and @mmmavis primarily work on this repo you can find them in the IRC channel, too as `pomax` and `mavis`.

mmmavis:
Hey @PaulaPaul :tada: :dancers: :dancers: :dancers:  Thanks for updating the README! 

There's a ["help wanted" label in this repo](https://github.com/mozilla/teach.mozilla.org/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22).  I will go through the tickets today and mark some as "help wanted".  Feel free pick one (or more :wink: ) that interests you the most!  :+1: Thanks again for spending the effort contributing to this repo!

PR #1451, opened 2015-11-25
#1436: Updated README.md to include platform specific setup info

PaulaPaul:
Re-issued pull request with only the commits for README.md updates.  Added separate setup instructions in README for Mac, Linux and Windows/other.


simonwex:
Thanks, @PaulaPaul. Looks good to me!

PR #1455, opened 2015-11-28
Simplified sidebar

### LINKED ISSUE #1397: Implement simplified sidebar
Based on design https://github.com/mozilla/teach.mozilla.org/issues/1370

Redpen comp: https://redpen.io/p/jh0a4bc04ea2daf52b
Illustrator file: https://drive.google.com/a/mozilla.com/file/d/0B2irLVDJivhLUEtuYUI5cVJJUGc/view?usp=sharing
Assets: https://drive.google.com/a/mozilla.com/folderview?id=0B2irLVDJivhLV3lXS2pOd21QblE&usp=sharing

I guess ignore Credentials for now and ignore Hive as a footer link for now https://github.com/mozilla/teach.mozilla.org/issues/1368

cc/ @mmmavis @hannahkane

mmmavis:
Fixes #1397


mmmavis:
This is ready for review! @Pomax :smiley_cat:

components/sidebar.jsx
```diff
@@ -5,27 +5,85 @@ var LinkAnchorSwap = require('./link-anchor-swap.jsx');
 
 var Login = require('./login.jsx');
 
+var Subitems = React.createClass({
+  render: function() {
+    var items = this.props.subItems.map(function (item, key) {
```
Pomax: better not to specify `key` if we don't use it

components/sidebar.jsx
```diff
@@ -5,27 +5,85 @@ var LinkAnchorSwap = require('./link-anchor-swap.jsx');
...
+    };
+  },
+  componentDidMount: function() {
+    this.highlightActiveSubNav();
+  },
+  componentDidUpdate: function() {
+    this.highlightActiveSubNav();
+  },
+  highlightActiveSubNav: function() {
+    var activeState = !!this.getDOMNode().querySelector(".sidebar-subitems a.active");
```
Pomax: the fact that we need DOM querying here is a sign we're doing something wrong. React doesn't need DOM queries for anything, if done right; this should be a thing we can simply ask the relevant `this.refs` component. What does this code achieve, and can we do that "directly"?

components/sidebar.jsx
```diff
@@ -5,27 +5,85 @@ var LinkAnchorSwap = require('./link-anchor-swap.jsx');
...
+    if (activeState != this.state.activeSubNav) {
+      this.setState({
+        activeSubNav: activeState
+      });
+    }
+  },
+  render: function() {
+    var classes = this.props.className + " top-level-item";
```
Pomax: This is what `classnames` is ideal for:
```
var classnames = require("classnames");
...
var classes = classnames(this.props.className, "top-level-item", { sub-nav-active: this.state.activeSubNav });
```

components/sidebar.jsx
```diff
@@ -59,6 +109,12 @@ var Sidebar = React.createClass({
           link: "web-literacy"
         }
       ]
+    },
+    {
+      name: "Community",
+      href: 'http://discourse.webmaker.org/',
```
Pomax: https?

Pomax:
I'm seeing dotted borders when making menu selections, are those supposed to be there? (it feels like they shouldn't?)

Pomax:
I only see sub items on "maker party" and "teach like mozilla". Is that correct?

components/sidebar.jsx
```diff
@@ -5,27 +5,85 @@ var LinkAnchorSwap = require('./link-anchor-swap.jsx');
...
+var TopLevelNavItem = React.createClass({
+  getInitialState: function() {
+    return {
+      activeSubNav: false
+    };
+  },
+  componentDidMount: function() {
+    this.highlightActiveSubNav();
+  },
+  componentDidUpdate: function() {
```
Pomax: why not tap into the arguments this gets passed?
```
componentDidUpdate: function(oldProps, oldState) {
  ...
}
```
Then you can check whether the values from the old props/state are different from the current this.props and this.state, and call functions accordingly

mmmavis: ha, totally forgot that they come with oldProps, oldState arguments! Thanks for reminding me!

mmmavis:
@Pomax 

> I'm seeing dotted borders when making menu selections, are those supposed to be there? (it feels like they shouldn't?)
fixed

> I only see sub items on "maker party" and "teach like mozilla". Is that correct?
yup!

mmmavis:
@Pomax 

Also the "post undefined" error you discovered yesterday wasn't related to this PR. I've enter a separate ticket to fix that: https://github.com/mozilla/teach.mozilla.org/issues/1465

components/sidebar.jsx
```diff
@@ -1,31 +1,105 @@
...
 
+var Subitem = React.createClass({
+  componentDidMount: function() {
+    this.toggleHighlight();
+  },
+  componentDidUpdate: function() {
+    this.toggleHighlight();
+  },
+  toggleHighlight: function() {
+    var isActive = this.getDOMNode().classList.contains("active");
```
Pomax: This should be either from props, or from state. We should *never* have to consult the DOM node (since as far as React cares, that DOM node is an implementation detail).
```
getIniitialState: function() {
  return {
    active: false
  };
},
highlight: function() { this.setState({ active: true; }); },
unhighlight: function() { this.setState({ active: false; }); },
...
render: function() {
  // render this element while taking `this.state.active` into account
}
```

Pomax:
R+  if we file a followup ticket to figure out how to avoid DOM accesss as fast-followup.

mmmavis:
tsk tsk React Router :eyes:

PR #1480, opened 2015-12-05
React v14

Pomax:
partially addresses https://github.com/mozilla/teach.mozilla.org/issues/1460

This puts us on React 14, with code changes to make everything work again, *however*: there are some parts in the test suite that do not adhere to proper React 14 practices yet. They're warnings right now, but in a future version they'll be errors, so we should still try to fix them.

Note that this PR still leaves the obsolete lodash dependencies (which we're inheriting from nested modules, hopefully we can update whatever packages cause those still)


mmmavis:
will review this later today

mmmavis:
@Pomax I'm getting 
```
Warning: React.render is deprecated. Please use ReactDOM.render from require('react-dom') instead.
```

Should we tackle this in a followup PR?

mmmavis:
arrr :heavy_plus_sign: :+1: yay

mmmavis:
Woohoooo thanks! @Pomax

PR #1494, opened 2015-12-16
Dependency uplifting

Pomax:
addresses the rests of https://github.com/mozilla/teach.mozilla.org/issues/1460

**Current status**: everything except react-router has been uplifted. The router update changes the code calls so requires some careful reading up on how things changed.

Once that's been uplifted, too, we can start to move the build system over to babel (and possibly using npm scripts for all the task, rather than using gulp wrapping)


mmmavis:
verified, `npm test` is error and warning free. R+ :tada:

PR #1560, opened 2016-01-22
refactored gulpfile into component tasks

Pomax:
fixes https://github.com/mozilla/teach.mozilla.org/issues/1559

This is a pure refactor, without code changes other than relocations to their own files and the necessary wrapper code to call functions from these new files. Each set of tasks that seemed a logical cluster has been put in its own file in the `./gulp/` directory, with shared resources housed in `./gulp/shared/`.

The "npm start" command should still do exactly the same thing, but some of the gulp tasks that are used less frequently might still be missing dependencies, so if review can be run on all gulp tasks, that'll help us rule out missing deps., typos, etc.

This refactor allows us to examine which tasks we actually still use, and following that, how we can change as many of them as possible to "npm scripts" instead.


mmmavis:
r++++ :heavy_plus_sign:

PR #1567, opened 2016-01-25
removed lwip conversion entirely

Pomax:
fixes https://github.com/mozilla/teach.mozilla.org/issues/1565 by simply no longer doing the lwip conversion step - I don't see any missing images, but if there are, we should just put the images we need where we need them.


mmmavis:
R plus

PR #1583, opened 2016-01-26
refactored makes.jsx

Pomax:
partially adresses https://github.com/mozilla/teach.mozilla.org/issues/1571 with a minor refactor to makes.jsx - I'm keeping the single "Make" component inside the Makes.jsx file for now, since it doesn't feel like it warrants splitting (nothing else uses it as far as I can tell right now)


mmmavis:
:bowtie: R+

PR #1607, opened 2016-01-29
converted modal context to plain props passthrough

Pomax:
Fixes https://github.com/mozilla/teach.mozilla.org/issues/1555, fixes https://github.com/mozilla/teach.mozilla.org/issues/1597

This PR replaces the implicit use of `showModal` and `hideModal` through a mixin (a pattern that has been discontinued in React) with normal property passthrough, instead.

Browser tests have been updated accordingly.

Places where we invoke the modal: 
- [x] home page, after `localStorage.clear()` (FTU pledge modal)
- [x] Dev-ribbon (opens dev modal, in dev mode only)
- [x] http://localhost:8008/clubs, "apply" button
- [x] http://localhost:8008/clubs/list/, "apply" button
- [x] http://localhost:8008/?pledge=thanks, "thank you" modal


pages/clubs-list/ClubsListPage.jsx
```diff
@@ -30,6 +29,7 @@ var ClubsListPage = React.createClass({
     pageClassName: 'clubs-list-page'
   },
   componentDidMount: function() {
+    console.log(this.props);
```
mmmavis: :x: :grey_question: :grin:

mmmavis:
There's also this Thank You modal that is also used as the redirected designation after users submitted form on the MLN site (you can get to it via http://localhost:8008/?pledge=thanks.

It seems like there's no way to close that particular modal. Tried both clicking the "x" on top right as well as hitting ESC.

Pomax:
good catch, removed the console log and added the `hideModal` to the ThankYouModal

mmmavis:
ﾍ( ´∀`)ﾉ

PR #1615, opened 2016-02-02
Preparing `router` context for migration

Pomax:
Some cleanup around how the Page and current router are used in the code, and the router context (which is `this.props.router` in react-router 1.0)

These changes pass all tests, but there are warnings about setState() for the Select component being called while it's not mounted. This suggests that the modal for it expires during testing, but it also doesn't seem to affect testing, so something odd is going on here.

Note that some files were touched in order to make the test log less polluted.


mmmavis:
PR looks good to me. I wish there's some built-in router icon that I could attach here though. :stuck_out_tongue_winking_eye:

Pomax:
hmm... :railway_car::railway_car::railway_car::railway_car::railway_car: ?

PR #1617, opened 2016-02-02
anchormanager as hoc instead of mixin

Pomax:
fixes https://github.com/mozilla/teach.mozilla.org/issues/1554


mmmavis:
:anchor: :squirrel: :+1:

PR #1627, opened 2016-02-06
React router 1.0

Pomax:
this PR updates the codebase to use react-router 1.0 -- `npm start` works, but a lot of the tests still fail, and there are a few API deprecation warnings, so there's a little bit of work left to be done, but slotting in React-Router 1.0 this time round was actually entirely doable, as opposed to absolutely impossible.

Do not merge yet (although if you want to test the site via `npm start`, by all means, do)


test/app.test.js
```diff
@@ -72,14 +77,6 @@ describe('app', function() {
       .end(done);
   });
 
-  it('adds trailing slash to public HTML pages', function(done) {
```
mmmavis: should we keep this?

Pomax: the code no longer does this redirect, so we don't have to test for it.

components/page.jsx
```diff
@@ -125,7 +135,13 @@ var Page = React.createClass({
           <div className="row">
             <Sidebar/>
             <main className="content col-md-9" role="main" id="content" tabIndex="-1">
-              <RouteHandler showModal={this.showModal} hideModal={this.hideModal}/>
+            {
+              React.cloneElement(this.props.children, {
```
mmmavis: could you explain why we are doing `cloneElement` vs just passing extra `props` to the existing element?

mmmavis:
:clock5: = :chocolate_bar: :cake: :beers: :cocktail: :tada:

PR #1638, opened 2016-02-15
Pass currentPath to <Sidebar> 

### LINKED ISSUE #1636: <Sidebar> needs to know which page it is currently on
Looks like `currentPath` isn't being passed correctly - in `<Sidebar>` that prop is `undefined`. We should fix that so teach-api knows which page to redirect users back after they sign in/up/out.

mmmavis:
Fixes #1636


Pomax:
aye, R+

PR #1663, opened 2016-02-22
Email validation broken

alicoding:
We are missing `underscore` module therefore the validation util is throwing an error


alanmoo:
I haven't pulled it but I can't see how simply requiring underscore will break anything. And I see where the _ variable is required, so R+

PR #1685, opened 2016-03-04
server processes now say which server file is involved

Pomax:
fixes https://github.com/mozilla/teach.mozilla.org/issues/1684


mmmavis:
This is a lot more clearer. :+1:

PR #1687, opened 2016-03-07
Render pages created on WP

### LINKED ISSUE #1686: Add ability to create new pages from within Calypso
implementation ticket for https://github.com/mozilla/wp-calypso/issues/12

mmmavis:
Fixes #1686


app.js
```diff
@@ -119,10 +122,44 @@ app.use(function(req, res, next) {
...
+      var PageContent = React.createElement(Page, 
+        {
+          routes: [
+            { path: '', component: { pageClassName: '', pageTitle: ''} }
+          ] 
+        }, 
+        React.createElement('div', { dangerouslySetInnerHTML: { __html: wpContent }})
+      );
+
+      res.status(200).send("<!DOCTYPE html>" + ReactDOMServer.renderToStaticMarkup(PageContent));
```
mmmavis: (Note that `PageContent` will need to be wrapped in HTML/Page wrapper still, e.g., `<html><head>...</head><body>...</body></html>`)

1. run `npm run app`
2. visit `http://localhost:8008/new-page` (`new-page` is a slug exists in the WP instance)
3. `ReactDOMServer.renderToStaticMarkup` throws an error as below
```bash
Warning: React.createElement: type should not be null, undefined, boolean, or number. It should be a string (for DOM elements) or a ReactClass (for composite components). Check the render method of `Page`.

/Users/mavisou/Development/teach.webmaker.org/node_modules/fbjs/lib/invariant.js:39
      error = new Error(format.replace(/%s/g, function () {
              ^
Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object. Check the render method of `Page`.
    at invariant (/Users/mavisou/Development/teach.webmaker.org/node_modules/fbjs/lib/invariant.js:39:15)
    at instantiateReactComponent (/Users/mavisou/Development/teach.webmaker.org/node_modules/react/lib/instantiateReactComponent.js:64:134)
    at instantiateChild (/Users/mavisou/Development/teach.webmaker.org/node_modules/react/lib/ReactChildReconciler.js:29:28)
    at traverseAllChildrenImpl (/Users/mavisou/Development/teach.webmaker.org/node_modules/react/lib/traverseAllChildren.js:98:5)
    at traverseAllChildrenImpl (/Users/mavisou/Development/teach.webmaker.org/node_modules/react/lib/traverseAllChildren.js:114:23)
    at traverseAllChildren (/Users/mavisou/Development/teach.webmaker.org/node_modules/react/lib/traverseAllChildren.js:186:10)
    at Object.ReactChildReconciler.instantiateChildren (/Users/mavisou/Development/teach.webmaker.org/node_modules/react/lib/ReactChildReconciler.js:52:5)
    at ReactDOMComponent.ReactMultiChild.Mixin._reconcilerInstantiateChildren (/Users/mavisou/Development/teach.webmaker.org/node_modules/react/lib/ReactMultiChild.js:197:41)
    at ReactDOMComponent.ReactMultiChild.Mixin.mountChildren (/Users/mavisou/Development/teach.webmaker.org/node_modules/react/lib/ReactMultiChild.js:232:27)
    at ReactDOMComponent.Mixin._createContentMarkup (/Users/mavisou/Development/teach.webmaker.org/node_modules/react/lib/ReactDOMComponent.js:591:32)
```
I found out the error came from our custom React components used in `Page.jsx` (e.g., `<Sidebar>`, `<Footer>` etc). If we remove those components from the Page's render then the errors are gone. I'm not exactly sure why... `babel-preset-react` is there and if it's complaining about `jsx` wouldn't we hit error at the first "<" occurrence in `Page.jsx`?

@Pomax any pointers? :cold_sweat:

package.json
```diff
@@ -83,6 +85,7 @@
     "react-ga": "^1.0.12",
     "react-router": "^1.0.3",
     "react-select": "^0.9.1",
+    "react-wp-content-loader": "latest",
```
Pomax: `"react-wp-content-loader": "git://github.com/mmmavis/react-wp-content-loader.git",` =P

mmmavis: Oh yea, I've been `npm link` them since I'm developing both at the same time. Should've changed this to "git://github.com/mmmavis/react-wp-content-loader.git".  Let me update my PR.

app.js
mmmavis: Updated discovery - is it only `<Sidebar>` that's causing errors. Including `<Footer>` seems fine.

mmmavis: Updated discovery -

It's related some path issue, namely if you have a `file.extension` and a directory called `file`, things may not work all that well.

In this case `Sidebar` was module exported via `require('./sidebar')`

mmmavis: suggestion from @Pomax 

> we need to say `module.exports = require('./sidebar/');` if there's a possible naming conflict due to implicit file extensions

app.js
```diff
@@ -94,34 +99,24 @@ app.use(function(req, res, next) {
...
     if (props) {
       indexStatic.generate(location, {}, function(err, location, title, html) {
         if (err) {
-          return next(err);
-        }
-        return res.type('html').send(html);
-      });
-    }
-    // this is not a specific component - although it might be if we massage the path:
-    else {
```
mmmavis: React Router v1+ takes care of path both with and without trailing slash.

test/app.test.js
```diff
@@ -30,7 +30,7 @@ describe('app', function() {
       return redirects.indexOf(route) === -1;
     }).map(function(url) {
       if (url === '/') return '/';
-      return '/' + url;
+      return '/' + url + '/';
```
mmmavis: for some reason, without trailing slash we'll get 303 instead 200... :grimacing:

lib/routes.jsx
```diff
@@ -67,12 +67,19 @@ var redirectElements = Object.keys(redirects).map(function(path) {
...
+// as the WP page slug in a WP API call. If WP sends back 200, we display the content.
+// Else error message will be shown on the page. Note that the API call is made on client side.
+// With this setup there will never be any status 404 page.
 var routes = (
-  <Route path='/' component={require('../components/page.jsx')} >
-    <IndexRoute component={require('../pages/home.jsx')} />
-    {routeElements}
-    {redirectElements}
-  </Route>
+    <Route path='/' component={require('../components/page.jsx')} >
```
Pomax: why did this indent go from 2 to 4?

package.json
```diff
@@ -83,6 +83,7 @@
     "react-ga": "^1.0.12",
     "react-router": "^1.0.3",
     "react-select": "^0.9.1",
+    "react-wp-content-loader": "git://github.com/mmmavis/react-wp-content-loader.git",
```
Pomax: let's not do this: even if it's alpha, just publish it to npm, and let's make package.json load it from there

pages/wp-content.jsx
```diff
@@ -0,0 +1,15 @@
+var React = require('react');
+var Page = require('../components/page.jsx');
+var WpContentLoader = require('react-wp-content-loader');
+
+// [TODO] 404 needs to be handled properly. Now we just make all the 
+//        non-component pages as wp-pages
+var WpContent = React.createClass({
+  render: function() {
+    return (
+      <WpContentLoader wpUrl="teachmozillaorg.wordpress.com" wpPostSlug={this.props.params.wpSlug} />
```
Pomax: as a [stateless component](https://toddmotto.com/stateless-react-components/), we can probably rewrite this:
```
var WpContent = function(props) { return <WpContentLoader wpPostSlug={props.params.wpSlug} ... />; };
```

test/app.test.js
Pomax: That makes sense: app.js does a redirect; so the question is do we test to make sure that redirect does the right thin in addition to changing this string?

test/app.test.js
```diff
@@ -77,10 +77,11 @@ describe('app', function() {
       .end(done);
   });
 
-  it('reports 404s', function(done) {
-    request(app)
-      .get('/asdfasdfasdf')
-      .expect(404)
-      .end(done);
-  });
+  // [Note] With our React Router routes setup, there will never be any 404 page. (See routes.jsx)
```
Pomax: This seems wrong - we do generate a 404 page.

lib/routes.jsx
mmmavis: heh, will fix

test/app.test.js
mmmavis: I can'ts seem to understand why it is 303 instead of 302. :fearful:

test/app.test.js
mmmavis: Yup, totally agree. There's shouldn't be a 200 pretending to be a 404 page. I don't like my approach here either so that's why I came to you for suggestions ha.

test/app.test.js
Pomax: 303 is when it's a "moved" for things that are not plain GET operations - it's a "moved and what you expected to happen probably didn't happen" code

test/app.test.js
Pomax: let's take out this test entirely - it'll still be in the git log

pages/wp-content.jsx
```diff
@@ -0,0 +1,13 @@
+var React = require('react');
+var WpContentLoader = require('react-wp-content-loader');
+var NotFoundMessage = require('../components/not-found.jsx');
+
+var WpContent = function(props) { 
+  return (
+    <WpContentLoader wpUrl="teachmozillaorg.wordpress.com" wpPostSlug={props.params.wpSlug}>
```
Pomax: should we make this an env variable? It feels dangerous to hardcode this

test/app.test.js
Pomax: so it looks like this is being caused by the `app.use('/', express.static` call being "too early" in the chain. Static doesn't have a clean `next()` fallthrough so it does unexpected things. Can we try to keep the static bind just before the 404?

app.js
```diff
@@ -87,41 +89,51 @@ app.use(function(req, res, next) {
 });
 
 /**
+ * Is this a static asset?
+ */
+app.use('/', express.static(DIST_DIR));
```
Pomax: this relocation is what's causing the 303 error in app.test.js, because `express.static` does not have graceful `next()` fallthrough

package.json
mmmavis: kk, will do.

Pomax:
jar plus

mmmavis:
follow-up ticket for adding unit tests has been filed here: https://github.com/mozilla/teach.mozilla.org/issues/1730

Gonna merge this PR now! Yeah! :tada:

PR #1729, opened 2016-03-15
fix images on the community page getting horizontal scrollbars

ScottDowne:
fixes https://github.com/mozilla/teach.mozilla.org/issues/1731

### DIFF
--- a/less/components/illustration.less
+++ b/less/components/illustration.less
@@ -44,7 +44,7 @@
   }
   img {
     border-radius: @image-border-radius;
-    max-width: 100%;
+    width: 100%;
   }
   @media screen and (min-width: @screen-sm-min) {
     display: flex;


ScottDowne:
fixes #1731

less/pages/community.less
```diff
@@ -30,12 +30,20 @@
 }
 
 .communities-info {
-  display: flex;
```
Pomax: can you explain why this should no longer be a flexbox?

ScottDowne: It's inline-block, which also works.

This way it just automatically goes to the next line if the width becomes too small. Then `text-align: center;` centers any lone elements, when they become lone (also works if two are side by side).

flex is really good if the sizes of things are fairly dynamic, but this probably doesn't needs it, and this way, we don't need to use media queries for updating flex.

In this particular case we would have 4 media queries if we wanted to be thorough, didn't seem necessary to me.

ScottDowne:
Updated!

I had Ricardo take a design pass on this. A lot of my changes were me trying to not scale the images. I had it in my head that we didn't want to be doing that? Turns out it's fine.

Anyway, with scaling images, flexbox works well, so it's back.

We also discovered the scrollbars were only present in firefox, and fixed in dev edition. Making with width 100% fixed it for all browsers.

ScottDowne:
@mmmavis this has become a super simple r?

mmmavis:
:eyes: :+1: to me

PR #1744, opened 2016-03-22
Auto deploy to production once we merge to master

alicoding:


alicoding:
@cadecairos R?

alicoding:
I also removed all gulp s3 related files

package.json
```diff
@@ -19,21 +19,12 @@
     "app": "gulp app",
     "build": "gulp",
     "postinstall": "gulp postinstall",
-    "preapp": "rm -rf dist",
-    "prebuild": "rm -rf dist",
-    "pres3": "rm -rf dist",
-    "presmoketest": "rm -rf dist",
-    "prestart": "rm -rf dist",
-    "pretest": "rm -rf dist",
-    "pretravis-after-success": "rm -rf dist",
```
mmmavis: Some of the tasks here shouldn't be removed just yet since `npm start` is still serving files under `/dist`. i.e., `.html` files are still being generated: https://github.com/mozilla/teach.mozilla.org/blob/develop/lib/build/gulp-index-file-stream.js

alicoding: I don't see we use these tasks anywhere at all?

alicoding: Or wait... The pre is the magic?

alicoding: I see... I'll see which one I have to keep here...

alicoding:
@mmmavis I added those `pre` scripts back, but not all of them since `travis` pushes to heroku doesn't use `npm start` there and it will be a new build anyway.

alicoding:
@cadecairos R?

cadecairos:
@alicoding Is there a reason we can't use the deploy hooks that heroku provides in the app settings?

alicoding:
@cadecairos maybe I didn't know that option existed! Let me check that out

cadecairos:
R+

PR #1745, opened 2016-03-23
Tag and release v0.0.34

mmmavis:
### Changed
- The site now has a real server! Pages on https://teach.mozilla.org are 
  served by the server and  are no longer static files live on S3.
- Turned off auto push to S3 when the `develop` branch is updated.
- Updated Mouse logo.
- Removed Private Eye page.

### Added
- Added Community Call page.
- Added 404 messages server side.


mmmavis:
@hannahkane r?

hannahkane:
I don't see the 404 page—might be misunderstanding what "Added 404 messages both server side and client side" means.

If so, r+

mmmavis:
@hannahkane 

> I don't see the 404 page—might be misunderstanding what "Added 404 messages both server side and client side" means.

Had to disable the CMS hook so that changed the 404 handling a bit. Will file followup tickets.

PR #1748, opened 2016-03-23
Updated email validation regex

### LINKED ISSUE #1740: Update email validation regex with Exact Target's pattern
@stephaniemcv reported that our client side email validation for the newsletter signup form doesn't accept emails with plus sign (+) in them. We should probably update [the regex](https://github.com/mozilla/teach.mozilla.org/blob/develop/lib/util.js) so that it matches ET's pattern.

@stephaniemcv @alicoding do you by any chance know what pattern ET(or Basket) uses?

Note the client side validation is the fallback for browsers that don't support HTML5 form validation. For server side submission failure handling see https://github.com/mozilla/teach.mozilla.org/issues/1676

/cc @Pomax

mmmavis:
Fixes #1740

- updated email validation regex

---

@Pomax I would love to get some help from you to check if that regex looks sane... It thinks `11%@hi.com` is a valid email which seems odd to me

but 
```
curl --data "lang=pt-BR&newsletters=mozilla-learning-network&email=11%@hi.com" https://basket-dev.allizom.org/news/subscribe/
``` 
returns `{"status": "ok"}` so maybe `11%@hi.com` is really a valid email?

/cc @alicoding

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,7 +1,7 @@
 var Utils = {
   isValidEmail: function(email) {
     // regex copied from http://stackoverflow.com/a/46181
-    var regex = /^([\w-]+(?:\.[\w-]+)*)@((?:[\w-]+\.)*\w[\w-]{0,66})\.([a-z]{2,6}(?:\.[a-z]{2})?)$/i;
+    var regex = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
     return regex.test(email);
   },
   convertToRoute: function(str) {


alicoding:
I would take a cautious step here and encoded them before sending a raw data which these special characters over.

alicoding:
Also... When you curl with + does it work for you?

mmmavis:
@alicoding 

> Also... When you curl with + does it work for you?

curl with + doesn't work but I think `superagent` takes care of it when we `POST` to `basket`

In https://github.com/mozilla/teach.mozilla.org/blob/f9e0abfa1d78bef5556d61f57ff34a0f18ba9dc2/components/newsletter-signup/SignupForm.jsx#L37, `type` is set to `form`, which sets `Content-Type` to `application/x-www-form-urlencoded`, see https://visionmedia.github.io/superagent/#post-/-put-requests for deatils (search for `.type('form')`). I tried passing a hardcoded email that contains "+" in it and got `status` 200 back. :grin:

Pomax:
If we're changing this, we need this function unit tested, because just switching patterns doesn't tell us where it'll pass and where it'll fail.

We need a set of email addresses that should pass, and that shouldn't pass, and that this function runs against during `npm test`, so that we *know* the new code does what we need it to do.  Ideally, rather than rewriting our own pattern, we ditch our own regexp entirely, write out the representative list of email addresses that should work, and should not work, and then check if any of the email validation packages on npm do the right thing on our unit test. If any of them do, we can just use those without having to maintain our own regular expression.

Pomax:
Also, using any regular expression here is a theoretical security issue, due to [Regexp DOS](https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS), so not being responsible for that ourselves would be nice.

mmmavis:
Basket does email validation server side and will send back a non-200 status if anything fails. That's why we have this [ticket](https://github.com/mozilla/teach.mozilla.org/issues/1676) open. I don't know the exact validation rules Basket checks email against with (@alicoding do you know?) I can try adding some unit tests to cover a basic set of emails to make sure our client side email validation is sane.

alicoding:
I think if you use the built-in browser validation and if you get non-200
error Basket will tell you if error is because of missing some field or
email is invalid, so really doing extra check using regex might me
unnecessary.

mmmavis:
and I should revive this PR... 💀

mmmavis:
Closing this PR now as my other [PR](https://github.com/mozilla/learning.mozilla.org/pull/1911) covers the fix for this.

PR #1751, opened 2016-03-24
Weblitcircle

ScottDowne:


ScottDowne:
WIP Still need to drop in two lines of copy tomorrow.

app.js
```diff
@@ -102,7 +102,7 @@ app.use(function(req, res, next) {
       return next();
     }
     // this belongs to one of the predefined urls, let's generate its associated page
-    if ( urls.indexOf(location) != -1 ) {
+    if ( urls.indexOf(decodeURIComponent(location)) != -1 ) {
```
ScottDowne: I'm not so sure about this, but urls are written in decoded human readable strings, so it makes some sense to ensure we only compare against decoded human readable string too.

Pomax: "location" is not a URL, it's a React Router path (see line 96)

lib/routes.jsx
```diff
@@ -46,11 +46,33 @@ var pages = {
  */
 var redirects = {
   'clubs/curriculum': 'activities/web-lit-basics',
-  'teach-like-mozilla/web-literacy': 'activities/web-literacy'
+  'teach-like-mozilla/web-literacy': 'activities/web-literacy',
```
Pomax: might as well point this straight to `web-literacy` too now, no need for a double redirect

package.json
```diff
@@ -13,7 +13,7 @@
     "url": "https://github.com/mozilla/teach.mozilla.org/issues"
   },
   "engines": {
-    "node": ">=4.2.0"
+    "node": "^4.3.0"
```
Pomax: 4.4.1? If we're pegging against the LTS version of Node

pages/web-literacy-skills.jsx
```diff
@@ -0,0 +1,2 @@
+//TODO clean up indirection
+module.exports = require('./web-literacy/Skills.jsx');
```
Pomax: do we need this file? If this is a new file, we don't, whatever needs to load this page can just directly `require('pages/web-literacy/Skills.jsx');`

pages/web-literacy/MakeLinksFromWebLitSkills.jsx
```diff
@@ -0,0 +1,21 @@
+var React = require('react');
+var weblitcontent = require('./weblitcontent');
+var webLitSkillsContent = weblitcontent.webLitSkills;
+var ReactRouter = require('react-router');
+var Link = ReactRouter.Link;
+
+var MakeLinksFromWebLitSkills = function(webLitSkills) {
+  return webLitSkills.map(function(webLitSkill, index) {
+    return (
+      <span className="comma-separated-links" key={webLitSkill}>
```
Pomax: do we need the span? Can we not just set these props directly on the link itself?

pages/web-literacy/Skills.jsx
```diff
@@ -0,0 +1,88 @@
...
+          <p>{skill.content}</p>
+          <h3>Competencies</h3>
+          <div className="competency-content">
+            { this.generateCompetencies(skill.competencies) }
+          </div>
+          <h3>Web Literacy Skills</h3>
+          {
+            Object.keys(skill.topics).map(function(topic) {
+              return (
+                <div><b>{topic}:</b> {makeLinksFromWebLitSkills(skill.topics[topic])}</div>
```
Pomax: never use `<b>` or `<i>`. In HTML5, you want `<strong>` and `<em>`, things like "bold" and "italic" are CSS properties, not markup.

pages/web-literacy/Skills.jsx
```diff
@@ -0,0 +1,88 @@
...
+              );
+            })
+          }
+        </Illustration>
+      );
+    }.bind(this));
+  },
+  generateCompetencies: function(competencies) {
+    competencies = competencies || [];
```
Pomax: I'd shortcircuit this a little differently:
```
if (!competencies) return null;
// then the other return/map
```
Just to speed up the return

pages/web-literacy/Skills.jsx
```diff
@@ -0,0 +1,88 @@
...
+    return competencies.map(function(competency) {
+      if (competency.name) {
+        return (
+          <div key={competency.name}>
+            <h4>{competency.name}</h4>
+            {
+              competency.content.map(function(contentItem) {
```
Pomax: this should probably be its own function, too. Inlining lots of maps makes things harder to follow.

pages/web-literacy/Skills.jsx
```diff
@@ -0,0 +1,88 @@
...
+          alt={"21c-skills-icon-" + skill.name}>
+          <h2 id={skill.name}>{skill.name}</h2>
+          <p>{skill.content}</p>
+          <h3>Competencies</h3>
+          <div className="competency-content">
+            { this.generateCompetencies(skill.competencies) }
+          </div>
+          <h3>Web Literacy Skills</h3>
+          {
+            Object.keys(skill.topics).map(function(topic) {
```
Pomax: we probably want to give this its own function, just like generateSkills

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -1,44 +1,314 @@
...
+var categories = require('./categories');
+var weblitcolors = require('./colors');
+var topicContent = weblitcontent.topics;
+var webLitSkillsContent = weblitcontent.webLitSkills;
+var makeLinksFromWebLitSkills = require("./MakeLinksFromWebLitSkills.jsx");
+
+function makeLinksFrom21CSkills(skills21C) {
+  return skills21C.map(function(skill21C, index) {
+    return (
+      <span key={skill21C} className="comma-separated-links">
```
Pomax: do we need the span? It doesn't really do anything to the content (it just wraps inline content as inline content) and we can set those classes directly on the `<a>`

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -1,44 +1,314 @@
...
+        link={this.props.href}
+        externalLink={true}
+        alt={this.props.name + "-activity-icon"}>
+        <h2><a href={this.props.href}>{this.props.name}</a></h2>
+        <div className="meta-container">
+          <span className="meta-item"><i className="fa fa-users"></i>{this.props.difficulty}</span>
+          <span className="meta-item"><i className="fa fa-clock-o"></i>{this.props.duration}</span>
+        </div>
+        <p>{this.props.content}</p>
+        <div><b>Web Literacy Skills:</b> {makeLinksFromWebLitSkills(this.props.webLitSkills)}</div>
```
Pomax: `<strong>` (also next line)

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -1,44 +1,314 @@
...
-        <h1>Web Literacy</h1>
+  updateMapNavState: function() {
+    var topic = this.props.params.verb || "";
+    var webLitSkill = this.props.params.webLitSkill || "";
+    var state = {};
+    if (this.state.topic !== topic || this.state.webLitSkill !== webLitSkill) {
+      state.topic = topic;
+      state.webLitSkill = webLitSkill;
+      this.setState(state);
+    }
```
Pomax: Why are we not just doing this in `componentWillUpdate` or `componentDidUpdate`? Both give you the current props/state and next/previous (respectively) props/state pair to work with

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -1,44 +1,314 @@
...
-  render: function() {
-    var whitepaperLink = "https://mozilla.github.io/webmaker-whitepaper";
-    return (
-      <div className="inner-container">
-        <h1>Web Literacy</h1>
+  updateMapNavState: function() {
+    var topic = this.props.params.verb || "";
+    var webLitSkill = this.props.params.webLitSkill || "";
+    var state = {};
+    if (this.state.topic !== topic || this.state.webLitSkill !== webLitSkill) {
```
Pomax: it feels like this is not relying on React properly: we should be able to just say
```
setState({
  topic: this.props.params.verb,
  webLitSkills: this.props.params.webLitSkills
});
```
And if that doesn't change anything, it simply has no effect (react sees no state changes, and so no render/update is triggered)

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -1,44 +1,314 @@
...
   },
-  render: function() {
-    var whitepaperLink = "https://mozilla.github.io/webmaker-whitepaper";
-    return (
-      <div className="inner-container">
-        <h1>Web Literacy</h1>
+  updateMapNavState: function() {
+    var topic = this.props.params.verb || "";
+    var webLitSkill = this.props.params.webLitSkill || "";
+    var state = {};
```
Pomax: no need to allocate this until inside the conditional

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -1,44 +1,314 @@
...
+    if (this.state.topic !== topic || this.state.webLitSkill !== webLitSkill) {
+      state.topic = topic;
+      state.webLitSkill = webLitSkill;
+      this.setState(state);
```
Pomax: This can be just 
```
this.setState({
  topic: topic
  webLitSkill: webLitSkill
});
```

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -1,44 +1,314 @@
...
+  renderActivities: function() {
+    var activities = [];
+    activitydata.forEach(function(activity, index) {
+      if ((this.hasWebLitSkillIn(activity.webLitSkills) || this.hasMatchingWebLitSkillIn(activity.webLitSkills)) && this.hasMatching21CSkillIn(activity.skills)) {
+        activities.push(
+          <div className="activity-item" key={index}>
+            <Activity
+              selectedTopic={this.state.topic}
+              selectedWebLitSKill={this.state.webLitSkill}
+              topic={activity.topic}
```
Pomax: feels like we should do
```
activity.src1x = activity.imgSrc1x;
activity.src2x = activity.imgSrc2x;
...
<Activity selectedTopic={this.state.topic} selectedWebLitSKill={this.state.webLitSkill} {...activity}/>
```
where `{...activity}` is ES6 notation but is also part of JSX, so works even without ES6->ES5 conversions

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -1,44 +1,314 @@
...
+              href={activity.href}
+            />
+          </div>
+        );
+      }
+    }.bind(this));
+
+    if (!activities.length) {
+      return null;
+    } else {
```
Pomax: remember, you don't need an `else` keyword if your `if` case returns. If you're still in the function after the `if` block, you're automatically in the "else" case.

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -1,44 +1,314 @@
...
+                {value}
+              </p>
+            );
+          })
+        }
+      </div>
+    );
+  },
+  onMapToggle: function(labels) {
+    var verb =  labels[1] || "";
```
Pomax: I don't think we need the assignment fallback, do we? We're testing `if(thing)` before using them, so:
```
var url = "/web-literacy/"; // why is this hardcoded btw? This route might change, can we get it from somewhere?
var verb = labels[1];
if (verb) {
  ...
  var webLitSkill = labels[2];
  if (webLitSkills) {
    ...
  }
}
...
```

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -1,44 +1,314 @@
...
+    var webLitSkill = labels[2] || "";
+    var url = "/web-literacy/";
+    if (verb) {
+      url += verb + "/";
+      if (webLitSkill) {
+        url += webLitSkill + "/";
+      }
+    }
+    this.history.pushState(null, url);
```
Pomax: I thought we had access to `this.context.history`, which would allow use to not need a history mixin and would let us say
```
this.context.history.push({
  pathname: '/the/path',
  search: '?a=query',
  state: { the: 'state' }
}); 
```
(the history that React Router relies on should get autobound if you specify `contextTypes: { history: React.propTypes.object }` as part of the component definition)

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -1,44 +1,314 @@
...
+  },
+  skillCheckboxUpdated: function(checkbox, checked) {
+    var state = {
+      filter: this.state.filter
+    };
+    state.filter[checkbox] = !checked;
+    this.setState(state);
```
Pomax: This feels like a little too much unpacking
```
var filter = this.state.filter;
filter[checkbox] = !checked;
this.setState({ filter: filter });
```

pages/web-literacy/activitydata.js
```diff
@@ -0,0 +1,189 @@
+var search = "Search";
```
Pomax: it feels like these are constants, not variables; can we make them all caps to make their role clear?

pages/web-literacy/categories.js
```diff
@@ -0,0 +1,8 @@
+var categories = {
+	PS: "Problem-Solving",
```
Pomax: nit: that is a whole lot of whitespace. Can we turn that into just two spaces indentation?

pages/web-literacy/color.js
```diff
@@ -0,0 +1,33 @@
+var categories = require('./categories');
```
Pomax: Do we still rely on this? I don't think we do anymore given that we're now using CSS for the coloring. Can we remove this? (and do we need to remove mention of a colors object for the react-circletree package?)

pages/web-literacy/colors.js
```diff
@@ -0,0 +1,33 @@
+var categories = require('./categories');
```
Pomax: is github showing the same file twice? O_o

pages/web-literacy/skillsdata.js
```diff
@@ -0,0 +1,131 @@
+var search = "Search";
```
Pomax: Are these the same as in `pages/web-literacy/activitydata.js`? That needs some DRYing out

pages/web-literacy/weblitdata.js
```diff
@@ -0,0 +1,33 @@
+var categories = require('./categories');
+
+var PS = categories.PS;
+var Com = categories.Com;
+var Cre = categories.Cre;
+var Col = categories.Col;
+
+var weblitdata = {
+	"WEB LITERACY" : {
```
Pomax: funky indent

pages/web-literacy/weblitdata.js
```diff
@@ -0,0 +1,33 @@
+var categories = require('./categories');
```
Pomax: do we still use this file, or is it a left-over from the initial circletree testing?

pages/web-literacy/MakeLinksFromWebLitSkills.jsx
ScottDowne: I initially tried that, but the comma ends up as part of the link, because the pseudo element is inside the `<a>`

This span way, the pseudo element is outside of the `<a>` thus it gets the effect we want with links separate from the comma.

pages/web-literacy/WebLiteracyPage.jsx
ScottDowne: Same as above, we need the span so the comma is not inside the link.

pages/web-literacy/WebLiteracyPage.jsx
ScottDowne: Yeah, this is better.

I think at a point this was doing more, but was removed but this didn't get cleaned up.

pages/web-literacy/WebLiteracyPage.jsx
ScottDowne: Thanks this works and is probably cleaner.

pages/web-literacy/categories.js
ScottDowne: Heh, somehow a tab snuck in here.

pages/web-literacy/color.js
ScottDowne: Right now this file just seems to be storing key names to English words, so it makes it hard to type "Problem-Solving" in one spot and "problem solving" in another. Keeps it consistent. We jut interact with keys.

I don't mind keeping it, but it's really not required.

pages/web-literacy/colors.js
ScottDowne: These colour files are actually new to me, I have not touched them. I'll dig into them a bit see what they are doing and what we can remove.

pages/web-literacy/weblitdata.js
ScottDowne: More sneaky tabs.

pages/web-literacy/colors.js
ScottDowne: Yup, safe to remove.

pages/web-literacy/color.js
ScottDowne: Yeah, we probably don't need a colors object but I dunno if we want to remove it from circletree or not. Probably easier to always do it in css.

pages/web-literacy/MakeLinksFromWebLitSkills.jsx
Pomax: make the comma CSS?

pages/web-literacy/WebLiteracyPage.jsx
Pomax: as the comma is just decoration, comma through CSS?

mmmavis:
Seems like this PR has been reviewed and okayed by @Pomax . Travis is happy as well so I'm just gonna hit merge. :+1: :tada:

ScottDowne:
@mmmavis thanks!

Pomax:
note that there were some open questions on this PR

pages/web-literacy/MakeLinksFromWebLitSkills.jsx
ScottDowne: The comma is css already, thought, right?

It's in :after

ScottDowne:
If there was a communication error on this between @mmmavis and @Pomax it would be my fault.

ScottDowne:
@Pomax I think the only open question was on the comma, right?

I missed those questions and was focused on chat, and thought you said if I can get it green we can merge.

Pomax:
I'm not a fan of `<span>` elements, so if we can file a follow-up to clean those up and use CSS for the list item separation (which this is) then that's fine. As long as we don't leave pointless wrappers in the code.

ScottDowne:
@Pomax works for me

Issue #1757, opened 2016-03-24
Lettercasing in weblit map related urls needs adjustment


mmmavis:
We should probably turn them into all lower cases.

e.g.,
https://teach.mozilla.org/web-literacy/Read/Navigate/
should be
https://teach.mozilla.org/web-literacy/read/navigate/

/cc @sabrinang @Pomax

mmmavis:
Oh we should also replace space with hyphen as well

<img width="547" alt="screen shot 2016-03-24 at 5 10 06 pm" src="https://cloud.githubusercontent.com/assets/2896608/14034745/55837d42-f1e3-11e5-853a-e7b69692068b.png">

PR #1761, opened 2016-03-28
convert url values for weblit map to lowercase and replace spaces wit…

ScottDowne:
…h dashes

### DIFF
--- a/pages/web-literacy/WebLiteracyPage.jsx
+++ b/pages/web-literacy/WebLiteracyPage.jsx
@@ -17,12 +17,15 @@ var topicContent = weblitcontent.topics;
 var webLitSkillsContent = weblitcontent.webLitSkills;
 var makeLinksFromWebLitSkills = require("./MakeLinksFromWebLitSkills.jsx");
 var WhitePaperLink = require("./WhitePaperLink.jsx");
+var Util = require("../../lib/util");
+var convertToRoute = Util.convertToRoute;
+var parseRoute = Util.parseRoute;
 
 function makeLinksFrom21CSkills(skills21C) {
   return skills21C.map(function(skill21C, index) {
     return (
       <span key={skill21C} className="comma-separated-links">
-        <a href={"/web-literacy/skills/#" + categories[skill21C]}>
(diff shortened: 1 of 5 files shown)


pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -17,11 +17,23 @@ var topicContent = weblitcontent.topics;
...
+  str = str || "";
+  return str.toLowerCase().replace(" ", "-");
+}
+
+function parseRoute(route) {
+  route = route || "";
+  return route.split("-").map(function(item) {
+    return item.charAt(0).toUpperCase() + item.substr(1).toLowerCase();
+  }).join(" ");
+}
```
ScottDowne: I considered making these a module that we can require in, and I'm fine with doing that if we have strong opinions about it.

only the convertToRoute is used in another file, and seemed simple enough to duplicate it, but we can make a module for it if we want.

ScottDowne:
fixes #1757

ScottDowne:
@mmmavis r?

pages/web-literacy/WebLiteracyPage.jsx
mmmavis: maybe we can move them to `/lib/util.js` ?

mmmavis:
Looks great to me! Left one comment https://github.com/mozilla/teach.mozilla.org/pull/1761/files#r57613278. Feel free to hit merge once we've moved the helper functions to `lib/util.js` :smile:

pages/web-literacy/WebLiteracyPage.jsx
ScottDowne: I might update the names too. Outside of this file the context gets lost, so I should be more clear.

But yeah, moving them to utils is fine with me.

ScottDowne: Eh, keeping the names as is, should be fine.

ScottDowne:
@mmmavis updated. I would prefer if you took another quick look.

mmmavis:
looks great to me. merging it now! :+1:

PR #1768, opened 2016-03-29
fix up case of attributes

ScottDowne:


ScottDowne:
@Pomax yeah, was as simple as that.

Pomax:
y'arr plus

PR #1837, opened 2016-04-22
scroll to top for internal page nav links

Pomax:
fixes https://github.com/mozilla/teach.mozilla.org/issues/1812 by giving the link-anchor-swap an onClick handler that calls `window.scrollTo(0,0)`, so that page navigation also makes sure to load the new page with the scrollbar all the way at the top of the new content


mmmavis:
R+ !
And we should probably start converting all `<Link>` and `<OutboundLink>` to `<LinkAnchorSwap>`

alanmoo:
FYI: Looks like ReactRouter was updated to support a middleware layer, and react-router-scroll is a thing. See implementation of https://github.com/taion/react-router-scroll at https://github.com/mozilla/science.mozilla.org/pull/451

PR #1888, opened 2016-06-03
Add mozmaker css files

alicoding:


alicoding:
@gideonthomas R?

alanmoo:
Shouldn't we not be pulling in bootstrap then? And why is this from the web, not local?

alicoding:
@alanmoo bootstrap is there, and this is going to be just for testing until we work on `mozmaker`/`onesie` to make it compile all the css needed for this.

alanmoo:
I know bootstrap is there, but mozilla.github.io/mozmaker/dest/css/mozmaker.css looks to be a superset of bootstrap. So aren't we pulling in duplicate CSS?

mmmavis:
Question - Are we ready to pull in `mozmaker.css`? It will overwrite lots of existing stylings on the site.

cc @sabrinang @Pomax

mmmavis:
and cc @kristinashu

alanmoo:
Agreed, I'm not sure it's the right approach to globally pull it in. I'm thinking we put the `<link>` tag in the component that wraps wordpress content.

Pomax:
putting the link in sounds fine to me - even if we loaded multiple snips, we'd just be linking to the same stylesheet several times, right? (caching will make that a nonissue)

alicoding:
@alanmoo @mmmavis our bigger problem of all this is that this repo uses bootstrap v3 and onesie (mozmaker) is using bootstrap v4... @Pomax

Pomax:
hm, do we have a list of the concrete issues that results in?

alicoding:
Let's close this until we have re-write this site with mofo-bootstrap implemented.

PR #1968, opened 2016-07-14
Add localization support

### LINKED ISSUE #1929: Add react-intl to all files that are user-facing
Any files that we have user facing we'll need to localize those (error messages as well).

- [x] Add dependencies (react-intl etc.)
- [x] Add IntlProvider to client and server code
- [x] Create directory structure to support Pontoon locale files

alicoding:
Fixes #1929 

This is initial work for l10n, but we need more to be done in another PR...


components/page.jsx
```diff
@@ -115,10 +117,16 @@ var Page = React.createClass({
...
+    var currentLocale = 'en-US';
+    var messages = locales[currentLocale];
+    var supportedLocale;
 
     // but we _actually_ want to rely on this, instead:
     if (typeof window !== 'undefined') {
       currentPath = config.ORIGIN + window.location.pathname;
+      supportedLocale = Object.keys(locales).indexOf(navigator.language) !== -1;
+      currentLocale = supportedLocale ? navigator.language : currentLocale;
+      messages = locales[currentLocale] || messages;
```
alanmoo: Isn't this already the case from line 121?

lib/page-generate.jsx
```diff
@@ -40,6 +42,16 @@ function generateStaticRedirect(fromURL, toURL, next) {
   });
 }
 
+function createElement(Component, props) {
+  var messages = locales['en-US'];
+  // make sure you pass all the props in!
+  return (
+    <IntlProvider locale='en-US' messages={messages}>
```
alanmoo: This is hard coded for now because it's the only locale we have, right?

alicoding: LOL I just left a comment :P

pages/about.jsx
```diff
@@ -1,4 +1,4 @@
-var React = require('react');
+var React = require('react')
```
alanmoo: ಠ_ಠ

alicoding: LOOOOOOL

alicoding:
@Pomax I can't seem to figure out what is an actual error here...

components/page.jsx
gideonthomas: Can we add a comment explaining this if block

Pomax:
Hmm, I'm seeing `SyntaxError: Parse error` in the travis log before it errors out, so the easiest way to figure out what's going on is to run `npm start`, fire up the site on localhost and then go to [http://localhost:8008/tests](http://localhost:8008/tests) with the browser console open. 

If that works, we'll need to have a look at where in the phantom tests things break. Obviously before tests actually run, but it could be something like "trying to use a component with a missing require" or something, causing a syntax error involving an unassigned variable (that's a guess though, let's find out what it really is =D)

.travis.yml
```diff
@@ -1,3 +1,7 @@
 language: node_js
 node_js:
 - '4.4.7'
+sudo: false
+cache:
+  directories:
+  - node_modules
```
Pomax: are you sure we want this?

pages/about.jsx
```diff
@@ -7,79 +7,107 @@ var Illustration = require('../components/illustration.jsx');
...
-          We are cultivating a global network of web literacy leaders who are teaching and
-          learning the most important skills of our age: the ability to read, write and
-          participate in the digital world. Our programs and resources can help you make
-          an impact in your community, and participate in an openly networked world.
-        </p>
-      </Illustration>
-    </section>
-  </div>
-);
+var Intro = React.createClass({
```
Pomax: is this prep work, or done? I don't see any localization happening here =D

pages/about.jsx
```diff
@@ -7,79 +7,107 @@ var Illustration = require('../components/illustration.jsx');
...
-          your own resources to receive feedback from peers.
-        </li>
-        <li>Develop your skills in facilitative leadership, participatory practices and collaboration.</li>
-      </ul>
-    </Illustration>
-  </section>
-);
+var OurFocus = React.createClass({
+  contextTypes: {
+    intl: React.PropTypes.object
```
Pomax: is this used atm or prep?

pages/about.jsx
```diff
@@ -7,79 +7,107 @@ var Illustration = require('../components/illustration.jsx');
...
-    <p className="learn-more">
-      In 2015, our goal is to foster and sustain web literacy activities in 500 cities around the globe.
-      Join us by teaching, learning, sharing and imagining the full power of the web.
-    </p>
-    <div className="horizontal-divider"></div>
-  </section>
-);
+var OurGoal = React.createClass({
+  contextTypes: {
+    intl: React.PropTypes.object
```
Pomax: is this used atm or prep?

pages/about.jsx
```diff
@@ -7,79 +7,107 @@ var Illustration = require('../components/illustration.jsx');
...
-      imgSrc="/img/pages/about/svg/icon-connect-blue.svg"
-      head="Say Hello"
-      subhead="Connect on the Discourse forum."
-      highlightedText="Discourse forum"
-    />
-  </IconLinks>
-);
+var IconLinksSets = React.createClass({
+  contextTypes: {
+    intl: React.PropTypes.object
```
Pomax: is this used atm or prep?

pages/home/HomePage.jsx
```diff
@@ -59,7 +60,7 @@ var HomePage = React.createClass({
     return (
       <div>
         <HeroUnit>
-          <h1>The Mozilla Learning Network</h1>
+          <h1>{this.context.intl.formatMessage({id: 'mln'})}</h1>
```
Pomax: I'd advocate clear, human-readable ids. As an acronym can we use "MLN"? (case matters in this instance)

pages/about.jsx
```diff
@@ -7,79 +7,107 @@ var Illustration = require('../components/illustration.jsx');
...
-        </p>
-      </Illustration>
-    </section>
-  </div>
-);
+var Intro = React.createClass({
+  contextTypes: {
+    intl: React.PropTypes.object
+  },
+  render() {
```
Pomax: can't do ES6, this codebase is stuck in ES5 land, so `rnder: function()` is still necessary

pages/about.jsx
```diff
@@ -7,79 +7,107 @@ var Illustration = require('../components/illustration.jsx');
...
-        <li>Develop your skills in facilitative leadership, participatory practices and collaboration.</li>
-      </ul>
-    </Illustration>
-  </section>
-);
+var OurFocus = React.createClass({
+  contextTypes: {
+    intl: React.PropTypes.object
+  },
+  render() {
```
Pomax: `render: function() {`

pages/about.jsx
```diff
@@ -7,79 +7,107 @@ var Illustration = require('../components/illustration.jsx');
...
-      Join us by teaching, learning, sharing and imagining the full power of the web.
-    </p>
-    <div className="horizontal-divider"></div>
-  </section>
-);
+var OurGoal = React.createClass({
+  contextTypes: {
+    intl: React.PropTypes.object
+  },
+  render() {
```
Pomax: `render: function() {`

pages/about.jsx
```diff
@@ -7,79 +7,107 @@ var Illustration = require('../components/illustration.jsx');
...
-      subhead="Connect on the Discourse forum."
-      highlightedText="Discourse forum"
-    />
-  </IconLinks>
-);
+var IconLinksSets = React.createClass({
+  contextTypes: {
+    intl: React.PropTypes.object
+  },
+  render() {
```
Pomax: `render: function() {`

Pomax:
The test problem was that you had some ES6 syntax for the `render` function changes you made, which had to be ES5 instead. Browsers wouldn't break on it because FF and Chrome both support most of ES6 by now, but node/phantom don't and so it would error out.

I changed `render() { ... }` back to `render: function() { ... }` and npm test made it through to the end, although with errors about React.Intl not being loaded properly. Which is good! It means tests run and is giving useful feedback on what needs to still be added.

.travis.yml
alicoding: This kinda speed up travis build like a lot

pages/about.jsx
alicoding: I don't want to have over 1000 lines changes in this patch, so follow up coming soon :)

pages/about.jsx
alicoding: Not at the moment, but I was testing to see if this will work after my integration.

pages/about.jsx
alicoding: prep

pages/about.jsx
alicoding: prep

pages/home/HomePage.jsx
alicoding: Will change this :)

alicoding:
@Pomax this is how I fixed it since `mocha-phantomjs` is not going to get that updated anytime soon. https://github.com/alicoding/learning.mozilla.org/blob/90728f5c8ebab94f4dcb27f6ef87033f3c33f05b/test/browser/main.js#L5-L15

alicoding:
@Pomax can you prioritize this PR? thanks!

pages/about.jsx
Pomax: might be worth removing if we want to land this PR, and then add it in, in follow-up

test/browser/main.js
```diff
@@ -1,4 +1,19 @@
 var originalConsoleWarn = console.warn;
+// Version of PhantomJS that's being used by `mocha-phantomjs` does not enabled `intl`
+// which causes most tests to fail because it relies on `Intl` to be in `Window/Global`
+// This can be removed once the library is updated to use latest version that supports it.
+var areIntlLocalesSupported = require('intl-locales-supported');
+if (global.Intl) {
+    if (!areIntlLocalesSupported(['en-US'])) {
```
Pomax: I assume this means if `areIntlLocalesSupported` tests true, we're all good and no shims are necessary

alanmoo: Yeah this variable name is confusing

package.json
```diff
@@ -19,11 +19,11 @@
     "app": "gulp app",
     "build": "gulp",
     "l10n": "node scripts/properties2json",
-    "preapp": "rm -rf dist",
+    "preapp": "rm -rf dist && npm run l10n",
```
Pomax: we can probably compact this, with a task like:
```
...
"cleanup": "rm -rf dist && npm run l10n",
"preapp": "npm run cleanup",
"presmoketest": "npm run cleanup",
"pretest": "npm run cleanup",
...
```

Pomax:
I don't see any errors on `npm test` anymore, and the site works just fine, so... R+!

(we can do a tiny bit of compacting in the package.json, but it's not strictly related to the changes in this PR)

test/browser/main.js
alicoding: it's a copy paste from `intl` library and its only for test and it's well commented, so I won't change that unless you think it's critically important to.

PR #1972, opened 2016-07-18
Adding CSP rules to allow Pontoon to talk to the app

### LINKED ISSUE #1954: Add CSP rules for pontoon to communicate with the app for iframe related
We will have to allow Pontoon to communicate with our server for front-end localization.

alicoding:
Closes #1954

@cadecairos R?

### DIFF
--- a/app.js
+++ b/app.js
@@ -88,7 +88,8 @@ app.use(helmet.contentSecurityPolicy({
   directives: {
     defaultSrc: [
       'www.youtube.com',
-      'https://public.etherpad-mozilla.org'
+      'https://public.etherpad-mozilla.org',
+      'https://pontoon.mozilla.org'
     ],
     scriptSrc: [
       '\'self\'',
@@ -99,12 +100,14 @@ app.use(helmet.contentSecurityPolicy({
       'cdn.optimizely.com',
       'https://www.google.com',
       'https://s.ytimg.com',
-      'https://www.mozilla.org'
+      'https://www.mozilla.org',
+      'https://pontoon.mozilla.org'
     ],
     fontSrc: [
       '\'self\'',
       'fonts.googleapis.com',
-      'fonts.gstatic.com'
+      'fonts.gstatic.com',
+      'https://pontoon.mozilla.org'
     ],
     styleSrc: [
       '\'self\'',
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -88,7 +88,8 @@ app.use(helmet.contentSecurityPolicy({
   directives: {
     defaultSrc: [
       'www.youtube.com',
-      'https://public.etherpad-mozilla.org'
+      'https://public.etherpad-mozilla.org',
+      'https://pontoon.mozilla.org'
```
cadecairos: Why'd you add it to every directive?

Surely pontoon doesn't need to be allowed by default.

app.js
```diff
@@ -99,20 +100,23 @@ app.use(helmet.contentSecurityPolicy({
...
-      'https://www.mozilla.org'
+      'https://www.mozilla.org',
+      'https://pontoon.mozilla.org'
     ],
     fontSrc: [
       '\'self\'',
       'fonts.googleapis.com',
-      'fonts.gstatic.com'
+      'fonts.gstatic.com',
+      'https://pontoon.mozilla.org'
```
cadecairos: do we load fonts from pontoon?

app.js
```diff
@@ -99,20 +100,23 @@ app.use(helmet.contentSecurityPolicy({
...
     ],
     styleSrc: [
       '\'self\'',
       '\'unsafe-inline\'',
       'https://www.google.com',
       'fonts.googleapis.com',
       'https://api.tiles.mapbox.com',
-      'https://s.ytimg.com'
+      'https://s.ytimg.com',
+      'https://pontoon.mozilla.org'
```
cadecairos: and style?

app.js
```diff
@@ -127,6 +131,7 @@ app.use(helmet.contentSecurityPolicy({
       '*.mywebmaker.org',
       '*.makes.org',
       'bitly.mofoprod.net',
+      'https://pontoon.mozilla.org',
```
cadecairos: and images?

app.js
alicoding: Yup we do

app.js
alicoding: Yup

app.js
alicoding: Yup

app.js
alicoding: Pontoon needs everything in order to function properly...

cadecairos: wat.

alicoding: Well... Its in an iframe and trying to load everything for users to translate the site and it needs to display everything correctly the same way on production.

cadecairos:
R+

PR #1978, opened 2016-07-19
Fix build error for postinstall - related to l10n

alicoding:
I broke the build after introducing l10n here, but this is what causes the issue. 

```
Error: ./lib/page-generate.jsx
       Module not found: Error: Cannot resolve 'file' or 'directory' ../dist/locales.json in /tmp/build_d6aca50ad29e8ab3a65c281def8a57f2/mozilla-learning.mozilla.org-229092d/lib
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "presmoketest": "npm run cleanup",
     "prestart": "rm -rf dist",
     "pretest": "npm run cleanup",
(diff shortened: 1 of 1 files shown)


alicoding:
@Pomax R?

Pomax:
hahaha R+

PR #1980, opened 2016-07-19
npm run l10n should run first for prod

alicoding:
I'm sorry, but for production it needs to be called first :(


Pomax:
:shipit:

PR #1985, opened 2016-07-21
Localize sidebar component

alicoding:
@alanmoo R?


locales/en-US/messages.properties
```diff
@@ -1 +1,15 @@
 MLN=The Mozilla Learning Network
+teaching_activities=Teaching ActivitiesL10N
```
alanmoo: ^

alanmoo:
Just fix that one string

locales/en-US/messages.properties
alicoding: I porgottttr

alicoding:
@alanmoo done

PR #1983, opened 2016-07-20
Locale negotiation

alanmoo:
Fixes #1958

**Changes proposed in this pull request:**
- Automatically redirects users to appropriate locale (via mofo-localize module)
- Update routes to support locales
- Update links to point to current locale url

I believe I've gotten everything, but it's hard to test every link on the site

PR #1986, opened 2016-07-21
Fix react client/server mismatch warning

alanmoo:
#1983 introduced a bug with client/server rendering. Fixing that.


alicoding:
r+

PR #1987, opened 2016-07-21
Localize login component

alicoding:
@alanmoo R?

### DIFF
--- a/components/login/Login.jsx
+++ b/components/login/Login.jsx
@@ -4,6 +4,8 @@ var Link = require('react-router').Link;
 var ga = require('react-ga');
 var OutboundLink = require('react-ga').OutboundLink;
 var PureRenderMixin = require('react-addons-pure-render-mixin');
+var FormattedMessage = require('react-intl').FormattedMessage;
+var FormattedHTMLMessage = require('react-intl').FormattedHTMLMessage;
 
 var withTeachAPI = require('../../hoc/with-teach-api.jsx');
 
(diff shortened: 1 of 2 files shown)


components/login/Login.jsx
```diff
@@ -107,32 +109,30 @@ var Login = React.createClass({
           <p>
             <small>
               <span className="fa fa-wrench"/>
-              Unable to contact login server.
-              <br/>
-              Refresh the page to try again.
+              {this.context.intl.formatMessage({id: 'unable_to_contact_login_server'})}
```
alanmoo: the `<br/>` renders as text

locales/en-US/messages.properties
```diff
@@ -1 +1,9 @@
+unable_to_contact_login_server=Unable to contact login server.<br/> Refresh the page to try again.
+loading_message=Loading&hellip;
+your_project=Your Projects
```
alanmoo: Add an s for future clarity?

locales/en-US/messages.properties
```diff
@@ -1 +1,9 @@
...
+log_out=Log Out
+sign_in=Sign in
+or=or
+sign_up=Sign Up
+hi_username=Hi, {name}
```
alanmoo: Call this "user_greeting" instead?

alicoding: I think either `hi_name` or `hi_username` would make the most sense. The pattern we're doing is string+var

alicoding:
fixed @alanmoo

locales/en-US/messages.properties
alanmoo: In English- isn't it possible that other languages the appropriate greeting starts with the username? Remember, this is localization, not just translation. We're the only people that see this (translators don't, right?), but I still think it helps to be conscious of this so that we can break strings up appropriately based not just on English structure but semantics of the string. You don't need to change this specific example, but keep it in mind when pontoon-ifying future pages.

PR #1994, opened 2016-07-22
Add another domain to CSP for Pontoon

alicoding:


cadecairos:
R+ Once you switch to dot notation for the directives

PR #2008, opened 2016-07-25
Localize more components

alicoding:
@alanmoo R?


components/not-found.jsx
```diff
@@ -9,9 +9,9 @@ var NotFoundMessage = function(props) {
                 height='auto'
                 src1x='/img/pages/not-found/book_singlepageflip.gif'
                 alt='' />
-      <h2>Hey, this 404 is a teachable moment!</h2>
-      <p>Did you know that a 404 is the generic error code used across the internet to mean "page not found"? It might mean the page used to exist, but doesn't anymore. Or that someone made a typo somewhere. Regardless, there's no page at this address.</p>
-      <Link to={"/" + this.context.intl.locale + "/"} className="btn">Go To Home Page</Link>
+      <h2>{this.context.formatMessage({id: 'hey_404'})}</h2>
```
Pomax: these two are probably better off with generic IDs around 404 data, like `404_header` and `404_message`, instead of the first (few) words of the content that's getting localized. Especially if the phrasing can change later in the life of the project (which is not unreasonable to expect)

alanmoo: +1 to that (I was checking with Pomax if that seemed reasonable or not, he just beat me to the comment 😛 )

alicoding:
@alanmoo @Pomax fixed

alanmoo:
Looks good once Travis confirms, though interesting to note the 404 page doesn't actually show up when looking at an invalid URL (though it's like this on production too).

alicoding:
@alanmoo yeah the page is not actually being rendered with that component, yet...

PR #2016, opened 2016-07-26
Removing X-Frame-Options

alicoding:
Since we already added `frame-ancestor` in CSP, we don't need this no more!


alicoding:
@cadecairos R?

cadecairos:
R+ LGTM

PR #2018, opened 2016-07-26
login emulation the teach-api connector

### LINKED ISSUE #2017: add login emulation to teach-api.js
technically already implemented in the clubform branch, but this should be landed separaterly really.

Pomax:
fixes #2017

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,9 @@ software works.
 `WORDPRESS_DOMAIN` | The domain to the wordpress.com site this app retrieves data from. e.g., `helloworld.wordpress.com` (with no protocol),
 `HPKP` | [HTTP Public Key Pinning](https://developer.mozilla.org/en/docs/Web/Security/Public_Key_Pinning) - requires at least two sha256s, space separated. i.e "asdf= qwer="
 `supportedLocales` | enables specific locales, for example: `['en-US', 'id']` will only enable 2 locales. Default value is `*` which will enable all locales under [locales](./locales) directory.
+`LOGIN_EMULATION` | determines whether or not the teach-api is used for login state management. defaults to false.
+`LOGIN_EMULATION_LOGGEDIN` | determines whether, when login emulation is on, the login state should be logged in or out. defaults to false.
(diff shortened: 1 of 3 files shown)


alanmoo:
What am I looking for when testing this? I change the .env variables but see no change on the homepage of the learning site.

alanmoo:
^ @Pomax

Pomax:
@alanmoo with this branch checked out, create an `.env` file and set the login emulation values as indicated in the README.md, then without a teach-api or id etc running see if the site's sign in button logs in you, and sign out signs you out, without any errors around "there is no teach-api halp halp".

alanmoo:
I'm just constantly getting `Unable to contact login server.
Refresh the page to try again.` in the sign-in area.

Current .env:
````
ENABLE_BADGES="yes please, that would be great"
LOGIN_EMULATION=0
````

Pomax:
I've updated the PR - turns out I had forgotten to add the new environment variables to the Webpack config, so they only worked on the server, not on the client after bundling. Yay! Want to re-review?

alanmoo:
If `LOGIN_EMULATION` is on and `LOGIN_EMULATION_LOGGEDIN` is off, I'm not sure I understand the difference in what's happening- the "sign in" link directs to the same spot regardless of `LOGIN_EMULATION`. (the state query param is different but I'm not clear on what that is).

@Pomax can you explain?

Pomax:
With login emulated I needed a way to be both logged in or logged out, without hitting the teach-api, but I think in this case the default should be "logged in" unless explicitly logged out. Let me reverse that var (and its name)

alanmoo:
But with login emulated you can't change state from the front end, right?

Pomax:
no, but you can determine the login state without needing to hit the teach-api for dev purposes, so you can bootstrap a logged in or logged out state, and then `npm start`, without most of the code not running because the teach-api connector crashes out.

I could probably rewrite this so the sign in/out buttons work more like a mocked teach-api so you can click sign in and sign out rather than needing to bootstrap

alanmoo:
👍

PR #2019, opened 2016-07-26
Fix a bunch of linking issues for locales

### LINKED ISSUE #2006: Certain links fail as a result of locale negotiation
React-router is having a hard time handling some links on https://teach.mofostaging.net/en-US/web-literacy/skills

alanmoo:
Fixes #2006

- Add localization to links for web-literacy skills
- Make badges work
- Add a l10n string that was missing

@pomax can you also take a look over this, specifically `routes.jsx`? I removed the redirects from the one array because they don't seem to specifically be necessary- they get passed as a separate object and redirects work without that line.

### DIFF
--- a/components/sidebar/topLevelNavItems.jsx
+++ b/components/sidebar/topLevelNavItems.jsx
@@ -33,7 +33,7 @@ var MENU_ENTRIES = [
         to: '/clubs'
       },
       {
-        name: "Guides & Resources",
+        name: "guides_and_resources",
         to: '/clubs/guides'
       },
       {
--- a/pages/badges/badge-single.jsx
+++ b/pages/badges/badge-single.jsx
@@ -16,16 +16,19 @@ var React = require('react'),
     Modal = require('../../components/modal.jsx');
 
 var Navigation = React.createClass({
+  contextTypes: {
+    intl: React.PropTypes.object
+  },
   render: function() {
     var prev = this.props.prev;
     var next = this.props.next;
     return (
       <div className="badge-navigation">
-        <a className="previous" href={prev.url}>
+        <a className="previous" href={"/" + this.context.intl.locale + prev.url}>
           <img src={prev.img} />
           <span clasName="label">← {prev.title}</span>
         </a>
-        <a className="next" href={next.url}>
+        <a className="next" href={"/" + this.context.intl.locale + next.url}>
           <span clasName="label">{next.title} →</span>
           <img src={next.img} />
         </a>
@@ -44,7 +47,8 @@ var BadgePage = React.createClass({
   },
 
   contextTypes: {
-    history: React.PropTypes.object
+    history: React.PropTypes.object,
+    intl: React.PropTypes.object
   },
 
   getInitialState: function () {
@@ -169,7 +173,7 @@ var BadgePage = React.createClass({
 
     return (
       <div className="individual-badge">
-        <div> <a href="/badges">← back to credentials</a> </div>
+        <div> <a href={"/" + this.context.intl.locale + "/badges"}>← back to credentials</a> </div>
         <BadgeHorizontalIcon badge={this.state.badge} />
         { content }
         <Divider />
--- a/pages/badges/badges.jsx
+++ b/pages/badges/badges.jsx
@@ -155,6 +155,7 @@ var BadgesPage = React.createClass({
   },
 
   generateBadgeList: function() {
+    var that = this;
     var anonymous = !this.state.teachAPI.getLoginInfo();
 
     if (this.state.badges.length === 0) {
@@ -170,7 +171,7 @@ var BadgesPage = React.createClass({
       var linkUrl = '/badge/' + badge.id + '/' + urlize(badge.title);
       return (
         <div key={badge.id} className="col-md-3">
-          <Link to={ "/" + this.context.intl.locale + linkUrl } className={'badge-icon-link'}>
+          <Link to={ "/" + that.context.intl.locale + linkUrl } className={'badge-icon-link'}>
             <BadgeVerticalIcon badge={badge} anonymous={anonymous} />
           </Link>
         </div>
--- a/pages/web-literacy/MakeLinksFromWebLitSkills.jsx
+++ b/pages/web-literacy/MakeLinksFromWebLitSkills.jsx
@@ -5,12 +5,12 @@ var ReactRouter = require('react-router');
 var Link = ReactRouter.Link;
 var convertToRoute = require("../../lib/util").convertToRoute;
 
-var MakeLinksFromWebLitSkills = function(webLitSkills) {
+var MakeLinksFromWebLitSkills = function(webLitSkills, locale) {
   return webLitSkills.map(function(webLitSkill, index) {
     return (
       <span className="comma-separated-links" key={webLitSkill}>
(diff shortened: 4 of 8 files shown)


alicoding:
LGTM

components/sidebar/topLevelNavItems.jsx
```diff
@@ -33,7 +33,7 @@ var MENU_ENTRIES = [
         to: '/clubs'
       },
       {
-        name: "Guides & Resources",
+        name: "guides_and_resources",
```
Pomax: in the interested of dev sense, should these become `locale_id` rather than `name``?

pages/badges/badge-single.jsx
```diff
@@ -16,16 +16,19 @@ var React = require('react'),
...
+  contextTypes: {
+    intl: React.PropTypes.object
+  },
   render: function() {
     var prev = this.props.prev;
     var next = this.props.next;
     return (
       <div className="badge-navigation">
-        <a className="previous" href={prev.url}>
+        <a className="previous" href={"/" + this.context.intl.locale + prev.url}>
```
Pomax: Reading this code makes it feels like it would be very sensible to capture this value as a `this.locale`, or better yet, since `locale` doesn't trickle down but is static for an entire site, having a lib that can act as central locale consultation point... that may be out of scope for this PR, but having to get this value out of a `this.context.intl` namespace seems too elaborate

components/sidebar/topLevelNavItems.jsx
alanmoo: This has already been merged in from another PR, it was a random warning that I noticed while writing this patch.

pages/badges/badge-single.jsx
```diff
@@ -16,16 +16,19 @@ var React = require('react'),
...
           <img src={prev.img} />
           <span clasName="label">← {prev.title}</span>
         </a>
-        <a className="next" href={next.url}>
+        <a className="next" href={"/" + this.context.intl.locale + next.url}>
```
Pomax: it also feels like this is unnecessarily not-DRY: we can easily write a component `<LocaleLink>` such that we can simply pass in the urls you see in routes.jsx, without having to worry about whether you forgot the local prefix or not. If that is out of scope for this PR: that absolutely should be filed as followup so we're not introducing a new way in which to break the code with a simple typo or forgetting about localization.

pages/badges/badges.jsx
```diff
@@ -155,6 +155,7 @@ var BadgesPage = React.createClass({
   },
 
   generateBadgeList: function() {
+    var that = this;
```
Pomax: Hmm, this should not be necessary. If we need to refer to the instance, then the function later on should love on the React component and passed into the handler as `this.[functionname]`, rather than aliassing 'that' (since React does autorebinding for component functions)

pages/badges/badge-single.jsx
alanmoo: I see where you're coming from, but that does require remembering to use the appropriate link element, so it's not a foolproof win.

pages/web-literacy/MakeLinksFromWebLitSkills.jsx
```diff
@@ -5,12 +5,12 @@ var ReactRouter = require('react-router');
...
 var convertToRoute = require("../../lib/util").convertToRoute;
 
-var MakeLinksFromWebLitSkills = function(webLitSkills) {
+var MakeLinksFromWebLitSkills = function(webLitSkills, locale) {
   return webLitSkills.map(function(webLitSkill, index) {
     return (
       <span className="comma-separated-links" key={webLitSkill}>
         <Link
-          to={"/web-literacy/" + convertToRoute(webLitSkillsContent[webLitSkill].topic) + "/" + convertToRoute(webLitSkill) + "/"}
+          to={"/" + locale + "/web-literacy/" + convertToRoute(webLitSkillsContent[webLitSkill].topic) + "/" + convertToRoute(webLitSkill) + "/"}
```
Pomax: This, too, feels like we should make a `<LocalLink>`-like component that takes a regular `to` and then rewrites it internally, putting the locale code in without needing a dev to type it anywhere.

pages/web-literacy/Skills.jsx
```diff
@@ -6,14 +6,18 @@ var makeLinksFromWebLitSkills = require("./MakeLinksFromWebLitSkills.jsx");
...
 var Skills21CPage = React.createClass({
+  contextTypes: {
+    intl: React.PropTypes.object
+  },
   statics: {
     pageTitle: '21st Century Skills',
     pageClassName: 'web-literacy-skills-page'
   },
   generateSkillItem: function(skill) {
+    var that = this;
```
Pomax: this should not be necessary, whatever ends up using `that` should be a component function, with the `this` binding left to React.

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -21,11 +21,11 @@ var Util = require("../../lib/util");
 var convertToRoute = Util.convertToRoute;
 var parseRoute = Util.parseRoute;
 
-function makeLinksFrom21CSkills(skills21C) {
+function makeLinksFrom21CSkills(skills21C, locale) {
   return skills21C.map(function(skill21C, index) {
     return (
       <span key={skill21C} className="comma-separated-links">
-        <a href={"/web-literacy/skills/#" + categories[skill21C].toLowerCase()}>
+        <a href={"/" + locale + "/web-literacy/skills/#" + categories[skill21C].toLowerCase()}>
```
Pomax: see previous note on hardening code against fallible human tendencies =)

pages/badges/badges.jsx
alanmoo: Even considering the map below? (line 174)

pages/web-literacy/WebLiteracyPage.jsx
```diff
@@ -184,6 +187,7 @@ var WebLitPage = React.createClass({
     }.bind(this));
   },
   renderTopics: function() {
+    var that = this;
```
Pomax: see previous notes on React autobinding `this` for component functions

Pomax:
Hmm, I can't seem to run this PR. If I check it out, run `npm i`, then `npm start`, and try `http://localhost:8008` in any browser, both plain and private browsing mode, I get a 307 chain that resolves some 20-odd times and then gives up

components/sidebar/topLevelNavItems.jsx
Pomax: Yeah, I don't think I was on that PR, hence the comment now, really =/

pages/badges/badge-single.jsx
Pomax: it makes it very easy to check files for use of raw `<a>` though, which is easily automated.

pages/badges/badges.jsx
Pomax: yeah, at this point it means that map function should be refactor out of the current code into its own component function instead.

Pomax:
This looks good on mac - I'll file the windows problems a separate issue.

alanmoo:
Merging this in so staging can be un-broken, but filing follow up tickets for some linkage/intl context cleanups.

PR #2023, opened 2016-07-27
Fix l10n warning when running tests

alicoding:
Currently when run `npm test` there are bunch of warning something like:

```
[React Intl] Cannot format message: "learn_more", using message id as fallback.
[React Intl] Missing message: "follow_us" for locale: "en"
```

This is because we haven't pass the messages object to the `IntlProvider` for test runner.


alicoding:
@alanmoo R?

PR #2026, opened 2016-07-27
fixes the prop2json script by making a path lookup cross-platform

Pomax:
fixes https://github.com/mozilla/learning.mozilla.org/issues/2025 by replacing a `'/'` in a path, which only works on Unixy systems, with `path.sep`, which works on all operating systems.


alicoding:
r+

I've tested it on my Mac machine and this worked the way we expected it to be!

alicoding:
I'm cool with that check +1

Pomax:
shadoobi - thanks for the review!

PR #2033, opened 2016-07-28
Remove old component for weblit

alicoding:
These files are not being used anywhere at all, so I thought removing them to avoid confusion is the best idea here.


alicoding:
@Pomax I forgot to mention your name here...

Pomax:
R+ if you also remove the `expander.less` file, and its entry in the LESS `index.js`?

alicoding:
thanks @Pomax <3

PR #2037, opened 2016-07-29
Localize /me page and update default ENV

alicoding:
This PR cover 2 things:

1. Localized /me page
2. Update ENV to point to our new staging URL

I have to do number 2 here to test my work and felt it's not a bad idea to do it in this PR
if anyone who's going to review this patch want to test it will be able to do so.


locales/en-US/messages.properties
```diff
@@ -95,3 +96,8 @@ terms=Terms
 privacy=Privacy
 privacy_policy=privacy policy
 privacy_consent=I'm okay with you handling this info as you explain in your {privacy_policy_link}
+updated_relative_time=Updated {relativeTime}
+loadig_projects=Loading projects
```
alanmoo: typo

alanmoo:
I'm also getting an error

[Error] Refused to connect to 'https://alanmoo.makes.org/makes.json' because it violates the following Content Security Policy directive: "connect-src 'self' https://www.google.com *.tiles.mapbox.com *.log.optimizely.com *.github.io *.mywebmaker.org *.makes.org bitly.mofoprod.net https://teach-api-staging.herokuapp.com basket-dev.allizom.org".

Followed by

SecurityError: DOM Exception 18: An attempt was made to break through the security policy of the user agent.

alicoding:
@alanmoo I think the error it causes because the staging site is not being configured to allow the request for localhost

PR #2043, opened 2016-08-02
Add the CSRF and withCredentials code back into the teach-api

Pomax:
For unclear reasons the `teach-api.js` request function had lost the code that determines whether or not to supply the teach-api with credentials and the required Django CSRF token.

This PR restores that.

### DIFF
--- a/lib/teach-api.js
+++ b/lib/teach-api.js
@@ -111,6 +111,17 @@ _.extend(TeachAPI.prototype, {
                      url + '. Not sending auth token.');
       }
     }
+ 
+    // do we need credentials + Django CSRF token?
+    if (['post','put','delete'].indexOf(method) > -1) {
+      var csrf = document.cookie.match(/csrftoken=\S+/);
+      if (csrf) {
+        csrf = csrf[0].replace(';','').split('=');
+        var token = csrf[1];
+        req.set('X-CSRFtoken', token ? token : "FALSE");
+        req.withCredentials();
+      }
+    }    
 
     return req;
   },


lib/teach-api.js
```diff
@@ -111,6 +111,17 @@ _.extend(TeachAPI.prototype, {
...
       }
     }
+ 
+    // do we need credentials + Django CSRF token?
+    if (['post','put','delete'].indexOf(method) > -1) {
+      var csrf = document.cookie.match(/csrftoken=\S+/);
+      if (csrf) {
+        csrf = csrf[0].replace(';','').split('=');
+        var token = csrf[1];
+        req.set('X-CSRFtoken', token ? token : "FALSE");
```
gideonthomas: isn't it `X-CSRF-Token`?

gideonthomas:
r+ with one question

lib/teach-api.js
Pomax: I went on the example code in https://docs.djangoproject.com/en/1.9/ref/csrf, which gives:
```
function csrfSafeMethod(method) {
    // these HTTP methods do not require CSRF protection
    return (/^(GET|HEAD|OPTIONS|TRACE)$/.test(method));
}
$.ajaxSetup({
    beforeSend: function(xhr, settings) {
        if (!csrfSafeMethod(settings.type) && !this.crossDomain) {
            xhr.setRequestHeader("X-CSRFToken", csrftoken);
        }
    }
});
```
And this header seems to work, when the token is correct.

PR #2044, opened 2016-08-02
L10n bugfixes

alanmoo:
Addresses @hannahkane's findings at https://github.com/mozilla/learning.mozilla.org/issues/2030#issuecomment-236704850


locales/en-US/messages.properties
```diff
@@ -31,7 +31,7 @@ discourse_forum=Discourse forum
 teaching_activities=Teaching Activities
 web_literacy=Web Literacy
 web_lit_skills=Web Literacy Skills
-web_lit_intro=A framework for entry-level web literacy &amp; 21st Century skills. Explore the map by selecting what you want to learn more about, to see definitions and activities.
+web_lit_intro=A framework for entry-level web literacy & 21st Century skills. Explore the map by selecting what you want to learn more about, to see definitions and activities.
```
gideonthomas: I'm not sure I fully understand this change, shouldn't react-intl accept html entities in strings?

gideonthomas:
one question, r=me otherwise.

locales/en-US/messages.properties
alanmoo: We're using `formatMessage` to embed this into the page. I'm guessing `formatHTMLMessage` would do that. Before this PR, `&amp;` was just getting rendered as a string, so it made sense to just un-encode it.

PR #2057, opened 2016-08-04
Add pontoon script

alicoding:
Of all the thing I added and I forgot to add this! Derp me <.<

### DIFF
--- a/components/pontoon.jsx
+++ b/components/pontoon.jsx
@@ -1,0 +1,14 @@
+var React = require('react');
+
+var Pontoon = React.createClass({
+  render: function() {
+    if (process.env.ENABLE_PONTOON) {
+      return (
+        <script src='https://pontoon.mozilla.org/pontoon.js'></script>
+      );
+    }
+    return null;
+  }
+});
+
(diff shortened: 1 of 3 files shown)


components/pontoon.jsx
```diff
@@ -0,0 +1,16 @@
+var React = require('react');
+
+var enablePontoon = process.env.ENABLE_PONTOON === 'true';
```
alanmoo: process.env.ENABLE_PONTOON isn't a boolean?

alicoding: Apparently not when it's being exposed through Webpack...

alanmoo: *mutters something about webpack documentation under his breath*

alanmoo: I just tested this- you should just be able to drop `process.env.ENABLE_PONTOON` in line 7 and get rid of this bit and it works as expected, just like the other env vars in this repo.

alicoding: Updated

alanmoo:
R+ Pending travis

PR #2064, opened 2016-08-08
Change properties2json to convert locale code

alicoding:
There was a bug on Pontoon caused all language code to be committed as locale code (en_US vs en-US).

I'm also fixing test because we didn't pass `accept-language` in the test runner...


alanmoo:
Is it just doing that with en-US or all language/region pairs?

alicoding:
@alanmoo whatever in the `locales` directory.

alanmoo:
So this changes the underscores to hyphens when writing them to the .json file, so that the rest of the workflow is the same?

(You may want to change the timeout on the tests- since we've got so many more routes now, the 1000ms is probably too slow which is why travis is failing)

alicoding:
@alanmoo just fixed the test timeout.

PR #2071, opened 2016-08-09
Remove default language when rendering pages

### LINKED ISSUE #2065: Fix hardcode locale in createElement function
https://github.com/mozilla/learning.mozilla.org/blob/065ca608fce4fb0ae9b670c5bbaf253b10044e55/lib/page-generate.jsx#L49-L57

This causes other locale to not render with its content and always fallback to en-US

cc @alanmoo

alanmoo:
Fixes #2065

- Get locale from URL (after locale negotiation) and pass it to react to get right localization


alanmoo:
@ali I know my tests are still failing (though I have an idea why- probably needs mocks), but can you take a look at this and let me know if I'm heading in the right direction?

lib/page-generate.jsx
```diff
@@ -100,9 +101,10 @@ function run(location, el) {
     we will be using the value there otherwise we will default to
     what was assigned above.
   */
-  supportedLocale = Object.keys(locales).indexOf(navigator.language) !== -1;
-  currentLocale = supportedLocale ? navigator.language : currentLocale;
+  currentLocale = window.location.pathname.split('/')[1];
   messages = locales[currentLocale] || messages;
+  // Load React's data for the current locale, but if it's not available for that specific country, load the general language data
+  addLocaleData(window.ReactIntlLocaleData[currentLocale] || window.ReactIntlLocaleData[currentLocale.split('-')[0]]);
```
alicoding: I don't think we need to find locale here anymore since we did in the `index.html` 

let's do this instead.

```
  var ReactIntlLocaleData = window.ReactIntlLocaleData;

  Object.keys(ReactIntlLocaleData).forEach((lang) => {
    addLocaleData(ReactIntlLocaleData[lang]);
  });
```

alanmoo: I'm not sure I understand the benefit of this approach- right now I'm just pulling in the one language to memory, which is the only one that's loaded in index.js's script tag. What else would be in window.ReactIntlLocaleData at this point? (Also, I needed to do the split because `bn-BD` isn't a recognized locale by react-intl and it defaulted back to a broken 'en')

alicoding: `window.ReactIntlLocaleData` is an object and the key of the object is the locale name which is being loaded from `index.html` anything. It's a bit redundant here to try and figure out what locale where it's already been done in `index.html`.

alanmoo: But I still need it for `messages`

alicoding: Your call in this case R+ to keep it or change it :)

alanmoo: Ok, I cleaned it up a little but I feel like the loop makes it harder to read and there's no need for multiple locales to load at once, since changing locale hits the server again, plus we need to read the window anyway for messages.

alicoding:
once travis is green then land this baby!

alicoding:
I merged this so I can test something on staging site, hope you don't mind.

PR #2072, opened 2016-08-09
Passing locale from the server

alicoding:


lib/build/index-static.jsx
```diff
@@ -38,7 +38,7 @@ function generateWithPageHTML(url, options, pageHTML) {
   options = _.defaults(options || {}, {
     meta: {}
   });
-  var locale = url.split('/')[0];
+  var locale = options.locale || 'en-US';
```
alanmoo: Not that I want to remove it, but there shouldn't be a case where we need to fall back to 'en-US' at this point, right? Or does testing need it?

lib/page-generate.jsx
```diff
@@ -79,7 +78,7 @@ function generateStatic(url, next) {
 
     var Component = renderProps.routes.slice(-1)[0].component;
     var title = Page.titleForHandler(Component);
-    var html = ReactDOMServer.renderToString(<RoutingContext createElement={createElement} {...renderProps} />);
+    var html = ReactDOMServer.renderToString(<RoutingContext locale={locale} createElement={createElement} {...renderProps} />);
```
alanmoo: Is this what makes line 50's `this.locale` work?

lib/build/index-static.jsx
alicoding: not for regular cases, but testing might and I don't know for sure since I haven't removed them...

lib/page-generate.jsx
alicoding: yup :D

PR #2074, opened 2016-08-09
adds 'should a nav link be a hard reload' functionality

Pomax:
Since this is a single page app and navigation is for a single user in a single client, this PR introduces a library that tracks whether "nav actions for the same page" should hard reload or not, tied to the `LinkAnchorSwap` component.

If some component anywhere sets "hardreload" to true, then a linkanchorswap component for the page we are currently on will, rather than "do nothing", trigger a `window.location.reload()`

fixes https://github.com/mozilla/learning.mozilla.org/issues/2054

### DIFF
--- a/components/link-anchor-swap.jsx
+++ b/components/link-anchor-swap.jsx
@@ -4,25 +4,53 @@ var ga = require('react-ga');
 var classNames = require('classnames');
 var OutboundLink = ga.OutboundLink;
 
+var resetreload = require('../lib/resetreload');
+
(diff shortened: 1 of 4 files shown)


Pomax:
I've rewritten this PR to something more reacty, although react-router has nothing built in for this (certainly not 1.x which we're using)

alanmoo:
This works for me. I'm still not sure why we don't just have a unique route for the application, but this solves the problem it's set out to solve.

PR #2078, opened 2016-08-10
Model for teaching activities API

alicoding:
This is the first draft of the model for teaching activities that we are going to create using django and consume them here instead of static code.

### DIFF
--- a/API.md
+++ b/API.md
@@ -9,7 +9,6 @@ difficulty | CharField (choices)
 duration | CharField (choices)  
 **Skills** | *:*  
 **WebLitSkills** | *:*  
-**Topic** | 1:*  
 **Modules** | *:*  
 **locale** | CharField (choices) (pycountry)  
 translation_of | ‘self’  
@@ -29,11 +28,6 @@ short_name | CharField
 description | TextField  
 **Competencies** | 1:*  
 **Skills** | *:*  
-**Topic** | 1:*  
-
-<pre>Topic</pre>
-name | CharField  
-description | TextField  
 
 <pre>Competencies</pre>
 text | TextField  
--- a/API.md
+++ b/API.md
@@ -10,7 +10,7 @@ duration | CharField (choices)
 **Skills** | *:*  
 **WebLitSkills** | *:*  
 **Topic** | 1:*  
-**LessonPlan** | *:*  
+**Modules** | *:*  
 **locale** | CharField (choices) (pycountry)  
 translation_of | ‘self’  
 
@@ -38,7 +38,7 @@ description | TextField
 <pre>Competencies</pre>
 text | TextField  
 
-<pre>LessonPlan</pre>
+<pre>Modules</pre>
 title | CharField  
 description | TextField  
 short_description | TextField  
@@ -46,4 +46,5 @@ learning_objectives | TextField
 image_url | UrlField  
 image_retina_url | UrlField  
 level | CharField (choices)  
-developed_by | TextField  
+author | CharField
+author_link | CharField
--- a/API.md
+++ b/API.md
@@ -1,0 +1,49 @@
+<pre>Activities</pre>
+title | CharField  
+subtitle | CharField  
+description | TextField  
+image_url | UrlField  
+image_retina_url | UrlField  
+url | UrlField  
+difficulty | CharField (choices)  
+duration | CharField (choices)  
+**Skills** | *:*  
+**WebLitSkills** | *:*  
+**Topic** | 1:*  
+**LessonPlan** | *:*  
+**locale** | CharField (choices) (pycountry)  
+translation_of | ‘self’  
+
+
+
+<pre>Skills</pre>
+name | CharField  
+description | TextField  
+**Competencies** | 1:*  
+image_url | UrlField  
+image_retina_url | UrlField  
+
+<pre>WebLitSkills</pre>
+name | CharField  
+short_name | CharField  
+description | TextField  
+**Competencies** | 1:*  
+**Skills** | *:*  
+**Topic** | 1:*  
+
+<pre>Topic</pre>
+name | CharField  
+description | TextField  
+
+<pre>Competencies</pre>
+text | TextField  
+
+<pre>LessonPlan</pre>
+title | CharField  
+description | TextField  
+short_description | TextField  
+learning_objectives | TextField  
+image_url | UrlField  
+image_retina_url | UrlField  
+level | CharField (choices)  
+developed_by | TextField  


alicoding:
* Those that bolded (**) is a foreignKey to another instance
* 1:* = one to many
* *:* = many to many
* (choices) just a dropdown of presets choices
* Ignore pycountry it's just a library I'll be using for locale dropdown (it will be presets of locales)
* CharField (short text)
* TextField (long text / description)
* 'self' this will be use to display meta data for any forked or translation of activity in another language if available.

Pomax:
what is this in? It doesn't render as markdown suggesting the .md extension is not quite correct.

alicoding:
@Pomax this will be removed, I'm just trying to get everyone to see it before I start the implementation.

API.md
```diff
@@ -0,0 +1,49 @@
...
+
+<pre>LessonPlan</pre>
+title | CharField  
+description | TextField  
+short_description | TextField  
+learning_objectives | TextField  
+image_url | UrlField  
+image_retina_url | UrlField  
+level | CharField (choices)  
+developed_by | TextField
```
hannahkane: developed_by could have an optional url as well, so we can link to, for example, the author's twitter account

API.md
```diff
@@ -0,0 +1,49 @@
...
+
+<pre>Topic</pre>
+name | CharField  
+description | TextField  
+
+<pre>Competencies</pre>
+text | TextField  
+
+<pre>LessonPlan</pre>
+title | CharField
```
hannahkane: I think it would be useful to have a duration field at the LessonPlan level in addition to the Activity level

API.md
```diff
@@ -0,0 +1,49 @@
...
+
+<pre>WebLitSkills</pre>
+name | CharField  
+short_name | CharField  
+description | TextField  
+**Competencies** | 1:*  
+**Skills** | *:*  
+**Topic** | 1:*  
+
+<pre>Topic</pre>
```
hannahkane: what is Topic used for?

API.md
```diff
@@ -0,0 +1,49 @@
...
+image_url | UrlField  
+image_retina_url | UrlField  
+url | UrlField  
+difficulty | CharField (choices)  
+duration | CharField (choices)  
+**Skills** | *:*  
+**WebLitSkills** | *:*  
+**Topic** | 1:*  
+**LessonPlan** | *:*  
+**locale** | CharField (choices) (pycountry)
```
alanmoo: Will this be a full locale or only language? I'm guessing the former, but we should work out how we're going to show available activities in a supported language, regardless of locale.

Example: let's say we have Activities in en-US and a new one written in en-GB. Do we show en-GB just that one, or all English ones? Not necessarily a concern for the model, but something to ponder for API design.

API.md
```diff
@@ -0,0 +1,49 @@
...
+subtitle | CharField  
+description | TextField  
+image_url | UrlField  
+image_retina_url | UrlField  
+url | UrlField  
+difficulty | CharField (choices)  
+duration | CharField (choices)  
+**Skills** | *:*  
+**WebLitSkills** | *:*  
+**Topic** | 1:*
```
hannahkane: Does this mean there's one Topic per Activity? Or one Activity per Topic?

API.md
```diff
@@ -0,0 +1,49 @@
...
+**Skills** | *:*  
+**WebLitSkills** | *:*  
+**Topic** | 1:*  
+**LessonPlan** | *:*  
+**locale** | CharField (choices) (pycountry)  
+translation_of | ‘self’  
+
+
+
+<pre>Skills</pre>
```
alanmoo: What are some examples of current Skills vs. WebLitSkills?

API.md
alanmoo: I believe it means each Activity can have many topics.

API.md
```diff
@@ -0,0 +1,49 @@
+<pre>Activities</pre>
```
hannahkane: If I understand this model correctly, an Activity is made up of several Lesson Plans. If that's the case, I might suggest we use the term "Modules" instead of "Activities."

API.md
hannahkane: Oh, I get it. Thanks. :)

API.md
alanmoo: Is one a sum of the others though?

hannahkane: I think it's more important to have the duration at the specific lesson plan level than at the Activity level, since people will likely mix and match individual lesson plans, and not necessarily follow the entire Activity. But if we can dynamically generate a sum total duration for the Activity, that would be neat.

API.md
alanmoo: If this is 1 to many, does that mean the topics aren't shared across activities if we wanted to filter that way in the future?

API.md
```diff
@@ -0,0 +1,49 @@
...
+**Topic** | 1:*  
+
+<pre>Topic</pre>
+name | CharField  
+description | TextField  
+
+<pre>Competencies</pre>
+text | TextField  
+
+<pre>LessonPlan</pre>
```
hannahkane: Should LessonPlan also have a url field? How will people get to the actual main content of the Lesson?

API.md
alicoding: Actually this will be single activity with meta data which might be part of lesson plans.

API.md
```diff
@@ -0,0 +1,49 @@
+<pre>Activities</pre>
+title | CharField  
+subtitle | CharField
```
hannahkane: Do we have examples of Activities with subtitles?

API.md
alicoding: You can get list of activities by Topic and able to find out what topic the activity is part of, example here would be list all activities in `Read`

alanmoo:
Appropriate examples of the different data types from the existing site would be super useful for analyzing this. @alicoding, can you add a comment with examples for each?

API.md
alicoding: this https://learning.mofostaging.net/en-US/web-literacy/skills vs https://learning.mofostaging.net/en-US/web-literacy/write/design/

API.md
alicoding: Read, Participate and Write.

API.md
alicoding: You will be able to filter and get everything under each LessonPlan

hannahkane: Now that I realize I had Activity and LessonPlan backwards in my head, this makes more sense. Comment above still stands, though, just with the terms reversed. If we could provide a generated sum total for the collection of Activities that make up the LessonPlan, that would be useful.

alicoding: and LessonPlan can be created with whatever activity you want it to be part of the lesson

API.md
alicoding: I'm thinking of making this be something that we can use HTML in this field because right now it's really inconsistent as to how it's being format.

API.md
```diff
@@ -0,0 +1,49 @@
...
+url | UrlField  
+difficulty | CharField (choices)  
+duration | CharField (choices)  
+**Skills** | *:*  
+**WebLitSkills** | *:*  
+**Topic** | 1:*  
+**LessonPlan** | *:*  
+**locale** | CharField (choices) (pycountry)  
+translation_of | ‘self’  
+
```
hannahkane: The Activities will need an author field with url, I think

alicoding:
@alanmoo honestly it is hard to give example since every data I got for each model here is based on what's on the learning site itself...

API.md
alicoding: that's not part of the current model is it?

API.md
hannahkane: If we're thinking "Read", "Write", and "Participate" would be the topics, I think we can probably scrap that. We only used that for the Web Lit Basics modules, and it's been difficult to group activities into those categories for all of the modules we've created since. I also don't think there's much user value in those topics.

API.md
alicoding: https://dl.dropboxusercontent.com/spa/n5v3bx9nnjkdpzf/z-n1h90l.png

>Understanding credibility

API.md
alicoding: Will rename this :)

API.md
alicoding: Could you suggest something else? Categories? Or Tags?

API.md
hannahkane: Are "Skills" the four 21st Century Skills that have been identified (Problem Solving, Collaboration, Creativity, and Communication) and are "WebLitSkills" the 14 skills around the outer ring of the web lit map (Design, Revise, Synthesize, Navigate, Protect, etc.)?

alicoding: Yes! That's the one @hannahkane :D

API.md
hannahkane: Ah, okay. Can this field be optional?

alicoding: Right now in each Module for each activities they have subtitiles, but if you think some might not have them then I can totally make this optional.

API.md
hannahkane: I can imagine "tags" being useful. People might want to browse for activities with "javascript", "css", or "passwords" for example. 

The Read, Write, Participate framework was used to group and order content on the page (see https://learning.mozilla.org/en-US/activities/web-lit-basics/)—that does not seem necessary. But a tag field could be useful for searching/browsing.

API.md
```diff
@@ -0,0 +1,49 @@
...
+description | TextField  
+image_url | UrlField  
+image_retina_url | UrlField  
+url | UrlField  
+difficulty | CharField (choices)  
+duration | CharField (choices)  
+**Skills** | *:*  
+**WebLitSkills** | *:*  
+**Topic** | 1:*  
+**LessonPlan** | *:*
```
hannahkane: Can we do something to allow control over the order an Activity appears in the Lesson Plan? e.g. LessonPlanIndex, so that I can, say, put my six activities in a particular order within the lesson plan? 

If not, what order will they appear in? Chron by creation date?

API.md
alanmoo: @hannahkane it's also [here](https://learning.mozilla.org/en-US/activities/protect-your-data/)

API.md
hannahkane: Here's an example of an activity with an author: https://mozilla.github.io/mozilla-club-activity-kraken-the-code/#en

hannahkane: I wonder if we should store creation_date or last_updated date. I can imagine use cases for that. We might want to offer a view of recent content, for example.

API.md
hannahkane: Maybe good to rename Skills to something more descriptive - is "21CSkills" an acceptable name?

API.md
alanmoo: Ah, I was going to ask that earlier but got distracted. We want to consider this in the scope of something like network pulse, so those two fields would likely be very useful.

API.md
hannahkane: I would say optional. We've been following this convention internally, but I wouldn't expect others in the network to do the same.

API.md
hannahkane: Yeah, we've used it in a few places, but it has felt kind of awkward. Many activities could easily be under Read, Write, AND Participate, so the forced 1:1 has been kind of frustrating.

alanmoo: Fair enough. I'd be all for getting rid of this (and changing the templates of the pages that use them)

alanmoo: Oh, missed the above comment. Tags work too.

API.md
alanmoo: I'd rather enforce a style than leave it open- I don't really see a need for a ton of customizability here.

gideonthomas: The current data on the learning site suggests otherwise. There is no consistency between them. Some have multiple authors, some have one, in some only one author has a link, in others multiple authors have different links, some have "author - blah blah blah", etc.
Unless we are willing to modify the existing data, I think Ali's suggestion of having it be HTML is fine (we can validate it when users enter data but this shouldn't be reflected in the model)

API.md
gideonthomas: Yep, that duration aggregation can be handled in the "business logic" level but this part of the model looks right

API.md
gideonthomas: @alicoding maybe we should rename it to something else.

API.md
gideonthomas: @hannahkane probably not necessary here (although a comment in the code would be a good idea). These names are what the database/code know of. What is shown to the user via the API or via the Admin interface can be customized to whatever we want.

API.md
gideonthomas: @alicoding I would say add the author but make it use the same format that you use for `developed_by` for Modules

API.md
gideonthomas: It can be ordered by creation date. But if we want an order, it might be a bit complicated.

@hannahkane Am I correct in saying an activity can belong to multiple modules and a module has multiple activities?

If I am wrong in the above assertion, @alicoding it will actually be 1 to many not many to many.

Either way, @alicoding you can add ordering in a module via a bridge table (or more appropriately, a `through` table that will store the order).

alicoding: A bit hacky but common here is to add a hidden field number where we can put number that we want to order the activity by (lowest first) and we can do orderby number here

API.md
alicoding: Wait but what do we call those 3 though? Are they tags as well?

API.md
gideonthomas: I think a `through` table makes more sense here though because that's what it's meant for - storing extra data about a relation. If the relation is between a module and an activity, the extra data would be the order. I believe we did a similar thing in the science api as well so maybe that could be a good reference for implementing this

alicoding: But what if someone wanna change the order after they all have been published ?

API.md
hannahkane: +1 to @alanmoo's idea of getting rid of those and changing the templates so they're not needed.

API.md
hannahkane: It's true that we've not been consistent so far (we haven't had a database behind these so far!) I feel like giving people the option to filter by author would be kind of cool/useful, but not necessarily a requirement. If it were, we would definitely need to be able to handle the multiple author use case and the no-link use case, though, since those are common.

API.md
gideonthomas: You can totally do that through the admin interface. We expose the `through` table through the admin interface and we can make it look user friendly so that it is easy to change.

API.md
alicoding: I'm still a bit unclear what we want to do exactly here...

alicoding: Also the relation here is not 1:1 though it looks like it is

API.md
alicoding: created/updated date is actually a default on each views.

API.md
alicoding: I'm a bit unclear on this on still...

API.md
gideonthomas: @alicoding basically you can just get rid of this field (and the model)

API.md
gideonthomas: @alicoding you can delete this model based on the comment thread here: https://github.com/mozilla/learning.mozilla.org/pull/2078/files#r74445980

cadecairos:
I have changed the base branch of this repo to master, so we'll have to rebase 👍

gideonthomas:
@cadecairos this PR won't be merged in afaik

alicoding:
- Author will have its own Model with name and link
- Activity will have tags
- Tags will have its own Model
- Relationship between Activity and Author is *:*

PR #2085, opened 2016-08-12
Fall back to en-US properly on incomplete locales

### LINKED ISSUE #2080: Bug in properties2json causes string to not display correctly
For stuff that's not translated right now you will see something like this

It should actually merged any missing keys from default language (`en-US`) to other languages.
We actually have this behaviour in other repo, but not sure why this is not working as expected here.

alanmoo:
Fixes #2080

- When passing in messages for a locales, this first generates a new object with en-US keys, then overwrites it with supported strings from the current locale.

Could be optimized/refactored into mofo-localize, but this is a quick and easy win for now.

### DIFF
--- a/lib/page-generate.jsx
+++ b/lib/page-generate.jsx
@@ -23,6 +23,8 @@ var IntlProvider = ReactIntl.IntlProvider;
 var addLocaleData = ReactIntl.addLocaleData;
 var currentLocale;
 
+var assign = require('object-assign');
+
 /**
  * content for redirect pages
  */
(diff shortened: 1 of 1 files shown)


alicoding:
I know it says WIP in the title, but just want to point out (in case you haven't notice, but you might have) when you visit `/de` here you will see English first then `de` content will be displayed...

alanmoo:
This should work now. It's the slightest amount WET, but trying to integrate this into MoFo-Localize right now is going to take too much time. It's on [the list](https://github.com/alanmoo/mofo-localize/issues/1) though.

PR #2089, opened 2016-08-15
application form submission UX

### LINKED ISSUE #2053: application form followup: tweak the UX when submitting the form
Right now it's possible for the submission step to take a few seconds, which might make users think nothing is happening. We probably want to spend a bit of design effort on making sure the form somehow clearly indicates it's being submitted (maybe grey out the page?) before it transitions to the "you've submitted your application" screen.

Pomax:
This PR adds a "submitting" state to the club application form so that users get a visual signal that their form is being submitted, and see the submit/back buttons greyed out and inactive, so they don't try clicking on them "some more".

This fixes #2053 

The easiest way to test this is to comment off lines 156 through 159 in ClubForm.jsx, as well as lines 168-170: this turns off form validation as well as the actual http POST operation normally associated with the club application submission process, while showing the UI in the greyed out state when clicking through the first "next" button for the club application form, and then clicking the "submit" button.

### DIFF
--- a/less/pages/clubs-form.less
+++ b/less/pages/clubs-form.less
@@ -79,9 +79,14 @@
           background-color: @paleBlue;
           box-shadow: 0px 0.4rem 0px #999,
                       0px 0.1rem 0px rgba(255, 255, 255, 0.5) inset;
+
+          &[disabled] {
+            box-shadow: none;
+          }
+
           color: #666;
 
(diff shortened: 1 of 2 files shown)


flukeout:
Reviewin'

flukeout:
LOOKS GOOD!

PR #2097, opened 2016-08-18
Club Pages QA Fixes

flukeout:
Some fixes from the QA Issue here...
https://github.com/mozilla/learning.mozilla.org/issues/2087

* Moved Apply CTA to hero block (it's also at the bottom of the start page)
* Fixed typos that @hannahkane found

There are some more cosmetic changes left identified by @sabrinang but they can be fixed after this stiff goes live.

### DIFF
--- a/pages/clubs/index.jsx
+++ b/pages/clubs/index.jsx
@@ -18,6 +18,9 @@ var ClubsGuides = React.createClass({
     pageClassName: 'clubs-guides',
     pageTitle: 'Clubs Guides & Resources'
   },
+  contextTypes: {
+    intl: React.PropTypes.object
+  },
   showApplication: function(){
     this.setState({
       showApplication: true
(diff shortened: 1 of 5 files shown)


pages/clubs/index.jsx
```diff
@@ -34,6 +37,9 @@ var ClubsGuides = React.createClass({
         <HeroUnit>
           <h1>Mozilla Clubs</h1>
           <h2>Local groups that read, write, and participate on the web</h2>
+          <div className="single-button hero-button">
+            <a className="btn" onClick={this.showApplication} href="#">{this.context.intl.formatMessage({id: 'apply_to_be_captain_link'})}</a>
```
Pomax: this really should be a `<button>`, not an `<a>` - it's not a page navigation element

gvn:
Everything looks fine to me aside from the button/a comment above.

gvn:
Let's do the button/a as a follow on since Luke's on PTO.

PR #2099, opened 2016-08-18
Move babel build dependencies from 'devDependencies' to 'dependencies'

### LINKED ISSUE #2098: Babel should not be a devDependency
Babel was added as a devDependency to the webpack build in [this commit](https://github.com/mozilla/learning.mozilla.org/commit/ee484455a8a8fed267af00c3f425cf371a742716)

Unfortunately, Heroku's NodeJS buildpack runs `npm install --production` which means that these are build-time depenencies, which means we need to move them into the plain-old dependencies list.

P1 Because production pushes are currently blocked

cadecairos:
- [X] Localization addressed

Fixes #2098 

- Move babel build dependencies from 'devDependencies' to 'dependencies'

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,11 @@
     "watch": "gulp watch"
   },
   "dependencies": {
+    "babel-cli": "^6.11.4",
     "babel-core": "^6.3.21",
+    "babel-loader": "^6.2.4",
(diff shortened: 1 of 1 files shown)


alicoding:
R+

PR #2109, opened 2016-08-19
Club application cleanup

Pomax:
This refactor cleans up the club application form source files, turning the steps into basic definition JSON, which is then passed into a `<Form>` element which does the building for us.

fixes https://github.com/mozilla/learning.mozilla.org/issues/2236
fixes https://github.com/mozilla/learning.mozilla.org/issues/2052
fixes https://github.com/mozilla/learning.mozilla.org/issues/2116


gvn:
@Pomax Did you see the Travis failure?

Pomax:
@gvn want to give this a look over now that both steps 1 and 2 have been cleaned up? (I've not hooked up the actual step transitioning, including submission, yet so if you load up the application form you just see "Everything", even though validating is boxed to each form)

pages/clubs/ClubForm.jsx
```diff
@@ -142,20 +142,15 @@ var ClubForm = React.createClass({
...
-        this.setState({ currentStep: nextStep });
-      }.bind(this);
-
-      if (this.state.currentStep === 1) {
-        this.submitForm(goToNext);
-      } else {
-        goToNext();
-      }
+    var nextStep = Math.min(this.state.currentStep + 1, 2);
+    var goToNext = function() {
```
gvn: Can this be an arrow function to avoid bind?

pages/clubs/Formed.jsx
```diff
@@ -0,0 +1,281 @@
...
+      } else {
+        shouldHide = this.state[controller] !== controlValue;
+      }
+    }
+
+    if (label) {
+      label = <label key={name + 'label'} hidden={shouldHide}>{label}</label>;
+    } else { label = null; }
+
+    if (!label) {
```
gvn: Could this go inside the else case on line 79?

pages/clubs/ClubForm.jsx
Pomax: it can now that we updated the build system! \o/

pages/clubs/Formed.jsx
Pomax: good catch

Pomax:
@gvn this is as close a refactor as it's going to get for now - maybe we can run through this together on Tuesday as kind of a pair review - there is now a `Form` (does what you think), `MultiSectionForm` (which is one form built up of multiple sections of form fields), and `MultiPageForm` (which is a collection of forms presented in sequence instead of all at the same time, with back/next button navigation).

The `ClubForm` now simply builds a single `MultiPageForm` based on JSON located in `./pages/clubs/form/data`, and the way the second part of our form now works is clever but we should discuss if it's good enough to port to mofo-ui.

Still left to do:
- [ ] update the LESS so that this refactor looks the same as before
- [ ] discuss how best to trigger validation, as this triggers validation error highlighting on everything.

gvn:
**UX-ey fixups for follow ups:**

- It'd be nice to have the option to show validation helper text beneath the element it applies to. Correlating a bunch of feedback messages back to their respective inputs is more cognitively difficult.
- Don't run validation form elements before they've been at least focused. Getting a ton of red fields before you've had a chance to touch them is intense.
- For checkboxes and radio inputs it would be nice to use wrapping label elements so that you can click the adjacent text to set them.

pages/clubs/ClubForm.jsx
```diff
@@ -83,134 +114,41 @@ var ClubForm = React.createClass({
...
-    );
-
-    return (
-      <div key="buttons" className="proceed">
-        <div>{buttons}</div>
-        { (this.state.currentStep < 2) ? <ProgressBar value={this.state.progress}/> : null }
-      </div>
-    );
+  updateProgress: function(progress) {
+    console.log(progress);
```
gvn: Remove?

pages/clubs/ClubForm.jsx
```diff
@@ -83,134 +114,41 @@ var ClubForm = React.createClass({
...
     this.setState({
       submitting: true,
-      step: this.STEP_WAIT_FOR_NETWORK,
       networkError: false,
-    }, function() {
-      teachAPI.addClub(clubState, function(err, data) {
-        networkHandler(err, data, next);
+      step: STEP_SENDING_RESULTS
+    }, () => {
+      console.log(clubState);
```
gvn: remove

pages/clubs/ClubForm.jsx
```diff
@@ -83,134 +114,41 @@ var ClubForm = React.createClass({
...
+
+//      teachAPI.addClub(clubState, (err, data) => {
+//        if (err) {
+//          this.setState({
+//            networkError: !!err,
+//            result: err ? null : data
+//          })
+//        } else {
```
gvn: nuke instead of commenting out...

pages/clubs/form/builder/Form.jsx
```diff
@@ -0,0 +1,397 @@
...
+  field definition objects:
+
+   {
+     type: ["text"|"textarea"|choiceGroup"|"checkbox"|checkboxGroup"|ReactComponent],
+     label: string data,
+     placeholder: string data (optional)
+     validator: [instance or array of validator objects],
+     metered: boolean (optional),
+     controller: controller object (optional),
+   }
```
gvn: Instead of just documenting this and hoping people follow it, can you add `propTypes` validators as well?

pages/clubs/ClubForm.jsx
Pomax: Not in the slightest - this is commented off so it doesn't try to send the data to the Teach API for PR purposes,  which the real code definitely needs to do =)

But requires 3 additional services to run... =(

pages/clubs/form/builder/Form.jsx
Pomax: That's probably doable - unless we also want something along the lines of "if you don't pass a validator at all, without an error message, we run the default validator with error "You need to fill in {fieldname}". Can we do proptypes for optionals?

gvn: I think there may be confusion here. I'm just talking about adding [propTypes](https://facebook.github.io/react/docs/reusable-components.html#prop-validation) to validate props passed to the new form components, not anything with the form validation itself.

Pomax:
@gvn I fixed up the styling so that things should look the same as they do now on learning.mozilla.org, can you do one final review to check whether it looks the same enough to land?

@gideonthomas since you're more likely to have a functional teach-api setup going, can you verify that this cleanup preserves the ability to send all form fields to the teach-api for a club application, compared to the current implementation? It *should*, but we want to make sure no fields go missing.

gideonthomas:
@Pomax a bunch of fields are missing when I try to run the code:
- The "Hosting Reason" and "How they heard" fields are never saved regardless of which option I choose on the first/second page
- The "Regional Coordinator" field is not saved when I select the option to choose a Regional Coordinator on the first page
- The "intent" field is always set to "start" even if I choose "Integrate into existing Mozilla Club" on the second page
  - Related to this, I see the exact same fields for "Integrate into existing Mozilla Club" and "Start a new Club" while I see different fields on staging for both. A consequence of this is that I cannot enter the "Current website URL" field

Pomax:
"current website" was technically missing, that's been added in for the "integrate" field set. Odd about the other fields: I'll have a look at why they're not doing the right thing in the morning.

Pomax:
@gvn added website field validation, and removed some console logs that for some reason were still in the code. Can you do one last re-review?

gideonthomas:
@Pomax the "intent" error is fixed but all the others still happen. I still see the exact same form for both "Start a club" and "Integrate into another Mozilla Club".

gideonthomas:
@Pomax this works now!

PR #2110, opened 2016-08-22
Hide maker-party sidebar on partially-supported locales

alanmoo:
Fixes #2110, #2111 

- Adds routes for events page in Maker party focused locales
- Hides sidebar for all non `en-US` locales.

For use with uniquely localized maker-party pages

### DIFF
--- a/lib/routes.jsx
+++ b/lib/routes.jsx
@@ -2,10 +2,14 @@ var React = require('react');
 
 var ReactRouter = require('react-router');
 var Router = ReactRouter.Router;
+var IndexRedirect = ReactRouter.IndexRedirect;
 var Route = ReactRouter.Route;
 var Redirect = ReactRouter.Redirect;
 var IndexRoute = ReactRouter.IndexRoute;
 var locales = Object.keys(require('../dist/locales.json'));
+//Safety check
+var SUPPORTED_LOCALES = ['en-US'].filter(supportedLocale => locales.indexOf(supportedLocale) > -1);
+var MAKER_PARTY_LOCALES = ['de', 'fr', 'es', 'pl', 'it', 'nl', 'cs', 'bg'];
 
 
 // verify we have at least one locale
@@ -74,7 +78,7 @@ if (process.env.ENABLE_BADGES) {
     'badge/:id': require('../pages/badges/badge-single.jsx'),
     'badge/:id/:slug': require('../pages/badges/badge-single.jsx')
   });
-}
+};
 
 /**
  * Redirects from old URLs to new URLs
@@ -118,7 +122,6 @@ var routeElements = Object.keys(pages).map(function(path) {
   return <Route path={path} component={pages[path]} key={path}/>;
 });
 
-var makerPartyLocales = ['de', 'fr', 'es', 'pl', 'it', 'nl', 'cs', 'bg'];
 
 // <Redirect> elements
 var redirectElements = Object.keys(redirects).map(function(path) {
@@ -137,7 +140,7 @@ function buildRoutes() {
   var routes = [];
   var localeURLs = [];
 
-  locales.forEach(function(locale) {
+  SUPPORTED_LOCALES.forEach(function(locale) {
     routes.push(
       <Route key={locale} path={locale} component={require('../components/page.jsx')}>
         <IndexRoute component={require('../pages/home.jsx')} />
@@ -159,10 +162,11 @@ function buildRoutes() {
     });
   });
 
-  makerPartyLocales.forEach((locale) => {
-    localeURLs.push(`${locale}/events`);
+  MAKER_PARTY_LOCALES.forEach((locale) => {
+    localeURLs.push(`${locale}`, `${locale}/events`);
     routes.push(
       <Route key={locale} path={locale} component={require('../components/page.jsx')}>
+	<IndexRedirect to="/" />
 	<Route path="events" component={require('../pages/events.jsx')} />
       </Route>
     );
--- a/lib/routes.jsx
+++ b/lib/routes.jsx
@@ -118,6 +118,8 @@ var routeElements = Object.keys(pages).map(function(path) {
   return <Route path={path} component={pages[path]} key={path}/>;
 });
 
+var makerPartyLocales = ['de', 'fr', 'es', 'pl', 'it', 'nl', 'cs', 'bg'];
+
 // <Redirect> elements
 var redirectElements = Object.keys(redirects).map(function(path) {
   return <Redirect from={path} to={redirects[path]} key={path} />;
@@ -155,7 +157,15 @@ function buildRoutes() {
 
       localeURLs.push(newkey);
     });
+  });
 
+  makerPartyLocales.forEach((locale) => {
+    localeURLs.push(`${locale}/events`);
+    routes.push(
+      <Route key={locale} path={locale} component={require('../components/page.jsx')}>
+	<Route path="events" component={require('../pages/events.jsx')} />
+      </Route>
+    );
   });
 
   return {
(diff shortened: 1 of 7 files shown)


alanmoo:
Putting this up now as a start- just added some simple classes to expand the sidebar if a class is present on the body. Obviously we still need the logic to make it work, but check out the deployment for an example. @Pomax or @gvn do these styles make sense? I know it's a slight hack of the grid system, which I hate doing, but I think this is the simplest way to handle it considering the sidebar is `position:fixed`

alanmoo:
I...think I did it. See above deployment, events page as an example (https://learning-mozilla-org-s-pr-2110.herokuapp.com/bg/events)

alanmoo:
Needs changes before being merged in, but @ScottDowne and @alicoding can you pre-review it when you get a chance?

lib/routes.jsx
```diff
@@ -115,6 +115,8 @@ var routeElements = Object.keys(pages).map(function(path) {
   return <Route path={path} component={pages[path]} key={path}/>;
 });
 
+var makerPartyLocales = ['de', 'fr', 'es', 'pl', 'it', 'nl', 'cs', 'bg'];
```
alicoding: The list is actually longer here, and probably @scottdowne can tell us the rest

ScottDowne: Based on:

Tier 1: English, French, German, Spanish, Polish

Tier 2: Italian, Dutch, Czech, Bulgarian, 

from https://github.com/MozillaFoundation/Advocacy/issues/151

it looks right to me.5 primary (including English)

and 4 secondary

alicoding: What about `fr-FR`, `es-ES`? and maybe the rest if there are any...

alanmoo: (This is why I was thinking about env variables rather than hard-coding the list in.)

ScottDowne: I guess it depends on how learning.m.o works when it comes to fr-FR and fr.

We want France French, and Spain Spanish, so whatever codes fall under that.

Interestingly enough, in donate, fr-FR isn't a thing, and es-ES isn't even complete, so maybe it's low priority.

alanmoo: I'm not aware of the specific differences for French and Spanish, but if we're not talking about currency or units of measurement, I hardly notice the difference between en-US and en-GB. I'd guess that if these small pages are in the base language itself, that's good enough to get the message across such that no one's going to be insulted it's not localized in their region.

ScottDowne: Yeah, based on en-US and en-GB, you're right. But other languages are not quite the same for this. The differences are a bit more obvious.

alicoding: @ScottDowne it was totally a thing :) 215,642(6.42%) of users requested `es-es` and `fr-FR` was set on the users browser when they visit donate repo, but because the snippet dealt with that that's why we didn't have any issue with it.

alicoding: @alanmoo lets do `env` as you suggested.

ScottDowne: @alicoding How does learning.mozilla.org handle this? If fr and fr-FR are the same (I'm assuming this, but because I've never come across fr-FR I'm not sure) we need to flatten these codes to match whatever is stored in the repo (and pontoon) so does learning.mozilla.org at some point do a locale code convert, or does it store two locale files that are similar for each code? Or, otherwise, how does this work?

alicoding: I would suggest for Marker Party we don't care about the region. So es-MX and es-ES will be mapped to es and same goes for the rest

ScottDowne: And yes, the snippet solves this. It basically does the convert for us.

alanmoo: @ScottDowne in learning, we've got some logic that picks the language based on user requested locale. If they request es-XX but only es is available, it'll return that.

alanmoo:
I think something in the recent build updates broke a small part of this, so I'm going to fix that before one more review.  @alicoding do you still think I should move the makerparty supported locales to another env var?

@ScottDowne Do you want me to create a stub of a page at  `/makerparty` that this works with?

ScottDowne:
@alanmoo I don't really need a stub page, so don't make one on my behalf.

lib/routes.jsx
```diff
@@ -152,10 +157,19 @@ function buildRoutes() {
...
       localeURLs.push(newkey);
     });
+  });
 
+  MAKER_PARTY_LOCALES.forEach((locale) => {
+    localeURLs.push(`${locale}`, `${locale}/events`);
+    routes.push(
+      <Route key={locale} path={locale} component={require('../components/page.jsx')}>
+        <IndexRedirect to="/" />
+        <Route path="events" component={require('../pages/events.jsx')} />
```
ScottDowne: Just a friendly reminder that we also need `/events/resources` but I think you know that now.

lib/routes.jsx
```diff
@@ -152,10 +157,19 @@ function buildRoutes() {
...
-      
+
       localeURLs.push(newkey);
     });
+  });
 
+  MAKER_PARTY_LOCALES.forEach((locale) => {
+    localeURLs.push(`${locale}`, `${locale}/events`);
+    routes.push(
+      <Route key={locale} path={locale} component={require('../components/page.jsx')}>
```
ScottDowne: What is components/page.jsx

lib/routes.jsx
```diff
@@ -152,10 +157,19 @@ function buildRoutes() {
...
+        <IndexRedirect to="/" />
```
ScottDowne: Would this make the index be a maker party page or the main teach home page?

lib/routes.jsx
alanmoo: The main component that wraps all pages

lib/routes.jsx
alanmoo: Main teach homepage, where it re-does locale negotiation (so `/en-US` for now)

ScottDowne: Ah, so if they go to `de/` they get `en-US/`

I'm fine with that, actually, but I worry Ali might disagree.

ScottDowne: err, I messed that up. Give me a sec, I think I confused myself.

ScottDowne: I think I was wondering if we should redirect `/de/` to `/de/events/`

I'm fine if we don't it's just a thought.

ScottDowne:
I just had some questions, but, I think this is fine.

alanmoo:
@alicoding and @ScottDowne I think this is really ready for review this time. For realz.

Once merged, it will only affect the current site if a user from a partially supported locale lands on the events page/refreshes while on that page.

Pomax:
I'll pitch in and note that the odd "going to a page with an unsupported locale" works quite properly now

lib/routes.jsx
```diff
@@ -155,7 +159,18 @@ function buildRoutes() {
...
     });
+  });
 
+  MAKER_PARTY_LOCALES.forEach((locale) => {
+    localeURLs.push(`${locale}`, `${locale}/events`, `${locale}/events/resources`);
+    routes.push(
+      <Route key={locale} path={locale} component={require('../components/page.jsx')}>
+	<IndexRedirect to="/" />
+	<Route path="events" component={require('../pages/events.jsx')}>
+	  <Route path="resources" component={require('../pages/event-resources.jsx')} />
```
ScottDowne: Small nit, there are some tabs here.

alanmoo: Ack, editor funkiness. Will fix/check the rest of the PR for this.

ScottDowne: We already talked about this, but just leaving it here as a reminder.

There is funkiness in components/page.jsx and needs <Route path="resources" component={require('../pages/event-resources.jsx')} /> to be a direct child.

R+ with that done.

alanmoo:
I've updated this with the tab/spacing fix and the routing fix you mentioned in chat, @ScottDowne

ScottDowne:
Looks good to me.

Issue #2108, opened 2016-08-19
Update tests so they are less likely to time out


alanmoo:
With the implementation of l10n, we're now testing all pages getting 200 responses across all locales. That's a lot of pages, and the test has a timeout in it that needs to grow with each new locale. Let's change the test so it tests all routes in `en-US` and then only the homepage for all the other locales, so the test scales at `o`, not `n*o`.

alanmoo:
I'm starting to feel like this is dependent on #1509

alanmoo:
Fixed by #2121

PR #2102, opened 2016-08-18
Move properties2json to npm modules

alanmoo:
- [x ] Localization addressed

**Changes proposed in this pull request:**
- Move properties2json to the new [pontoon-2-json](https://www.npmjs.com/package/pontoon-to-json)

PR #2120, opened 2016-08-22
[DO NOT MERGE] deploy test, branch to be discarded after feature branch deploy testing

Pomax:
testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "scripts": {
     "app": "npm run l10n && gulp app",
     "build": "gulp",
-    "l10n": "pontoon-to-json",
+    "l10n": "node pontoon-test.js",
     "cleanup": "rm -rf dist && npm run l10n",
     "preapp": "npm run cleanup",
     "prebuild": "npm run cleanup",
@@ -82,6 +82,8 @@
     "mozilla-tabzilla": "^0.5.1",
     "object-assign": "^4.1.0",
     "pontoon-to-json": "^1.0.0",
+    "properties-parser": "0.3.1",
+    "q-io": "1.13.2",
     "react": "^0.14.0",
     "react-addons-linked-state-mixin": "^0.14.3",
     "react-addons-pure-render-mixin": "^0.14.3",
--- a/pontoon-test.js
+++ b/pontoon-test.js
@@ -1,0 +1,124 @@
+var propertiesParser = require(`properties-parser`);
+var path = require(`path`);
+var FS = require(`q-io/fs`);
+var argv = require(`minimist`)(process.argv.slice(2));
+
+var habitat = require('habitat');
+habitat.load();
+var env = new habitat();
+
+var supportedLocales = process.env.SUPPORTED_LOCALES || `*`;
+
+if (supportedLocales.indexOf('[') !== -1) {
+  // FIXME: TODO: this circumvents a problem with habitat and
+  //              heroku, where the environment variable ['en-US']
+  //              magically becomes '['en-US']', which thus gets
+  //              turned into a String, rather than an Array...
+  //
+  // see: https://github.com/mozilla/learning.mozilla.org/pull/2120
+  //
+  supportedLocales = JSON.parse(supportedLocales.replace(/'/g,'"'));
+}
+
+var config = {
+  "dest": argv.dest || `dist`,
+  "src": argv.src || `locales`
+};
+
+function getListLocales() {
+  return new Promise((resolve, reject)=> {
+    if (supportedLocales === `*`) {
+      FS.listDirectoryTree(path.join(process.cwd(), config.src)).then((dirTree) => {
+        var localeList = [];
+
+        // Get rid of the top level, we're only interested with what's inside it
+        dirTree.splice(0,1);
+        dirTree.forEach((localeLocation) => {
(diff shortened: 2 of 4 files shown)


Pomax:
Build breaking heavily. Bisecting between current head and August 2nd.

Pomax:
august 2nd is good.

Pomax:
midpoint 43d1b3f is good, but has a test-react-warnings task that takes 40 seconds.

Pomax:
7c58fd9 is technically good, locally, but has a `test-react-warnings` task that takes 40 seconds to run.

Pomax:
b52f9fa is technically good, locally, but has a test-react-warnings task that takes 40 seconds to run.

Pomax:
24fdd9f is good, locally, but has an >1min `test-react-warnings` task

Pomax:
c96476a is good, locally, but has a 1m50 `test-react-warnings` task and very long timeout on the app smoketests.

Pomax:
2a84424  is good, locally, but has a 1m50 test-react-warnings task and very long timeout on the app smoketests.

Pomax:
da59ca4  is good, locally, but has a 1m50 test-react-warnings task and very long timeout on the app smoketests.

Pomax:
ffdd0aa is good, locally, but has a 1m50 test-react-warnings task and very long timeout on the app smoketests.

Pomax:
Locally, at initial test:
```
2d770b43eeb405a9cf6c5b6e7fec5a5aa557394a is the first bad commit
commit 2d770b43eeb405a9cf6c5b6e7fec5a5aa557394a
Author: Pomax <pomax@nihongoresources.com>
Date:   Mon Aug 22 16:11:13 2016 -0700

    test timeout rollbacks

:040000 040000 65995e8a82caad6e8005006a42baf946719bc218 0e4d8c9292242a4d1f4d00a5a17a978a8b4ab8f6 M      pages
:040000 040000 7dc35bf41adda7baaa874280528b44fa05d0bcac 4f2e7a22500eef200d37583c1be2b53eb6b9e778 M  test
```

Pomax:
One thing revealed during process.stdout writing is that the reason our app.js test takes so incredibly long is that it literally tests all links exposed by the router: if we have 10 locales, that's every URL routes.jsx generates, times ten, because `en-US` is a different set from `sv-SE` which is a different set from etc. 

/cc @alanmoo

alanmoo:
Yeah, that's what I was starting to address in #2108, but realized that alone wouldn't solve our problems.

Pomax:
Ah, had not seen that issue. We might have to address some of it first, in order to get the code back to at least "compiling" in CI... or roll back master, which I really hope we don't end up having to do.

alanmoo:
I did a git bisect and it was weird- it pointed at f557e05a7f91fd53a5a1a50f87e01a1820386061 being the bad commit, but #2102 passed testing

Pomax:
... tests pass O_O @alanmoo 

passing commit binds the indexStatic timeout we use in the tests to the actual indexStatic generator itself, and then we use that both in the "index static test" and "app test" in the tests that rebuild the index.

Pomax:
still deploy errors based on pontoon, though:
```
> learning.mozilla.org@1.0.5 l10n /tmp/build_3f18d5d3cd863446cfba23e1e11be412/mozilla-learning.mozilla.org-a56724c
       > pontoon-to-json
       
       processing the following locales: ['en-US']
       [TypeError: locales.map is not a function]
       module.js:327
       throw err;
```

Pomax:
`console.log(process.env)` after a habitat load shows:
```
SUPPORTED_LOCALES: '['en-US']',
```
which is definitely not an array

Pomax:
```
> learning.mozilla.org@1.0.5 l10n /tmp/build_e97904cf3bb69b9f6531a20f7e9ae803/mozilla-learning.mozilla.org-97c8b2f
       > node pontoon-test.js
       
       process.env: ['en-US'] : string
       habitat env: ['en-US'] : string
       processing the following locales: ['en-US']
       constructor: function String() { [native code] }
       [TypeError: locales.map is not a function]
       module.js:327
       throw err;
```

So much wrongness....

Pomax:
Sooooo it looks like habitat is not doing what we expect it to do, in combination with heroku.

On Heroku, we set the `SUPPORTED_LOCALE` variable to `['en-US']`, which you'd think is right, but habitat ends up loading it as the *String* `"['en-US']"`, so the way I'm getting this to work is to check for that in the pontoon-to-json code and do a manual `JSON.parse` pass if I see `[` in the locale string:

```
var habitat = require('habitat');
habitat.load();
var env = new habitat();

var supportedLocales = process.env.SUPPORTED_LOCALES || `*`;

if (supportedLocales.indexOf('[') !== -1) {
  // FIXME: TODO: this circumvents a problem with habitat and
  //              heroku, where the environment variable ['en-US']
  //              magically becomes '['en-US']', which thus gets
  //              turned into a String, rather than an Array...
  //
  // see: https://github.com/mozilla/learning.mozilla.org/pull/2120
  //
  supportedLocales = JSON.parse(supportedLocales.replace(/'/g,'"'));
}
```

Pomax:
the fixes from this exploratory issue have been landed in the appropriate places. closing

PR #2121, opened 2016-08-23
fix the build

Pomax:
- Bumps pontoon-to-json to 1.0.1 which has a fix for habitat not interpreting true environment variables (as opposed to vars read from file), and
- a fix that gives the indexStaticWatcher a timeout to be used by any test that relies on it (plus fixes to app.test.js, which relies on it).
- Also restricts the routes tested in app.test.js to only the `en-US` routes, to prevent a linear slowdown in testing whenever a new locale is added to our list of supported locales.


alanmoo:
🎉

PR #2125, opened 2016-08-23
add goggles to CSP whitelist

Pomax:
adds goggles to the whitelist so that goggles-remixing on our own property is permitted


Pomax:
closing in favour of CSP PR

PR #2126, opened 2016-08-23
eslint, with very conservative initial rules

Pomax:
removed the gulp jshint+jscs tasks, replaced them with an eslint npm task instead. Initial rules are only the semicolon and strict rules from mofo-style, so we can start iteratively getting the code in line with mofo-style without completely breaking the build.


.eslintrc
```diff
@@ -0,0 +1,20 @@
...
+    "ecmaFeatures": {
+      "jsx": true
+    },
+  },
+  "rules" : {
+    "semi": [2, "always"],
+    "strict": [2, "global"]
+  },
+  "extends": [
+    // "./node_modules/mofo-style/.eslintrc.yaml"
```
Pomax: one day soon, this will get uncommented, and it will be a good day.

package.json
```diff
@@ -17,9 +17,11 @@
   },
   "scripts": {
     "app": "npm run l10n && gulp app",
-    "build": "gulp",
-    "l10n": "pontoon-to-json",
+    "build": "npm run lint && gulp",
```
gvn: Having a test as part of the build task seems a bit weird to me. Won't Travis run the linting task twice this way?

package.json
```diff
@@ -17,9 +17,11 @@
...
     "cleanup": "rm -rf dist && npm run l10n",
+    "l10n": "pontoon-to-json",
+    "lint": "eslint --ext .js --ext .jsx .",
```
gvn: Why not make this `test:eslint`?

gvn: Then we can eventually add `npm-run-all` and just run all the test:** tasks for `test`.

.eslintrc
gvn: I guess this file is either JSON that allows comments or unvalidated JSON?

gvn:
Some thoughts on task naming, but LGTM!

package.json
Pomax: good idea.

package.json
Pomax: yeah I've taken this task out in one of the follow ups. There... isn't actually anything that uses it! \o/

.eslintrc
Pomax: it's super nice eslint "json but then with comments" (:heart:)

.eslintrc
```diff
@@ -0,0 +1,20 @@
+{
+  "parserOptions": {
+    "ecmaVersion": 6,
+    "sourceType": "module",
+    "ecmaFeatures": {
+      "jsx": true
+    },
```
alanmoo: This file doesn't pass eslint because of the comma here 😆 😜

package.json
```diff
@@ -28,7 +30,7 @@
     "postinstall": "npm run l10n && gulp postinstall",
     "smoketest": "gulp smoketest",
     "start": "npm run app",
-    "test": "gulp smoketest && npm run test:mocha && npm run test:browser && gulp lint-test",
+    "test": "npm run lint && gulp smoketest && npm run test:mocha && npm run test:browser",
```
alanmoo: Finally. 👍

alanmoo:
LGTM too

.eslintrc
Pomax: it's okay, it's not technically JSON... I think =D

Pomax:
closing as reviewed and superceded.

PR #2129, opened 2016-08-23
No more gulp

Pomax:
builds on https://github.com/mozilla/learning.mozilla.org/pull/2128

removes all traces of gulp from our codebase, and uses npm scripts for all the things gulp did before that we're keeping.

One thing we're not keeping is the `tests` dir with mocha and phantom tests, because they weren't actually testing anything particularly meaningful and were impeding improving the build system. We will have to introduce unit and integration tests once the indexStatic nonsense has been removed, too.

Due to the "builds on", the actual changes are in https://github.com/mozilla/learning.mozilla.org/pull/2129/commits/18e103f2860db06dfb3c200e467a34b5eb813f15


alanmoo:
I'm...going to need some coffee before I start this.

app.js
```diff
@@ -292,16 +273,27 @@ app.DIST_DIR = DIST_DIR;
 app.updateIndexStatic = updateIndexStatic;
 module.exports = app;
 
+// FIXME: TODO: clean up the whole indexStatic mess
```
alanmoo: Moved this down just to make it appear closer to where it's finally called?

package.json
```diff
@@ -53,16 +51,6 @@
     "esprima-harmony-jscs": "^1.1.0-bin",
     "express": "^4.12.3",
     "font-awesome": "^4.3.0",
-    "gulp": "^3.8.11",
```
alanmoo: 👀 🔥 🔥 🔥 😆

alanmoo:
Are some of these npm tasks going to eventually be consolidated into webpack?

Some of these tests were definitely useful, caught issues I had (especially re: routes), will they come back before this whole thing gets merged in?

app.js
Pomax: and easier to rip out in the next PR

Pomax:
short answer: yes - long answer: maybe, it depends on whether it's a thing we want to have automatically happen on build, or whether they're best kept separate for "stage build" vs "prod build" purposes, weighed against "integrated all the things during dev work so that a change gets linted/lessed/etc and shows up"

alanmoo:
Fair enough. This looks sane.

Pomax:
closing as reviewed and superceded.

PR #2133, opened 2016-08-24
Nostaticindex

Pomax:
Builds on https://github.com/mozilla/learning.mozilla.org/pull/2129

Actual changes for this PR in commit https://github.com/mozilla/learning.mozilla.org/pull/2133/commits/77546fdf14954dea84dd7598278e99025f0acad3

This PR removes the most problematic bits around the whole static site content generation, but for the moment keeps the two separate webpack tasks:

1. `./config/webpack/webpack.client.config.js` is the client-side bundle, building from entry point `./lib/build/client.bundle.jsx`.
2. `./config/webpack/webpack.server.config.js` is roughly the same thing, but compiled as commonjs2 format from `./lib/build/server.library.jsx`, so that app.js can load it in and use it for static page generation (renamed from `indexStatic` to `serverBundle` in app.js)

The client builds to `./dist/client.bundle.js`, and the server to `./build/server.library.js`.

(I tried doing this as a unified task using a "umd" target library but there was no victory there. After an hour of trying the way towards a clean dev run involved keeping the two compiles but simply making it obvious when and why they're used)

The `npm start` task has been rewritten to run the normally-run-once tasks for LESS compilation and bundle generation into watched tasks, so that changes to js/jsx/less files automatically trigger rebuilds. For the webpack compile tasks, the start script simply passes the `--watch` flag so that webpack goes into file watch mode, and for LESS I added `chokidar` to monitor the entire `./less` directory for file changes.

Additionally, `app.js` got a chokidar watch instruction that kicks in when app.js does not run with node in production mode. This instruction centers around monitoring the `./dist/index-static` dir for changes to the server-side bundle. If that bundle is updated, it causes app.js to rebind the bundle, and then rebind the `router` it uses internally for generating page data, so that changes to the source code don't just generate the latest and greatest code client-side, but also server-side.

The following npm scripts should do the following things:
- `npm test` (runs linting and compiles, both of which may flag errors)
- `npm run launch` (runs the compiles sequentially, then starts the app.js server, currently meant as procfile instruction, to allow for prod-env-vars despite promoting a slug)
- `npm start` (runs the compiles and app.js server in parallel, and in watch mode for all the things)

Also note that linting uses `eslint` and is done as part of the client-side bundling. I considered putting it in the server-side bundle as well but that would have made no sense: you'd see the same error twice. There is still a separate `npm run lint:fix`, which runs eslint with the `--fix` flag, which is useful while we get the codebase in line with the mofo-style linting rules.


config/image-convert.config.js
```diff
@@ -1,35 +0,0 @@
-// This file contains a `patterns` export that describes how
```
alanmoo: This is going away because of the copy scripts, right?

alanmoo:
There's *a lot* going on here, but it all looks pretty reasonable. The only thing I'd ask is, as you mentioned, is that some things get renamed to be clearer about what they do. I really like the server/client naming convention- indexStatic just doesn't seem obvious about its intent.

config/image-convert.config.js
Pomax: correct

Pomax:
man, +9001 to that. I'll be updating the PR commit to name some of these files more sensible, the "index static" will be something more along the lines of "render to html" so that it makes sense in the context of the technology we're using.

And any other files you think need a rename: let me know, now is the right time to do all of those.

alanmoo:
I'd say entry points for client/server should be appropriately named (it's hard to tell what "main.jsx" is doing at first glance), as well as the appropriate webpack configs and aforementioned index static. Basically I want to have some idea what a file is doing before opening it (even if it's a long name). :)

Pomax:
entry points have been renamed, webpack configs have been renamed, "indexStatic" var in app.js has been renamed:

1. client bundle has webpack config `./config/webpack/client.bundle.js`, with entry point `./lib/build/client.bundle.jsx`
2. server library bundle has webpack config `./config/webpack/server.library.js`, with entry point `./lib/build/server.library.jsx`
3. app.js now uses `serverBundle` rather than `indexStatic`, making it way more obvious what it's accessing.

lib/build/client.bundle.jsx
```diff
@@ -0,0 +1,62 @@
...
+      require.ensure(['intl'], function(require) {
+        window.Intl = require('intl');
+        startRunningSite();
+      }, "IntlBundle");
+    } else {
+      startRunningSite();
+    }
+  }
+
+  else if (GA_ACCOUNT) {
```
alanmoo: I know this is existing code, but how does this get executed if JS is disabled?

package.json
```diff
@@ -16,26 +16,34 @@
...
+    "copy-static-files": "run-s copy:img copy:vendor:*",
     "l10n": "pontoon-to-json",
+    "launch": "npm run prep && node app.js",
     "less": "lessc less/index.less dist/styles.css --source-map=dist/styles.css.map --clean-css=\"--s0\"",
-    "lint": "eslint --ext .js --ext .jsx .",
-    "lint:fix": "npm run lint -- --fix",
-    "prep": "run-s cleanup copy-static-files lint less l10n webpack",
-    "postinstall": "npm run l10n",
-    "start": "npm run app",
+    "lint:fix": "eslint --ext .js --ext .jsx . -- --fix",
```
alanmoo: Heh, that syntax.

alanmoo:
Is webpack outputting as compressed a file as it can for the client?

The webpack config file for the server is outputting a file with the same name as itself, just in a different folder? That seems confusing. How about we keep the webpack.config pattern? `webpack.client.config.js`/`webpack.server.config.js`?

Not that it matters too much on the server side, but the generated server file seems big. Is it including files that node could otherwise normally simply `require`?

Pomax:
the generated server file is just a plain file, no compression, no nothing. Though that also means it's not leaving `require` alone when it should, so we can bring that size down by telling it that react atc. are externals probably. I'm not too worried about "The server library size", but bringing the client size down is always good - far as I know we're using Uglify when in production mode: set your NODE_ENV to `production` and it should yield a much smaller file

alanmoo:
You're right about the env. Do some renaming and this is good by me.

lib/build/client.bundle.jsx
Pomax: this is an excellent question....

Pomax: worth follow-up-ing on

.eslintrc
```diff
@@ -0,0 +1,20 @@
...
+    "ecmaFeatures": {
+      "jsx": true
+    },
+  },
+  "rules" : {
+    "semi": [2, "always"],
+    "strict": [2, "global"]
+  },
+  "extends": [
+    // "./node_modules/mofo-style/.eslintrc.yaml"
```
gvn: Comment in JSON?

package.json
```diff
@@ -41,45 +52,37 @@
...
-    "gulp-if": "^2.0.0",
-    "gulp-jscs": "^3.0.2",
-    "gulp-jshint": "^2.0.0",
-    "gulp-less": "^3.0.1",
-    "gulp-minify-css": "^1.0.0",
-    "gulp-plumber": "^1.0.1",
-    "gulp-rename": "^1.2.0",
-    "gulp-sitemap": "^2.0.1",
-    "gulp-sourcemaps": "^1.5.0",
-    "gulp-util": "^3.0.7",
```
gvn: :metal:

.eslintrc
Pomax: .eslintrc is technically not json (that would be `eslintrc.json`). The .eslintrc file is deprecated but given that we want to move the code to mofo-style, as a temporary file this seemed fine.

gvn:
It would be nice to log out a message like `App running at http://localhost:8008` when you run `npm start` and its ready for business. Right now it's hard to tell that it's done.

Pomax:
Yeah I'm not entirely sure how to make that part work. the `--watch` tasks mean that you don't get the normal "done" signal, so you can't strictly speaking "wait" for anything. Happy to get suggestions on how to tackle that problem

gvn:
LGTM! 🍰

Pomax:
filed https://github.com/mozilla/learning.mozilla.org/issues/2139 as followup for the "is app.js ready?" issue

PR #2141, opened 2016-08-29
Form field names

### LINKED ISSUE #2100: Update club form to use "name" instead of clubname and "owner" instead of name
There is only one "Name" field showing on the admin side of the applications, but 2 name fields in the application itself: "Name" and "Name your Club" I'm seeing applications with the name field listing both Club names and Club Captain names (either/or) on the admin side.

Can someone double check which name field is coming through on the admin site? If it is the "Name your club" field, perhaps we should rename to be "Club name" so those that are only reading the 1st word understand what we are looking for.

@hannahkane

Pomax:
fixes #2100 

This aligns the club application form with the fields as used in the teach api:

- `owner` is the user's name
- `name` is the club name

The application was transmitting this information with the field names `name` and `clubName` respectively.

To test, fill in the application form and once submitted look at the clubs list, which should now show the club name rather than the owner name. (to see the wrong thing, try this in master and note that the user's name ends up as if it's club name)

### DIFF
--- a/pages/clubs-list/ClubListItem.jsx
+++ b/pages/clubs-list/ClubListItem.jsx
@@ -36,7 +36,7 @@ var ClubListItem = React.createClass({
         <h4>{clubName} <Map.ClubStatusLabel showApproved={isOwned} status={club.status}/></h4>
         <p><span className="club-location" onClick={this.handleLocationClick}><span className="glyphicon glyphicon-map-marker"/> {club.location.split(',')[0]}</span></p>
         <p>{club.description}</p>
-        <p><small>Led by {club.owner}</small></p>
+        <p><small>Led by {club.full_name || club.owner}</small></p>
         {ownerControls}
       </li>
     );
--- a/pages/clubs/StepOne.jsx
+++ b/pages/clubs/StepOne.jsx
@@ -4,7 +4,7 @@ var LocationSelector = require('../../components/LocationSelector.jsx');
 var Select = require('react-select');
 
 var progressFields = [
-  "name",
+  "owner",
   "location",
   "occupation",
   "regionalCoordinator",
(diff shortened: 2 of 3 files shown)


gideonthomas:
Hmm @Pomax, I tried this with the old code and it worked. Would you be able to rebase and I can test again? (Sorry for the delay on this)

gideonthomas:
The validation for step 2 needs to be updated to use `name` instead of `clubName` since I can't submit the form.

Pomax:
@gideonthomas do you get validation error redness and "please fill in a clubname"? Or console errors?

gideonthomas:
ah sorry, I get "please fill in a clubname"

Pomax:
@gideonthomas replaced the last vestiges of "clubName" with "name", should work now?

KGorr:
I was catching up on applications and tested the process. My name showed up on Club name field instead of the club name.

Pomax:
@KGorr you're commenting on the in-progress patch, it won't land until it has been okayed in review and merged into the master branch =)

(then it'll be live on learning.mofostaging.net, and then once we do a production push it'll also be live on learning.mozilla.org)

gideonthomas:
@Pomax this works for me! Although, one thing I noticed was that learning.mofostaging.net (which should be master right?) does not have this issue? So I'm not sure where this issue actually exists

KGorr:
Sorry and thanks @Pomax!

Pomax:
dang, there's actually another problem with this patch - `owner` gets autoset to the login id, not the full name people fill in, so this needs an additional tweak to StepOne to make sure it passes a different fieldname for the full name of the application.

@gideonthomas you mean with this patch but hooked up to the mofostaging database, or just pure test on mofostaging?

gideonthomas:
pure test on mofostaging

Pomax:
I just tested this on mofo staging logged in as "monkeymagic":

My full name: `catweasel`
My club name `get catweasel home club`

When I submit this form, the teach-api-staging instance shows this in its records:

owner: `monkeymagic`
club name: `catweasel`

Pomax:
@gideonthomas rereview?

PR #2142, opened 2016-08-30
App post compile

Pomax:
Improved the server startup and some of its internal organization, and added eslinting for the sever code (because it's worth linting all the code we write =)

Rather than running the webpack tasks in --watch mode and starting app.js at the same time, the new run simply does a sequential run of the two webpack compiles, then runs app.js, then runs the webpack tasks in --watch mode. This is functionally equivalent as the original except that it allows us to pick the exact time to start running app.js, solving the issue of required files possibly not existing, or being in a stale state.


gvn:
Nice! WFM!

PR #2143, opened 2016-08-30
webpack-loader for detecting undocumented environment variables

Pomax:
This PR adds a webpack loader that runs as part of both the client and server webpack compiles to see if there is any mention of environment variables that are not documented in the README.md.

This is, by its very nature, a crude check and as such will not cause `npm test` (or the individual webpack tasks) to break when an undocumented var is found, it does show a nice, coloured warning for you to act on.


Pomax:
Running this finds two undocumented environment variables during the server compile for me. So that's a nice instant result.

PR #2145, opened 2016-08-30
left-over mentions of teach.moz that should have been made learning.moz

Pomax:
There were quite a few instances of `teach.mozilla.org` that should now be `learning.mozilla.org`, however there are some instances that cannot be replaced.

@hannahkane: there are links to logoassets.jsx and tabs.jsx that point to `https://stuff.webmaker.org/teach.mozilla.org` resources, renaming of which requires we actually change links on the web. Do we want to do that, or do we just keep those URLs and have some historical baggage in our files?


hannahkane:
Would rather not change the links. I'm OK with the baggage if you are. :)

Pomax:
fine by me (saves us having to deal with the redirects if we want to make sure the old links still point to something =)

PR #2148, opened 2016-08-31
Eslint rules

Pomax:
So many rules! Each commit is one or more rules with the associated fixes to make eslint pass again.

Some of these were solved with `npm run lint:fix` (which runs `eslint --fix`), some by hand. Did you know `eslint --fix` can't fix the `newline-after-var` rule, for instance? That was fun... =P


.eslintrc
```diff
@@ -7,8 +7,45 @@
     },
   },
   "rules" : {
-    "semi": [2, "always"],
-    "strict": [2, "global"]
+    "valid-jsdoc": [1],
```
alanmoo: Is this a copy/paste from mofo-style until you extend mofo-style completely?

server/render-component-page.js
```diff
@@ -1,8 +1,13 @@
...
+ * @param {object} generator the server page generator
+ * @param {string} location the URL for this request
+ * @param {object} res express.js response object 
+ * @param {string} locale the locale/language for this page
+ * @returns {undefined}
  */
-module.exports = function renderComponentPage(serverBundle, location, res, locale) {
-  serverBundle.generate(location, {locale: locale}, function(err, location, title, html) {
+module.exports = function renderComponentPage(generator, location, res, locale) {
+  generator.generate(location, {locale: locale}, function(err, _location, title, html) {
```
alanmoo: Changed this back?

.eslintrc
Pomax: correct. Once all of them are in, these go away and the external link goes live.

server/render-component-page.js
Pomax: I'd prefer not to - the javadoc rule made it obvious that this argument, at least for this function, not its called, should talk about generators, not "whatever that generator was loaded as".

alanmoo: That's fine, I don't have an opinion just wanted to make sure it was intentional among the rest of the syntactical changes.

PR #2152, opened 2016-08-31
source mapping for the server.library.js bundle

Pomax:
Super handy when trying to figure out what on earth is going on when there's a server error rather than client error.

This now generates a `build/server.library.js` file that has all the source mapping to tell you that an error occurred in, for instance, page.jsx line 162, rather than in server.library.js line something-thousand-something-hundred pointing to the start of a massive bundle entry.


alanmoo:
WFM

PR #2164, opened 2016-09-01
Reintroducing mocha testing

Pomax:
This sets up mocha testing again, with an initial app.test.js that simply runs our server, and checks whether `/en-US/` and `/en-US/healthcheck` exist, by requesting them using `superagent` and checking what their titles are.


gvn:
LGTM!

PR #2165, opened 2016-09-02
Stub out UI and l10n in disabled URL for maker party 2016 landing pages.

ScottDowne:
- [x] Localization addressed

Fixes # https://github.com/MozillaFoundation/Advocacy/issues/81

- Stub our UI for maker party 2016 landing pages
- Add in l10n strings for maker party 2016 landing pages
- Hide maker party pages, until they are translated, and we're ready to ship maker party
- Needs to be on staging, and enabled, in pontoon, but not live.
- It can go live if needs be, it's hidden.
- It can have further QA from staging, before we go live.
- Going live mid Sept.
- Before we go live, we need to integrate with disabling the nav for certain languages.


ScottDowne:
I found at least one link to `/events`. So I'll go through and remove or disable any others.

alanmoo:
This just removes the events page from the sidebar?

Issues: (will edit and add more as found)

- [ ] Tab link text are getting link styling

alanmoo:
As a vehicle for getting this onto staging so that Pontoon can start handling it, this is good to merge. Open another issue for QA of these pages though and put in those 2 issues from the previous comment?

The real code review on this will come when replacing the existing events pages.

alanmoo:
When do you need this to land? After today I'm out for the rest of the week.

ScottDowne:
I dunno at this point. It just had copy and UI changes last night. So this week might be pushing it.

Take PTO, and it is what it is :)

ScottDowne:
@alicoding There were copy and UI changes to this today, so I needed to update it.

Alan is on PTO now, and there is hope above me that it can be on staging for tomorrow at around noon, if you could help review this tomorrow by before then, that would be great. Otherwise, I'll need to find another reviewer.

The links to these pages are now disabled, so it's not live, so the risk is low, and there might be small style changes, and links need to be filled in, but otherwise, the context and strings are staying the same.

components/events/events-header.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import HeroUnit from '../../components/hero-unit.jsx';
+import { FormattedHTMLMessage } from 'react-intl';
+
+var EventsHeader = React.createClass({
```
ScottDowne: This file is the hero unit for both pages, which have the same hero unit.

components/events/events-nav.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { Link } from 'react-router';
+
+var EventsNav = React.createClass({
```
ScottDowne: This file is the psudo nav, inside the hero unit, to switch between the two pages.

I was going to use mofo-ui but I needed routes for this too, so I'm just using the css for the moment.

components/sidebar/topLevelNavItems.jsx
```diff
@@ -32,10 +32,12 @@ var MENU_ENTRIES = [
         name: "mozilla_clubs",
         to: '/clubs'
       },
+      /*
```
ScottDowne: This can be enabled again when we ship.

less/pages/event-resources.less
```diff
@@ -42,6 +42,27 @@ html.no-js {
       color: @darkBlueSidebarColor;
     }
   }
+  ol {
+    counter-reset: li; /* Initiate a counter */
```
ScottDowne: All the css is to customize some numbers in an `<ol>` element.

less/pages/events.less
```diff
@@ -1,9 +1,12 @@
-.page.events .hero-unit {
-  background-image: url(/img/pages/events/hero-unit.jpg);
+.page.events,
+.page.event-resources {
```
ScottDowne: Just updating the hero unit for both pages.

pages/event-resources/EventDetail.jsx
```diff
@@ -1,11 +1,20 @@
 var React = require('react');
 
+import { FormattedHTMLMessage } from 'react-intl';
```
ScottDowne: I need to remove this file.

components/events/events-header.jsx
```diff
@@ -0,0 +1,37 @@
...
+        <div>
+          <a href="#" className="btn">
+            {this.context.intl.formatMessage({id: 'get_started'})}
+          </a>
+        </div>
+        <div>
+          <FormattedHTMLMessage
+            id="already_have_event"
+            values={{
+              login: "#"
```
alicoding: Why passing #?

alicoding:
only had one issue. Otherwise this looks good.

ScottDowne:
I'm passing # as the login link cause maker party on nation builder doesn't
have a login link yet cause it's not built yet.

alicoding:
@ScottDowne yeah, but why not just put it in the `.properties` file?

ScottDowne:
Oh I see. Yeah it really doesn't need to be a variable. I can move it to
props.

ScottDowne:
@alicoding updated, if this still looks good, I'll rebase, squash, and merge?

I want to pull out a chunk and store it before I rebase because there is something useful I built then removed, which I don't want to vanish in the rebase.

alicoding:
Good to go!

PR #2168, opened 2016-09-02
Internal link crawling

Pomax:
Reintroduces a web spider that can crawl the site looking for out-of-expectation links (e.g. links where the locale goes missing, or links that don't actually work).

It currently warns on the locale going missing, so `npm test` finishes, but it will error on 404s.

Builds on #2164 (mocha testing) so the actual spider code is in https://github.com/mozilla/learning.mozilla.org/pull/2168/commits/faa15f03f57c2f4825d14d9f424c1e84cdc12775


spider.js
```diff
@@ -0,0 +1,138 @@
+var request = require('superagent');
+var url = require('url');
+var jsdom = require('jsdom');
+var app = require('./app');
+var server;
+var serveraddress = "http://localhost:" + app.PORT;
```
gvn: Why not use template strings?

gvn:
A couple nits, but seems to work. 👍

spider.js
Pomax: "not enough to template" really. It's only one value and it's at the end instead of in the middle, so templating seemed less readable for exactly the same intent

PR #2191, opened 2016-09-08
fixed mapping warning in Running.jsx

Pomax:
fixes https://github.com/mozilla/learning.mozilla.org/issues/2190 by refactoring the nested nest of nesting and assigning keys to the elements returned by each function that is involved in element mapping

### DIFF
--- a/pages/clubs/Running.jsx
+++ b/pages/clubs/Running.jsx
@@ -43,40 +43,40 @@ var Running = React.createClass({
   componentDidMount: function() {
     this.fetchJSON("https://mozilla.github.io/learning-networks/clubs/clubs-resources.json",this.setData);
   },
-  getLinks : function(){
-    var categories = this.state.guides.map(function(category){
-      var guideLinks = category.guides.map(function(guide){
-
-        if(guide.translations && guide.translations.length > 0){
-          var translationLinks = guide.translations.map(function(translation){
-            return (
-              <li>
-                <a href={ translation.url }>{ translation.title }</a>
-                <span className="language"> ({ translation.language })</span>
-              </li>
-            );
-          });
-        }
-
+  getLinkForGuide: function(guide) {
+    if(guide.translations && guide.translations.length > 0){
+      var translationLinks = guide.translations.map(function(translation){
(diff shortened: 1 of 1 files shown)


pages/clubs/Running.jsx
```diff
@@ -43,40 +43,40 @@ var Running = React.createClass({
...
+      <section className="resourceCategory" key={cat}>
+        <h3> { cat } </h3>
+        <ul>
+          { guideLinks }
+        </ul>
+      </section>
+    );
+  },
+  getGuideList: function() {
+    return this.state.guides.map(this.getGuideLinks);
```
gvn: Since each of these 3 new methods in only called once, this feels a bit over-abstracted. 

Any reason to not just do it as 1 (or maybe 2) methods?

pages/clubs/Running.jsx
```diff
@@ -43,40 +43,40 @@ var Running = React.createClass({
...
     this.fetchJSON("https://mozilla.github.io/learning-networks/clubs/clubs-resources.json",this.setData);
   },
-  getLinks : function(){
-    var categories = this.state.guides.map(function(category){
-      var guideLinks = category.guides.map(function(guide){
-
-        if(guide.translations && guide.translations.length > 0){
-          var translationLinks = guide.translations.map(function(translation){
-            return (
-              <li>
```
gvn: Couldn't this issue be fixed just by adding a key here?

gvn:
Seems a tad overcomplicated, but R+ if you feel strongly about your approach. 👍

pages/clubs/Running.jsx
Pomax: two reasons really, it was a massive map-map-map block and poorly named, and by doing it this way the `this` context is preserved due to React policy rather than possibly getting switch to something else. I'd rather have it a little over abstracted than three nested maps.

pages/clubs/Running.jsx
Pomax: each map needs a key, so both functions need a `key` in their return element (we had three spots)

PR #2195, opened 2016-09-09
Use Django fo Clubs Guides

### LINKED ISSUE #2012: Create prototype for managing clubs guides in Django
This is an exploration of using Django to manage(list, add, delete, update) clubs guides and see if it will work as a long term solution. Hence, this should primarily be treated as a tracking issue for the relevant teach-api bugs as well as a place to note our findings.

The outcomes of this exploration are:
- [x] Figure out if we can design an appropriate API to get/add/update/delete the clubs guides
- [x] Figure out if the administrative interface is good as it is for adding new guides
- [ ] ~~An initial plan on how we manage user permissions for accessing these resources~~
- [x] Figure out what changes need to be made to the administrative tool to be more user friendly

gideonthomas:
Fixes #2012

### DIFF
--- a/lib/teach-api.js
+++ b/lib/teach-api.js
@@ -230,6 +230,12 @@ _.extend(TeachAPI.prototype, {
         ga.event({ category: 'Clubs', action: 'Deleted a Club' });
         callback(null);
       }.bind(this));
+  },
+  getClubsGuides: function(callback) {
+    callback = callback || function () {};
+    return this.request('get', '/api/clubsguides/categories?expand=clubsguides')
+      .accept('json')
+      .end(callback);
   }
 });
 
--- a/lib/teach-api.js
+++ b/lib/teach-api.js
@@ -231,8 +231,7 @@ _.extend(TeachAPI.prototype, {
         callback(null);
       }.bind(this));
   },
-  getClubsGuides: function(callback) {
-    callback = callback || function () {};
+  getClubsGuides: function(callback=() => {}) {
     return this.request('get', '/api/clubsguides/categories?expand=clubsguides')
       .accept('json')
       .end(callback);
--- a/pages/clubs/Running.jsx
+++ b/pages/clubs/Running.jsx
@@ -12,7 +12,7 @@ var Running = React.createClass({
     intl: React.PropTypes.object
   },
   componentDidMount: function() {
-    this.props.teachAPI.getClubsGuides(function(err, res){
+    this.props.teachAPI.getClubsGuides((err, res) => {
       if(err) {
         this.setState({error: err});
         return;
@@ -20,19 +20,20 @@ var Running = React.createClass({
 
       var data = res.body;
 
-      for(var i = 0; i < data.length; i++){
-        var categoryLabel = data[i].category.toLowerCase();
+      data.forEach(dataItem => {
+        var categoryLabel = dataItem.category.toLowerCase();
 
         categoryLabel = "guide_category_" + categoryLabel.split(' ').join('_');
 
         var localizedCategory = this.context.intl.formatMessage({id: categoryLabel});
 
         if(localizedCategory !== categoryLabel){
-          data[i].category = localizedCategory;
+          dataItem.category = localizedCategory;
         }
-      }
+      });
+
       this.setState({guides : data });
-    }.bind(this));
+    });
   },
   getLinkForGuide: function(guide) {
     if(guide.translations && guide.translations.length > 0){
--- a/pages/clubs/Running.jsx
+++ b/pages/clubs/Running.jsx
@@ -20,6 +20,11 @@ var Running = React.createClass({
 
       var data = res.body;
 
+      if(!Array.isArray(data)) {
+        this.setState({error: "Empty or invalid response body returned!"});
+        return;
+      }
+
       data.forEach(dataItem => {
         var categoryLabel = dataItem.category.toLowerCase();
 
--- a/pages/clubs/Running.jsx
+++ b/pages/clubs/Running.jsx
@@ -1,5 +1,7 @@
 var React = require('react');
 
+var withTeachAPI = require('../../hoc/with-teach-api.jsx');
+
 var Running = React.createClass({
   getInitialState: function(){
     return {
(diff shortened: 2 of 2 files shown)


pages/clubs/Running.jsx
```diff
@@ -41,7 +41,7 @@ var Running = React.createClass({
     httpRequest.send();
   },
   componentDidMount: function() {
-    this.fetchJSON("https://mozilla.github.io/learning-networks/clubs/clubs-resources.json",this.setData);
+    this.fetchJSON("https://teach-api-staging.herokuapp.com/api/clubsguides/categories?expand=clubsguides",this.setData);
```
Pomax: `config.TEACH_API_URL`?

gideonthomas: Ah neat! I'll fix that when I'm actually going to merge this in since for now it's just for QA

pages/clubs/Running.jsx
```diff
@@ -41,7 +44,7 @@ var Running = React.createClass({
     httpRequest.send();
   },
   componentDidMount: function() {
-    this.fetchJSON("https://mozilla.github.io/learning-networks/clubs/clubs-resources.json",this.setData);
+    this.fetchJSON(`${TEACH_API_URL}/api/clubsguides/categories?expand=clubsguides`,this.setData);
```
Pomax: why is this not just a `getClubsGuides` function in the teach-api.js library? That's what the library is for =)

pages/clubs/Running.jsx
```diff
@@ -1,5 +1,8 @@
 var React = require('react');
 
+var config = require('../../config/config');
+const TEACH_API_URL = config.TEACH_API_URL.replace(/\/?$/, '');
```
Pomax: this should tap into the teach-api (which can be passed into component from a parent)

config/config.js
```diff
@@ -35,6 +35,7 @@ exports.LIGHT_BEAM_URL = "https://addons.mozilla.org/firefox/downloads/latest/36
 exports.MAKE_METADATA_URL = process.env.MAKE_METADATA_URL || 'https://{username}.makes.org/makes.json';
 exports.MOZFEST_SITE_LINK = 'https://mozillafestival.org';
 exports.ORIGIN = ORIGIN;
+exports.TEACH_API_URL = process.env.TEACH_API_URL || 'https://teach-api-staging.herokuapp.com';
```
Pomax: if we're fixing the value here, we should make `teach-api.js` cap into `config.js` too for its teach api URL

pages/clubs/Running.jsx
```diff
@@ -9,39 +11,28 @@ var Running = React.createClass({
   contextTypes: {
     intl: React.PropTypes.object
   },
-  setData : function(data){
-    for(var i = 0; i < data.length; i++){
-      var categoryLabel = data[i].category.toLowerCase();
+  componentDidMount: function() {
+    this.props.teachAPI.getClubsGuides(function(err, res){
```
Pomax: protip: make this a component function, so `this.props.teachAPI.getClubsGuides(this.processClubsGuides)`

pages/clubs/Running.jsx
```diff
@@ -9,39 +11,28 @@ var Running = React.createClass({
...
         }
       }
-    };
-    httpRequest.open('GET', path);
-    httpRequest.send();
-  },
-  componentDidMount: function() {
-    this.fetchJSON("https://mozilla.github.io/learning-networks/clubs/clubs-resources.json",this.setData);
+      this.setState({guides : data });
+    }.bind(this));
```
Pomax: we support ES6 syntax, so if you need `bind`, you don't need bind, you need an ES6 arrow function =)

lib/teach-api.js
```diff
@@ -230,6 +230,12 @@ _.extend(TeachAPI.prototype, {
         ga.event({ category: 'Clubs', action: 'Deleted a Club' });
         callback(null);
       }.bind(this));
+  },
+  getClubsGuides: function(callback) {
+    callback = callback || function () {};
```
Pomax: we support ES6, so `callback = callback || ()=>{};` will also work

pages/clubs/Running.jsx
```diff
@@ -9,39 +11,28 @@ var Running = React.createClass({
...
-      var categoryLabel = data[i].category.toLowerCase();
+  componentDidMount: function() {
+    this.props.teachAPI.getClubsGuides(function(err, res){
+      if(err) {
+        this.setState({error: err});
+        return;
+      }
 
-      categoryLabel = "guide_category_" + categoryLabel.split(' ').join('_');
+      var data = res.body;
```
Pomax: this looks like it might need some kind of "is there actual data, if not, error"?

pages/clubs/Running.jsx
```diff
@@ -9,39 +11,28 @@ var Running = React.createClass({
...
 
-      var localizedCategory = this.context.intl.formatMessage({id: categoryLabel});
+      for(var i = 0; i < data.length; i++){
```
Pomax: foreach for the win:
```
data.forEach(entry => {
  let categoryLabel = entry.categoryLabel;
  ...etc...
});
```

pages/clubs/Running.jsx
gideonthomas: It was like that before but the logic was really small so thought it wasn't necessary

pages/clubs/Running.jsx
gideonthomas: Does super-agent check that the status code is 200?

lib/teach-api.js
gideonthomas: The entire file does this though (literally copy/pasted from the other functions) so I think we should convert the entire file to ES6 in a separate patch

Pomax: we introduced ES6 well after the entire codebase was set up, at this point it's easier to update the bits we touch to ES6 as we touch them rather than converting entire files at a time.

pages/clubs/Running.jsx
Pomax: even if the status code is 200, that does not mean the body has anything in it. You can have an empty document response despite HTTP200

pages/clubs/Running.jsx
```diff
@@ -9,39 +11,29 @@ var Running = React.createClass({
...
 
-      if(localizedCategory !== categoryLabel){
-        data[i].category = localizedCategory;
-      }
-    }
-    this.setState({guides : data });
-  },
-  fetchJSON : function(path,callback){
-    var httpRequest = new XMLHttpRequest();
+        categoryLabel = "guide_category_" + categoryLabel.split(' ').join('_');
```
Pomax: this seems extremely fragile, where is this label format coming from? (can we somehow tap into whatever dictates this format?)

gideonthomas:
r+ over chat

PR #2198, opened 2016-09-09
remove the JS feature detection

Pomax:
fixes https://github.com/mozilla/learning.mozilla.org/issues/2138

Removes the "ENABLE_JS" flag and simply assumes people know what they're doing. Most people will have javascript enabled anyway, and those that don't do so intentionally (and so have the agency to decide whether to keep running JS-less or enable it on our site).

Previous this flag would add a "you don't have JS enabled, please do a thing" bar at the top of the website, but the way it decided what to load still used JS, so it was more a limited feature detection that falsely concluded JS would not be available.


gvn:
LGTM!

No JS?
riker

Pomax:
hahaha so good

PR #2201, opened 2016-09-12
quick fix to maker party tabs

ScottDowne:
@gvn so Initially, I needed a tab navigation functionality, that looked like the clubs nav.

I didn't see the clubs nav working with routes, so I just did the quickest thing (we needed it on pontoon asap, so the link needed to function, not look good) which was just some `<Link>` elements and stole some css. Seemed to work fine, but was fragile.

Not surprisingly the events nav broke. I noticed there were some changes to the classNames in way mofo-ui tab switcher, which is fine, and I updated the classNames in my quick implementation, and that's all this PR is addressing. The Nav looks good again.

However, I did notice that the clubs page seems to be using routes in the nav somehow, and I think I need a more stable solution for my nav so it doesn't break again. Thoughts I have or ideas for a future patch would be:

1. Are you open to changes to the clubs nav UI to just be some css and some `<Link>` elements? Or is there something a little more complex going on in the `TabSwitcher` that prevents this?
2. I update my page to be more like clubs, but I'm actually confused as to how the `TabSwitcher` works.
3. Do nothing. Maybe you don't plan on updating the css in the `TabSwitcher`, but this is unlikely, and not a great option.
4. I freeze the css in another css file so if you make changes to `TabSwitcher` the maker party page stay the same. Also not a great option.

I like options 1 and 2 best. 3 and 4 are hacks. Thoughts?

### DIFF
--- a/components/events/events-nav.jsx
+++ b/components/events/events-nav.jsx
@@ -10,11 +10,11 @@ var EventsNav = React.createClass({
     return (
       <div className="inner-container">
         <div className="mui-tab-switcher">
-          <div className="tabs">
-            <Link className="btn" activeClassName="active" to={"/" + this.context.intl.locale + "/events"}>
+          <div className="mui-tabs">
+            <Link className="mui-btn" activeClassName="mui-active" to={"/" + this.context.intl.locale + "/events"}>
               {this.context.intl.formatMessage({id: 'overview'})}
             </Link>
-            <Link className="btn" activeClassName="active" to={"/" + this.context.intl.locale + "/events/resources"}>
+            <Link className="mui-btn" activeClassName="mui-active" to={"/" + this.context.intl.locale + "/events/resources"}>
               {this.context.intl.formatMessage({id: 'host_resources'})}
             </Link>
           </div>
--- a/less/pages/events.less
+++ b/less/pages/events.less
@@ -10,6 +10,11 @@
   .secondary-button {
     display: inline-block;
   }
+  font-size: 18px;
+  .mui-btn {
+    line-height: inherit;
+    text-decoration: none;
+  }
 }
 
 .events {


gvn:
I don't really want to refactor the clubs tabs because it is working as intended.

I don't think the CSS will change much at this point, and I will try my best to avoid any class name changes. Now that we're namespacing the classes in Mofo UI we shouldn't have to change them again. There was a conflict without the namespaces.

Are there questions you have that I can answer about the Tabswitcher component? I realize that we need to add individual component docs to the repo.

On LMO I'm using React Router programmatically and external to the Tabswitcher (via callbacks & public API) to avoid having a RR dependency within MUI. I'm trying to keep MUI with as few deps as possible...

ScottDowne:
I would volunteer to refactor the clubs page if switching the TabSwitcher to static css and putting it inside mofo-bootstrap is a solution you can get behind?

For the interim then, it sounds like you don't have any issues with the current patch as is, based on your answer.

gvn:
No, I don't want to refactor TS to static CSS. If it's unclear how to use it then let's add documentation.

This patch approach is ok, but you need to fix the style a bit since you're using `A` elements under the hood instead of buttons and there's some inconsistencies because of that (font size, line height, shouldn't underline on hover).

Do a side-by-side comparison of your page and `/clubs` and you can see the differences.

ScottDowne:
Why `button` over `a`?

We lose the browser link targeting of an `a` with `button`. Also cannot right click, open in new tab, etc.

I'm using the default `Link` with is appropriately an `a`

I was OK with the slight styling difference, but would you be OK if I just restyled the anchor to match the button side by side?

gvn:
> I was OK with the slight styling difference, but would you be OK if I just restyled the anchor to match the button side by side?

Yes, I am suggesting that you update your CSS to fix the inconsistencies and leave everything else as is.

> Why button over a?

It's a button because tabs are not *always* associated with a URL. That's a choice you make when implementing them in a project. A page could have multiple tabswitchers within it that you might not want explicit URLs for every state of.

ScottDowne:
Ah, I misunderstood, and was reading it as you wanting me to switch to buttons because the anchor looked different. My bad, I can restyle to match.

ScottDowne:
@gvn updated. R?

gvn:
Just one tweak: set line-height on the buttons to `inherit` as it's a tad off still.

After that: R+!

ScottDowne:
Done, and seeing how you gave me a green, merging!

PR #2203, opened 2016-09-12
twitter and facebook plain-link sharing

Pomax:
fixes https://github.com/mozilla/learning.mozilla.org/issues/2154 although for the moment this PR stubs out the share links but we need to:

- [ ] style these to look nice (fb/tw icons?)
- [ ] tweak the copy to be appropriate

### DIFF
--- a/pages/badges/badge-single.jsx
+++ b/pages/badges/badge-single.jsx
@@ -234,16 +234,26 @@ var BadgePage = React.createClass({
     );
   },
 
+  getShareCodes: function() {
+    var url = encodeURIComponent(window.location.toString());
+    var msg = encodeURIComponent(`I earned the Mozilla ${this.state.badge.title} badge`);
+
+    return (
+      <div className="social-share">
+        <a href={`https://twitter.com/home?status=${msg + encodeURIComponent(': ') + url}`} target={'_blank'}>facebook</a>
+        <a href={`http://www.facebook.com/sharer.php?u=${url}&t=${msg}`} target={'_blank'}>twitter</a>
+      </div>
+    );
+  },
+
   renderAchieved: function() {
     var badgeCriteria = this.formBadgeCriteria(this.state.badge.criteria);
-    var share = null; // <SocialShare />
 
-    // FIXME: TODO: retrieve the information on when/how this badge was earned.
-    //              ... IF we use this information at all.
+    // FIXME: TODO: retrieve the information on when/how this badge was earned... IF we use this information at all.
     return (
       <div className="badge-achieved">
         <h3 className={'text-light'}>Congrats, you were awarded this credential.</h3>
-        { share }
+        { this.getShareCodes() }
         <div className="badge-reward-text">
           <div className="date">
             DATE FROM API (although we may not end up using this);
@@ -275,7 +285,6 @@ var BadgePage = React.createClass({
       </div>
     );
   },
-
 
   renderEligible: function() {
     var badgeCriteria = this.formBadgeCriteria(this.state.badge.criteria);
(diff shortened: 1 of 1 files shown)


pages/badges/badge-single.jsx
```diff
@@ -234,16 +234,26 @@ var BadgePage = React.createClass({
     );
   },
 
+  getShareCodes: function() {
+    var url = encodeURIComponent(window.location.toString());
+    var msg = encodeURIComponent(`I earned the Mozilla ${this.state.badge.title} badge`);
```
Pomax: @hannahkane do we have some preferred copy?

pages/badges/badge-single.jsx
```diff
@@ -234,16 +234,26 @@ var BadgePage = React.createClass({
...
+
+    return (
+      <div className="social-share">
+        <a href={`https://twitter.com/home?status=${msg + encodeURIComponent(': ') + url}`} target={'_blank'}>facebook</a>
```
gideonthomas: Text should be `twitter`

pages/badges/badge-single.jsx
```diff
@@ -234,16 +234,26 @@ var BadgePage = React.createClass({
...
+        <a href={`http://www.facebook.com/sharer.php?u=${url}&t=${msg}`} target={'_blank'}>twitter</a>
```
gideonthomas: Text should be `facebook`

gideonthomas: Also, do you know if they allow sharing through SSL? i.e. https://www.facebook... vs. http?

pages/badges/badge-single.jsx
hannahkane: @anmechung - any thoughts on copy? This is for the "Share" copy on Badges, where people will be able to post on Twitter and Facebook. Right now we have "I earned the Mozilla [badge name] badge." I'd suggest something like, "I just earned the Mozilla [badge name] Badge! Learn more about Mozilla Badges [link to /credentials]."

pages/badges/badge-single.jsx
Pomax: lol, nice.

oh yeah I'm pretty sure that was supposed to be https and I forgot to add the `s` there.

pages/badges/badge-single.jsx
Pomax: do we not want to link to the badge itself? As a user I don't want to share "mozilla's badge program", I want to show off what *my* stuff, in this case the badge I just earned =)

gideonthomas:
@Pomax code looks good to me, am I able to test this locally? If not, r+

pages/badges/badge-single.jsx
hannahkane: @Pomax - are you suggesting the page with details about the particular badge, or a page that shows what badges the user has earned?

Pomax:
You are, if you have learn + teach-api (+ id + login) set up, and make sure to give teach-api the correct CREDLY variable values.

pages/badges/badge-single.jsx
Pomax: either/or - from a user perspective I imagine they want to share a link that shows "them" as having earned a specific badge, but that would be a credly page (since we don't have pages for "a specific user's badge"). So alternatively we could at least make the share link point to the learning site page for that particular badge.

gideonthomas:
This works with a few caveats:
- No styling (although this is mentioned above) for the links
- The facebook share does not have any text in the share box

Pomax:
it appears we can't presupply a facebook share message unless we use the Javascript with-app-id approach, which we probably don't wan to do.

Pomax:
Filed https://github.com/mozilla/learning.mozilla.org/issues/2241 and https://github.com/mozilla/learning.mozilla.org/issues/2242 as follow-ups around this patch.

PR #2207, opened 2016-09-13
Some cleanup and QA

ScottDowne:
@alanmoo some fixes for the feedback I have gotten up to this point, and some generic cleanup. R?

### DIFF
--- a/less/pages/event-resources.less
+++ b/less/pages/event-resources.less
@@ -1,38 +1,3 @@
-.tabulator-hide-nothing() {
-  .tabulator {
-    .tabulator-head {
-      display: none;
-    }
-    .tabulator-head-no-js {
-      width: 100%;
-      display: block;
-      background: #BCC7E6;
-      padding: 1.3rem 1.8rem;
-      font-size: 1.5rem;
-      color: #2D4A8C;
-      font-weight: bold;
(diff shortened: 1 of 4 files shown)


Pomax:
can you change the commit title so it reflects the work that was done? (we uses the commit title/message to build the changelog, "some cleanup and QA" doesn't say *what* got cleaned up and QA'd =)

ScottDowne:
@Pomax updated.

alanmoo:
This looks good, though there's still some work to go before this is ready when looking at it on mobile.

ScottDowne:
@alanmoo yeah, I just took a quick look at mobile. Things I noticed that are unique to these pages:

The events page
- [ ] The "event resources" button nav button text doesn't fit.
- [ ] The hero unit image is totally off. Needs to be positioned better.

The events/resources page
- [ ] Hosting a Maker Party list text looks too big, needs some work.
- [ ] The Logos & Assets items look a little off. This wasn't something I built, so might not be new. Should be fixed anyway.
- [ ] Become a Maker Party partner section needs some centering.

And in general I feel like a lot of the headers are too big for mobile, but I wonder if that's all of l.m.o? and if it's intended?

PR #2214, opened 2016-09-15
Remove the static site check for GA

Pomax:
This should simply always enable google analytics and initialize it (if there is a GA_ACCOUNT value set) for the client code.


gvn:
So, this LGTM code-wise, but I'm not sure how to actually test it. Any thoughts?

Pomax:
I'm not sure we really can - the code doesn't actually change any behaviour, just cleans up a thing that was trivially true already anyway since the client bundle *always* runs in the browser.

PR #2222, opened 2016-09-19
Update events pages logog and assets

ScottDowne:
Updates to some assets in the events pages.


ScottDowne:
@alanmoo another one for tomorrow.

PR #2226, opened 2016-09-20
Add new string specific to newsletter sign up

### LINKED ISSUE #2173: Create separate localizable entry for Sign Up string
Apparently a localizable "Sign Up" entry is currently shared between the "Sign in or Sign Up" string on the main page and the button needed to sign up for the news letter on the community page. Since some locales may need to use different localizations for them (one is more like "registering", the other refers to "subscribing"), there is no chance this can be fixed at this point. (In general, it’s never a good idea to share localizable strings for this reason.)

Please create a separate entry for the signup button to make this possible. Unless community page content is about to change, since some of its strings do not appear to be localizable at all.

alanmoo:
- [x] Localization addressed

@TonnesM, I believe this Fixes #2173. Can you confirm?

- Change newsletter sign up to a new string key
- Duplicate all `sign_up` strings across locales as `newsletter_sign_up`


TonnesM:
@allanmoo I think it should, thanks. Not sure if localizers need to be notified?

PR #2273, opened 2016-09-27
link size fix, labels around radio/checks, outline ZERO, other spacing

Pomax:
LESS fixes and fixes to `<Form>` to enable styling.

fixes https://github.com/mozilla/learning.mozilla.org/issues/2272
- pledge link font is now the right size

fixes https://github.com/mozilla/learning.mozilla.org/issues/2261
- radio and checkboxes now have a wrapping label

fixes https://github.com/mozilla/learning.mozilla.org/issues/2260
- outline is 0. You hear me, LESS? ZERO.

### DIFF
--- a/less/pages/clubs-form.less
+++ b/less/pages/clubs-form.less
@@ -4,9 +4,18 @@
     width: 80%;
     margin: auto;
 
+    * {
+      // serious, get out of here, outline.
+      outline: 0 !important;
+    }
+
     .learn-more {
       text-align: center;
       margin: 2em 0;
+    }
+
+    input.controlled {
+      margin-top: 0.5em;
     }
 
     textarea {
@@ -35,7 +44,7 @@
       }
     }
 
-    .pledge {
+    .pledgeAgreement {
       margin-top: 1.5em;
       font-size: 90%;
       a {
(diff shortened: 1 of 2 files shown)


less/pages/clubs-form.less
```diff
@@ -4,11 +4,20 @@
     width: 80%;
     margin: auto;
 
+    * {
+      // serious, get out of here, outline.
+      outline: 0 !important;
+    }
```
gvn: Hmm. Not so sure about this?

If we are gonna add this rule we shouldn't put it in this file, but in one of the more "globally oriented" ones.

gvn: Oh...I didn't see it was nested inside `.clubs-form`. Seems ok then...

flukeout:
I ready something yesterday about ``* { }`` rules being frowned upon now because they are inefficient and take a while to apply (every element is considered), could we scope this to inputs or something more relevant to our problem?

Not sure if this is a valid concern.

flukeout:
Looks good, PR fixes all of the issues listed above. R+

less/pages/clubs-form.less
Pomax: yeah, it's explicitly for the form only. I'm not a fan of !important, except for three things: `.hidden` class, `outline:0` and `user-select:none` =)

PR #2274, opened 2016-09-27
focus on "other" field when it gets revealed

Pomax:
If a selector or radio/checkbox group has an "other" option that the user selects, and this reveals a text field for filling in a custom value, the code will now automatically move focus to this "other" field.

This has a nasty hack that is necessary to deal with `react-select` being a jerk and stealing focus back when the browsers calls `.focus()` on something else. Completely page breaking, so... setTimeout. Yeah.

fixes https://github.com/mozilla/learning.mozilla.org/issues/2256

### DIFF
--- a/pages/clubs/form/builder/Form.jsx
+++ b/pages/clubs/form/builder/Form.jsx
@@ -1,4 +1,5 @@
 var React = require('react');
+var ReactDOM = require('react-dom');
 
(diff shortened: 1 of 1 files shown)


flukeout:
Testing now.

flukeout:
Works! R+

PR #2288, opened 2016-10-04
Events resources host pack localized links.

ScottDowne:
@alanmoo This swaps out the link in the "get host pack" link in the events/resources page with localized links for the links we have localized.


ScottDowne:
@alanmoo updated. R?

PR #2291, opened 2016-10-05
Turn on maker party link in navigation.

ScottDowne:
This just turns on Maker Party links in the nav, when we're ready to ship.


ScottDowne:
@alanmoo So this can be reviewed, but don't merge quite yet. It'll be merged later today.

I'm also not sure how to push learning to prod, but we'll need to do that today too.

PR #2336, opened 2016-10-12
Fix #2313 - Badge details not loading due to accessing incorrect credly property

### LINKED ISSUE #2313: Badge details not loading
I tested the Credly -> Learning.m.o workflow by adding a new badge to the Credly staging server. The badge appears on the Learning staging site (yay!), but when I click to view the detail page, the badge details do not load (even after waiting a long time).

screen shot 2016-10-06 at 9 17 19 pm

gideonthomas:
This will allow you to load eligible badges based on what the issue described.

I haven't touched "Achieved" or "Pending" badges because right now it shows some hard-coded text. But I don't know what the HTML is that needs to go there (maybe @Pomax can advise). Accordingly, for the data that I think will be used there, I have left that regex pattern alone but have separated it out from the criteria to apply for a badge regex.

### DIFF
--- a/pages/badges/badge-single.jsx
+++ b/pages/badges/badge-single.jsx
@@ -133,8 +133,10 @@ var BadgePage = React.createClass({
     if (data.earned) { status = Badge.achieved; }
     if (data.pending) { status = Badge.pending; }
 
-    var reqs = bdata.require_claim_evidence_description;
-    var matched = reqs.match(/((\d+(\.)?)?[^.]+)/g);
+    var evidenceRegex = /((\d+([.\!?])?)?[^.]+)/g;
+    var criteriaRegex = /[\n\r]+/g;
+    var criteria = bdata.criteria.length > 0 ? bdata.criteria.trim().split(criteriaRegex) : [];
+    var evidence = bdata.require_claim_evidence_description.match(evidenceRegex) || [];
 
     this.setState({
       badge: {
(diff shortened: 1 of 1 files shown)


pages/badges/badge-single.jsx
```diff
@@ -133,8 +133,10 @@ var BadgePage = React.createClass({
     if (data.earned) { status = Badge.achieved; }
     if (data.pending) { status = Badge.pending; }
 
-    var reqs = bdata.require_claim_evidence_description;
-    var matched = reqs.match(/((\d+(\.)?)?[^.]+)/g);
+    var evidenceRegex = /((\d+([.\!?])?)?[^.]+)/g;
```
Pomax: regex are giant red warning lights with lots of bells - what are you actually extracting here, and can that be done without regex (because if it can: let's do that instead)

gideonthomas: I honestly have no idea! It seems like the "evidence" and "criteria" on Credly are inputted as a single long string containing multiple points. From what I understand from the current code, this regex is meant to take that string and break into multiple strings based on each point. All of this is based on assumptions though that it was decided that there would be a set input format for those two fields and that we want to split them up on the client to display them better. It would be good for us to chat about this though!

Pomax: sounds like a plan, let's chat today

PR #2344, opened 2016-10-14
coordinator page basics

### LINKED ISSUE #2280: Implement page for Regional Coordinator bios
Implementation of the project #1815.
Here is the final mockups: https://redpen.io/p/nj4f7768ca3e7b79cf

gvn:
closes #2280


gvn:
@taisdesouzalessa Do you want to give this a design review?

I have it staged here:

https://learning-mozilla-org-s-pr-2344.herokuapp.com/en-US/clubs/start

taisdesouzalessa:
Hey @gvn thanks for that! It looks great :). There are a some design fixes to do, should I add them as individual comments on this thread? Or open a new ticket?

gvn:
If they're relatively small I'm fine with you just adding them to this ticket. 👍

taisdesouzalessa:
ok, thanks! I'll add numbers to the issues so we can refer to them in the thread (like QA#1, QA#2...)
QA#1 - DESKTOP 
Please change the order of the list items so the Regional Coordinators item gets closer to the section that leads to Regional Coordinators page.
items-order

taisdesouzalessa:
QA#2 - DESKTOP 
I thought the base typeface of this page was 18px but the live site has it in 16px, so we have to adjust the other typeface sizes to keep the hierarchy (sorry for that...) and also there are some color adjustments in the "location" sentence: 
 
rc-desktop-page

14.4px for a typeface seems a weird number but that is what I see in when I use a font size inspector. The font size (that I see as 14.4 px) should be the same as in this page, for your reference: https://learning-mozilla-org-s-pr-2344.herokuapp.com/en-US/clubs/running - footer part: Connect on Twitter/ Facebook (the 3-line copy)

taisdesouzalessa:
QA#3 DESKTOP
Could you please set an anchor ID for each person?

taisdesouzalessa:
QA#4 DESKTOP
For "Gauthamraj Elango" and "Sayak Sarkar" there is only 1 club. Please replace "Clubs [NAME] supports:" with the singular version "Club [NAME] supports:"

taisdesouzalessa:
QA#5 MOBILE
Is it possible to stack the picture in the mobile version so the column is not very narrow?
mobile-featured-rc-section

taisdesouzalessa:
QA#6 MOBILE
If possible, could we limit the size of the ellipse for screens = or smaller than 767px (width)? it seems quite big on this view. It looks good on the phone though (portrait mode). On landscape mode the ellipse is pretty big. I think this issue is not a deal breaker but a more nice to have.
screen shot 2016-10-14 at 2 28 08 pm

gvn:
@taisdesouzalessa Ok, I made all those tweaks if you want to take another look!

taisdesouzalessa:
@gvn thank you so much for that. They look great...yay!

I just have 2 last final comments:
QA#7 DESKTOP Could we remove the margin-bottom of the "Clubs [name] supports"?:
clubs-name-supports

QA#8 DESKTOP Is is possible to remove the link behavior of each name? 
I want the anchor (so the Regional Coordinators can share a specific point of the page) but I didn't want that to be triggered by a click on the name (if the name looks like a link on hover, users may think they will be taken to other page). I was thinking about removing the <a></a> from the names and adding an id to h2. Not sure if this is the right approach though, you may know a better one.
screen shot 2016-10-18 at 1 47 28 pm
link-names

After those I have no other comments. Thank you, Gavin!

taisdesouzalessa:
Hey @gvn - talked to Kristina and Luke here and we talked about how to solve the anchor link problem you brought up. Thanks for bringing the issue!

Could we add an extra link like this one here?
screen shot 2016-10-20 at 11 44 26 am

On mobile, we should pile up the 3 infos (location/twitter/direct link) and remove the "I" (thing that divides items on desktop). One hing to watch out is to make that there is enough vertical space between the twitter handle and the direct link to tap. So if you have to increase the spacing of those items a little bit on mobile, it is ok:
screen shot 2016-10-20 at 11 44 43 am

taisdesouzalessa:
QA#9 DESKTOP: @gvn IMPORTANT - This is a nice to have not a must have. It doesn't break the page or anything, just a little detail I noticed. If it's too much for you to change, it is ok not to apply this feedback. -  While you are in this page, could you please change the line height of the ".coordinator-bio .bio" to 1.75 em to it is the same standard as the first paragraph on the page? "Regional Coordinators mentor multiple Club Captains..." - 
line-height

gvn:
@taisdesouzalessa Ok, this should all be fixed up if you want to look at staging again.

taisdesouzalessa:
Yeah. It is awesome. Thank you so much, @gvn.  

When you added the direct link, something happened to the pin icon (besides the location) - it seems a bit cut on the left side. Would you mind taking a look at that? I solved it in my browser by adding a margin-left of 0.1 em in ".coordinator-bio .glyph" - but you know better than me if this is a good solution:
screen shot 2016-10-20 at 3 31 03 pm

All the rest is super cool! Thank you :)

gvn:
@taisdesouzalessa Now that we're done with design review it needs a code review before landing.

@flukeout do you have time to do a code review?

flukeout:
Yep, will look over it now

less/pages/clubs.less
```diff
@@ -628,3 +628,131 @@ hr.square-divider {
...
+    padding: 0;
+  }
+
+  div:nth-child(2) {
+    text-align: center;
+    padding: 15px 0;
+  }
+}
+
+.clubs-tabswitcher {
```
flukeout: Could we remove this tab from the markup? Is it never shown?

less/pages/clubs.less
```diff
@@ -628,3 +628,131 @@ hr.square-divider {
...
+    content: "\f041";
+  }
+
+  .glyph-twitter:before {
+    content: "\f099";
+    color: #1DA1F2;
+  }
+
+  .glyph-link:before {
+    content: "\f0c1"; // fa-link
```
flukeout: Label other glyphs with icon type as well?

less/pages/clubs.less
gvn: Not without refactoring mofo-ui.

flukeout: Oh dang, it just comes with 4?

PR #2347, opened 2016-10-17
Badge prefil link email

Pomax:
Based on the pending PR for badge followup - the relevant commit is *only* https://github.com/mozilla/learning.mozilla.org/commit/a5acffdce0dedd13f1077e3718207d7f70d73286

checks off the "autofill user email" entry on https://github.com/mozilla/learning.mozilla.org/issues/2346

### DIFF
--- a/lib/teach-api.js
+++ b/lib/teach-api.js
@@ -72,17 +72,13 @@ _.extend(TeachAPI.prototype, {
     }
   },
   getEmail: function(cb) {
-    var errMail = {
-      email: false
-    };
-
     // This is not allowed to be persistent information
     request.get(this.baseURL + '/auth/email')
       .withCredentials()
       .accept('json')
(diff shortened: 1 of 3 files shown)


lib/teach-api.js
```diff
@@ -71,6 +71,22 @@ _.extend(TeachAPI.prototype, {
...
+      email: false
+    };
+
+    // This is not allowed to be persistent information
+    request.get(this.baseURL + '/auth/email')
+      .withCredentials()
+      .accept('json')
+      .end(function(err, res) {
+        if (err) {
+          return errMail;
```
gideonthomas: shouldn't you return `errMail.email`?

Pomax: it should call `cb("")` even. Fixed

PR #2351, opened 2016-10-17
attachment styling

Pomax:
better attachment styling, and a few fixes to deal with attachment removal

### DIFF
--- a/components/badges/RequirementRow.jsx
+++ b/components/badges/RequirementRow.jsx
@@ -51,6 +51,12 @@ var RequirementRow = React.createClass({
       );
     }
 
+    var howmany = "a";
+
+    if (this.state.evidenceFiles.length > 0) {
+      howmany = "another";
+    }
+
     return (
       <div className="apply-send-qualifications">
         {/*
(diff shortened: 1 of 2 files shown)


components/badges/RequirementRow.jsx
```diff
@@ -86,7 +92,7 @@ var RequirementRow = React.createClass({
       return (
         <span className="attached" key={`evidence-${e.name}`}>
           {e.name}
-          <span className="fa fa-times" onClick={this.removeAttachment(e.name)}/>
+          <span className="close fa fa-times" onClick={this.removeAttachment(e.name).bind(this)} title="Remove this attachment"/>
```
gideonthomas: why `.bind`?

gideonthomas:
@Pomax in this patch I only see the "Submit evidence" button for the first task and none of the others

Pomax:
@gideonthomas that would be because the teach-api now requests data with x-allow-html, so the splitting regex you still have in learning (until you merge in the x-allow-html branch for learning) will not work anymore - you'll only see a single task

components/badges/RequirementRow.jsx
gideonthomas: damn, ok this is fine then

PR #2352, opened 2016-10-17
full rather than short badge description

Pomax:
This swaps the `short_description` field for the (previously unavailable) `description` field, so that our badges actually have a decent description on the site.

### DIFF
--- a/pages/badges/badge-single.jsx
+++ b/pages/badges/badge-single.jsx
@@ -152,11 +152,15 @@ var BadgePage = React.createClass({
       criteria = criteria.split(splitOnItems).map(s => s.trim()).filter(s => s);
     }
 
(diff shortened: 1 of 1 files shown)


pages/badges/badge-single.jsx
```diff
@@ -152,11 +152,15 @@ var BadgePage = React.createClass({
       criteria = criteria.split(splitOnItems).map(s => s.trim()).filter(s => s);
     }
 
+    var desc = document.createElement("p");
+
+    desc.innerHTML = bdata.description;
```
Pomax: we do this to turn characters that Credly turned into html entities back into plain text. As these are our own badges, the risk of user generated content XSS is extremely low.

PR #2357, opened 2016-10-18
better 'return to badge' styling, and a content switch on application…

Pomax:
This centers the "return to badge" button in the application confirmation dialog, and also adds a context switch so that the modal cannot be closed until the POST request has properly finished - if the POST fails (not-200-return) the content is switch out for an error notice and a request to contact us, with a red button rather than yellow.

### DIFF
--- a/pages/badges/badge-single.jsx
+++ b/pages/badges/badge-single.jsx
@@ -245,7 +245,7 @@ var BadgePage = React.createClass({
           <p>
             Something went wrong with your application for this badge.
             Please let us know what you were doing so that we can
-            look into getting that fixes as soon as possible for you!
+            look into getting that fixed as soon as possible for you!
           </p>
           <button className="btn back" onClick={this.hideApplyModal}>Back to the badge</button>
         </Modal>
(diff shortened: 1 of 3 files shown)


pages/badges/badge-single.jsx
```diff
@@ -236,6 +229,40 @@ var BadgePage = React.createClass({
...
+    var error = this.state.applicationError;
+    var disabled = "disabled";
+
+    if (canClose || error) {
+      disabled = null;
+    }
+
+    if (error) {
+      return (
+        <Modal className="modal-credly error folded" hideModal={this.state.canCloseModal? this.hideApplyModal : false}>
```
gideonthomas: space before ?

Pomax: I don't understand? before what? and is there space, or should there be space?

gideonthomas: sorry, I meant in the ternary condition, can we add a space between the condition and the `?` operator?

Pomax:
added a space, although that's not a thing we;ve set up a js style rule for so atm either is "fine" from a linting perspective

gideonthomas:
feel free to merge in anytime

PR #2359, opened 2016-10-20
preserve evidence when application fails

Pomax:
How to test this: in `badge-single.jsx`, have the `claimBadge` function start as:

```
  claimBadge: function() {
    var evidences = [];
    var userSupplied = this.state.evidence;

    return this.setState({
      applying: true,
      showApplyModal: true,
      canCloseModal: false
    }, function() {
      this.handleClaimRequest(true, { error: false });
    });

    ...rest of code...
```

This bypasses the actual badge application and triggers the callback for a failed application - any evidence you may have filled in should be preserved now, rather than getting completely wiped empty.


gideonthomas:
code looks good and wfm, r+

PR #2365, opened 2016-10-25
Fixing upcoming date and restoring old date.

gvn:


gvn:
@hannahkane I think I changed the wrong date in the previous patch. This should be correct now.

hannahkane:
yikes. you're right. sorry I missed that. R+ for realsies.

PR #2386, opened 2016-11-16
Temporarily link to etherpad externally

gideonthomas:
Looks like etherpad has buffed their security and we can no longer put it in an iframe on our site.

Since our community call is today evening, for now I just made it clearer that they need to click on a link to open the etherpad in a new tab.


alanmoo:
Just wait for Travis

PR #2395, opened 2016-12-13
rephrased the evidence heading

Pomax:
fixes https://github.com/mozilla/learning.mozilla.org/issues/2377

### DIFF
--- a/pages/badges/badge-single.jsx
+++ b/pages/badges/badge-single.jsx
@@ -398,7 +398,7 @@ var BadgePage = React.createClass({
     return (
       <div className="badge-requirement" key={'badge-requirement-list'}>
         <h3 className={'text-light'}>Badge Requirements</h3>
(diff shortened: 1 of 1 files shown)


pages/badges/badge-single.jsx
```diff
@@ -398,7 +398,7 @@ var BadgePage = React.createClass({
     return (
       <div className="badge-requirement" key={'badge-requirement-list'}>
         <h3 className={'text-light'}>Badge Requirements</h3>
-        <p>Make or write something that demonstrates your understanding of any two or more of the following:</p>
+        <p>Complete the following three tasks and submit the required evidence:</p>
```
gideonthomas: are we sure it's always going to be three tasks?

Pomax:
I had the same question, an-me and iris assured me there would only ever be three tasks on every badge.

PR #2396, opened 2016-12-13
fix badge description cutoff by removing its (arbitrary) CSS height

Pomax:
fixes https://github.com/mozilla/learning.mozilla.org/issues/2376

### DIFF
--- a/less/pages/badges.less
+++ b/less/pages/badges.less
@@ -281,15 +281,6 @@
 }
 
 
-.badge-vertical-icon{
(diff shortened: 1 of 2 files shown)


Pomax:
hmm, I can't reproduce that, which os/browser?

gideonthomas:
@Pomax still can't reproduce? I'm on Mac. Tried it on:
- Chrome 54.0.2840.98
- FF 51.0b7

PR #2431, opened 2017-02-09
Update maxAge value of Strict-Transport-Security

### LINKED ISSUE #2430: maxAge for "Strict-Transport-Security" header is set to 90,000 days
😂 

https://github.com/mozilla/learning.mozilla.org/blob/master/app.js#L60-L62

I believe we want it to be 90 days instead. I will file a PR.

/cc @Pomax @cadecairos

mmmavis:
Fixes #2430

### DIFF
--- a/app.js
+++ b/app.js
@@ -57,24 +57,15 @@ app.use(helmet.xssFilter({
   setOnOldIE: true
 }));
 
+
+// maxAge for HSTS header must be at least 18 weeks (see https://hstspreload.org/)
 app.use(helmet.hsts({
-  maxAge: 1000 * 60 * 60 * 24 * 90
+  maxAge: 60 * 60 * 24 * 7 * 18 // 18 weeks in seconds
 }));
 
 app.use(helmet.ieNoOpen());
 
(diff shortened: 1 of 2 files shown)


mmmavis:
@Pomax @cadecairos can either of you review this PR please? 🍰

mmmavis:
oops there's another place i should fix

app.js
```diff
@@ -67,7 +67,7 @@ app.use(helmet.noSniff());
 
 if (process.env.HPKP) {
   app.use(helmet.hpkp({
-    maxAge: 1000 * 60 * 60 * 24 * 90,
+    maxAge: 60 * 60 * 24 * 90, // 90 days in seconds
```
cadecairos: This is hpkp - something we should actually remove, since we don't use it and it's somewhat dangerous to use.

the hsts time should still be fixed however. 

Set it to at least 18 weeks (a requirement for hsts preloading - https://hstspreload.org/)

mmmavis: kk will update

mmmavis:
🎉

PR #2432, opened 2017-02-11
Hide email signups

### LINKED ISSUE #2426: Hide email sign-up links from homepage and community page
Let's remove the links for email-sign ups. 

One is on the homepage - the first CTA button. Another is on the Community page towards the bottom. Might be worth double-checking if there are any other links on the site to sign-up but I think those are the only two ;)

mmmavis:
Fixes #2426 

@gvn can you 👀 ? Since this is a temp hide, I can't decide if I should just comment out related bits with some note (seen in my PR) OR remove that code on both `HomePage.jsx` and `CommunityPage.jsx `.

### DIFF
--- a/components/icon-buttons.jsx
+++ b/components/icon-buttons.jsx
@@ -1,14 +1,19 @@
 var React = require('react');
 
 var IconButtons = React.createClass({
+  propTypes: {
+    children: React.PropTypes.object.isRequired
+  },
   render: function() {
+    let numColSpan = 12 / this.props.children.length; // Bootstrap uses 12-col layout
+
     return (
       <div className="icon-buttons inner-container">
         <section className="row">
           {
              React.Children.map(this.props.children, function(iconButton){
                return(
-                 <div className="col-sm-4 col-md-4 col-lg-4">{iconButton}</div>
+                 <div className={`col-sm-${numColSpan} col-md-${numColSpan} col-lg-${numColSpan}`}>{iconButton}</div>
                );
              })
           }
--- a/components/icon-buttons.jsx
+++ b/components/icon-buttons.jsx
@@ -5,13 +5,16 @@ var IconButtons = React.createClass({
     children: React.PropTypes.object.isRequired
   },
   render: function() {
-    let numColSpan = 12 / this.props.children.length; // Bootstrap uses 12-col layout
+    let nonNullChildren = this.props.children.filter((child) => {
+      return child !== null;
+    });
+    let numColSpan = 12 / nonNullChildren.length; // Bootstrap uses 12-col layout
 
     return (
       <div className="icon-buttons inner-container">
         <section className="row">
           {
-             React.Children.map(this.props.children, function(iconButton){
+             React.Children.map(nonNullChildren, function(iconButton){
                return(
                  <div className={`col-sm-${numColSpan} col-md-${numColSpan} col-lg-${numColSpan}`}>{iconButton}</div>
                );
--- a/components/icon-buttons.jsx
+++ b/components/icon-buttons.jsx
@@ -1,24 +1,32 @@
 var React = require('react');
+var classnames = require('classnames');
 
 var IconButtons = React.createClass({
   propTypes: {
     children: React.PropTypes.object.isRequired
   },
   render: function() {
-    let nonNullChildren = this.props.children.filter((child) => {
-      return child !== null;
-    });
-    let numColSpan = 12 / nonNullChildren.length; // Bootstrap uses 12-col layout
+    let children = this.props.children.filter(child => !!child);
+    let numChildren = children.length;
+    // Bootstrap uses 12-col layout.
+    // We want icon buttons to fill up the entire row when possible
+    // Let's calculate how many columns each button should occupy
+    let numColSpan = Math.floor(12 / numChildren);
+    // if buttons can't fill up the entire row, add offset to the first button
+    // so the group of buttons looks centered in the row
+    let colOffset = Math.floor((12 - numColSpan*numChildren) / 2);
 
     return (
       <div className="icon-buttons inner-container">
         <section className="row">
           {
-             React.Children.map(nonNullChildren, function(iconButton){
-               return(
-                 <div className={`col-sm-${numColSpan} col-md-${numColSpan} col-lg-${numColSpan}`}>{iconButton}</div>
-               );
-             })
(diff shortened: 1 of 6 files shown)


gvn:
If this is temporary can we do it through a feature flag in the env instead?

mmmavis:
Yup agree.

pages/community/CommunityPage.jsx
```diff
@@ -38,9 +37,20 @@ var communities = communityList.map(function(community) {
 var CommunityPage = React.createClass({
   statics: {
     pageTitle: 'Community',
-    pageClassName: 'community-page',
-    SignupForm: SignupForm,
-    validateSignupForm: validateSignupForm
```
mmmavis: the above 2 lines seem to be legacy code

components/newsletter-signup/SignupForm.jsx
```diff
@@ -74,7 +77,7 @@ var SignupForm = React.createClass({
     var privacy_policy_link = <a href="https://www.mozilla.org/privacy/websites/"><FormattedMessage id="privacy_policy" defaultMessage="privacy policy"/></a>;
 
     return (
-      <form className="mailinglist-signup" action={process.env.NEWSLETTER_MAILINGLIST_URL} method="POST" onSubmit={this.handleSubmit}>
+      <form className="mailinglist-signup" action={process.env.NEWSLETTER_MAILINGLIST_URL} method="POST">
```
mmmavis: `onSubmit` gets triggered only if form has passed browser validation. Since we are not relying on that for form validation, I've moved the event handler to line 96 (as submit button's onClick event handler)

pages/home/HomePage.jsx
```diff
@@ -15,17 +15,19 @@ var IconButton = require('../../components/icon-button.jsx');
...
 var BlogSection = require('./BlogSection.jsx');
 
 var fixLocation = require('../../lib/fix-location.js');
 
+const HIDE_NEWSLETTER_SIGNUP_FORM = process.env.HIDE_NEWSLETTER_SIGNUP_FORM.toLowerCase() === 'true';
+
 var HomePage = React.createClass({
   statics: {
     pageClassName: 'home-page',
-    validateSignupForm: validateSignupForm,
```
mmmavis: seems to be legacy code to me

mmmavis:
@gvn PR updated!

gvn:
I'll try to review this soon...in the meantime there's a build failure on Travis that needs fixin'.

components/icon-buttons.jsx
```diff
@@ -1,14 +1,22 @@
...
 
 var IconButtons = React.createClass({
+  propTypes: {
+    children: React.PropTypes.object.isRequired
+  },
   render: function() {
+    let nonNullChildren = this.props.children.filter((child) => {
+      return child !== null;
+    });
+    let numColSpan = 12 / nonNullChildren.length; // Bootstrap uses 12-col layout
```
gvn: This could result in floating point numbers if you don't use Math.floor or Math.ceil

components/icon-buttons.jsx
```diff
@@ -1,14 +1,22 @@
...
+
     return (
       <div className="icon-buttons inner-container">
         <section className="row">
           {
-             React.Children.map(this.props.children, function(iconButton){
+             React.Children.map(nonNullChildren, function(iconButton){
                return(
-                 <div className="col-sm-4 col-md-4 col-lg-4">{iconButton}</div>
+                 <div className={`col-sm-${numColSpan} col-md-${numColSpan} col-lg-${numColSpan}`}>{iconButton}</div>
```
gvn: Kinda confused by this. I think you only need `col-${numColSpan}` not all 3.

components/newsletter-signup/SignupForm.jsx
```diff
@@ -86,11 +89,11 @@ var SignupForm = React.createClass({
           <label htmlFor={idPrefix+"email"} className="sr-only"><FormattedMessage id="email" defaultMessage="email"/></label>
           <div className="icon-field-container">
             <i className="fa fa-envelope"></i>
-            <input id={idPrefix+"email"} name="email" type="email" size="30" placeholder="email@example.com" valueLink={this.linkState("email")} required />
+            <input id={idPrefix+"email"} name="email" type="email" size="30" placeholder="email@example.com" valueLink={this.linkState("email")} ref="email" required />
```
gvn: Don't concatenate strings with `+`, just use the template literal syntax `${idPrefix}email`.

pages/home/HomePage.jsx
```diff
@@ -58,18 +60,17 @@ var HomePage = React.createClass({
...
+    let newsletterCTA = !HIDE_NEWSLETTER_SIGNUP_FORM ?
+                          <IconButton
+                            imgSrc="/img/pages/home/svg/icon-newsletter.svg"
+                            head={this.context.intl.formatMessage({id: 'get_email_update'})}
+                            onClick={this.handleEmailBtnClick}
+                            className={"newsletter"}
+                          /> : null;
+
+    return <IconButtons>
+            { newsletterCTA }
```
gvn: Just put the stuff at line 64 inline here since it's only used once. No need for a variable.

For conditional JSX like this where you only have content for an `if` case, use [this syntax](https://facebook.github.io/react/docs/conditional-rendering.html#inline-if-with-logical--operator) instead of a ternary. It's easier to read and maintain.

components/icon-buttons.jsx
gvn: Also, not sure why this is needed?

components/icon-buttons.jsx
```diff
@@ -1,16 +1,32 @@
...
   render: function() {
+    let children = this.props.children.filter(child => !!child);
+    let numChildren = children.length;
+    // Bootstrap uses 12-col layout.
+    // We want icon buttons to fill up the entire row when possible
+    // Let's calculate how many columns each button should occupy
+    let numColSpan = Math.floor(12 / numChildren);
+    // if buttons can't fill up the entire row, add offset to the first button
+    // so the group of buttons looks centered in the row
+    let colOffset = Math.floor((12 - numColSpan*numChildren) / 2);
```
mmmavis: Not sure if we need this as we can't guarantee group of buttons *can really be* positioned at center. e.g., when `numChildren` is 9 `colOffset` will be `1` instead of `1.5`(which is a number that doesn't make sense in Bootstrap's world)

PR #2437, opened 2017-02-17
fix: Relative link fix for LinkAnchorSwap

### LINKED ISSUE #2171: fix LinkAnchorSwap locale with relative link
Code right now:

```
  getLocalizedTo: function() {
    return "/" + this.context.intl.locale + this.props.to;
  },
```

which suggests that we could go from `/en-US/current/page`, through a link `../other` to `/en-US../other` rather than to `en-US/current/page/../other`.

This should probably be:

```
  getLocalizedTo: function() {
    var to = this.props.to;

    if (to.indexOf('./') === 0 || to.indexOf('../') === 0) {
      return to;
    }

    return "/" + this.context.intl.locale + to;
  },
```

Or something even more encompassing.

aayushsanghavi:
- [x] Localization addressed

Fixes #2171 

- as @Pomax suggested in the issue itself

### DIFF
--- a/components/link-anchor-swap.jsx
+++ b/components/link-anchor-swap.jsx
@@ -30,7 +30,13 @@ var LinkAnchorSwap = React.createClass({
   },
 
   getLocalizedTo: function() {
-    return "/" + this.context.intl.locale + this.props.to;
+    var to = this.props.to;
+
+    if (to.indexOf('./') === 0 || to.indexOf('../') === 0) {
+      return to;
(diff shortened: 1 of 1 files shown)


aayushsanghavi:
Could someone please review this PR?

Pomax:
tests pass, it looks like the locale is lost on the /en-US/clubs page when linking to the activities, but that's a separate issue.

PR #2470, opened 2017-04-20
Updated gigabit city map & added two more cities

### LINKED ISSUE #2466: Update Gigabit Cities Map
- [ ] Add two dots on map on the [main Gigabit page](https://learning.mozilla.org/en-US/gigabit/) representing Eugene, OR and Lafayette, LA. 
- [ ] Add these cities to the list on the right, under a "Coming Soon" heading (we don't yet have contact info for these cities, so it will just be the names)

flukeout:
Fixes #2466

Check ``/gigabit`` there should be a map with 5 cities and the following layout lower in the page...

image

cc @alanmoo @hannahkane


flukeout:
Sorry for the late merge @alanmoo - thanks for the updates!

PR #2471, opened 2017-04-21
City update tweaks

alanmoo:
Some fixes for mobile

### DIFF
--- a/less/pages/gigabit.less
+++ b/less/pages/gigabit.less
@@ -31,13 +31,13 @@
 .gigabit-contact-block {
   padding: 0 30px;
   margin-bottom: 20px;
-
-  .row {
-    margin-bottom: 20px;
-  }
 }
 
 .gigabit-contact {
+   @media (min-width: @screen-sm-min) {
+    min-height: 10em;
+  }
+
   h4 {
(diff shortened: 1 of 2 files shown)


less/pages/gigabit.less
```diff
@@ -31,13 +31,13 @@
...
   padding: 0 30px;
   margin-bottom: 20px;
-
-  .row {
-    margin-bottom: 20px;
-  }
 }
 
 .gigabit-contact {
+   @media (min-width: @screen-sm-min) {
```
alanmoo: This isn't great, but it does the job.

pages/gigabit/About.jsx
```diff
@@ -44,7 +44,7 @@ module.exports = React.createClass({
         </div>
 
         <div className="clearfix text-center">
-          <img className="w-75 p-a-2" src="/img/pages/gigabit/map-v2.svg"/>
+          <img className="img-responsive m-x-auto p-a-2" src="/img/pages/gigabit/map-v2.svg"/>
```
alanmoo: I wasn't sure what the w-75 was for. `img-responsive` gives max-width so the image scales down on mobile.

PR #2465, opened 2017-04-18
fix: Remove reference to Open News fellowship

aayushsanghavi:
- [ ] Localization addressed

Fixes #2463 

**Changes proposed in this pull request:**
- Removed reference to Open News fellowship from the opportunities page

This is how it looks (marked in the black border)- 

screenshot from 2017-04-18 08-39-04

PR #2495, opened 2017-06-21
#2465 follow-up

TheoChevalier:
> the recommendation for changes to strings is that you create a new key-value pair with the new string and use that key in your code instead

Yup, and the last part is to remove the obsolete one :D Not a major issue here because of the status of the project, but the idea is if you’re adding new locales (also true for incomplete locales), you don’t want to keep asking them to translate this obsolete string.


gideonthomas:
Ah totally missed that! Thanks for catching that!

PR #2500, opened 2017-07-12
update to secure version of Node 4

Pomax:


Pomax:
PR updated, travis is now set to "whatever is the latest version of 4" which by definition is going to be the latest LTS version of 4

Issue #2510, opened 2017-08-22
Allow for multiple links on the Gigabit portfolio detail pages


hannahkane:
In a given Gigabit Portfolio item detail page (see [an example](https://learning.mozilla.org/en-US/gigabit/portfolio/virtual-reality-stem-lessons)), there is a button at the bottom, labeled "website" that takes the user to an external site.

Some projects have up to three links to point to, so we need to allow for that in the UI.

Looking for a relatively quick solution that accommodates 0, 1, 2, or 3 links. Ideally, we'd be able to control the label on the button, but that's not a requirement.

hannahkane:
cc @jenntbeard

gvn:
You can now add as many links (with labels) as you want by editing `portfolio.json`.

All `Links` properties can either be `null` or an array of links, following this schema:

```json
    "Links": [
      {
        "url": "http://www.example.org/",
        "label": "label goes here"
      }
    ],
```

PR #2512, opened 2017-08-22
Gigabit multiple links

gvn:
related to #2510

### DIFF
--- a/pages/gigabit/Project.jsx
+++ b/pages/gigabit/Project.jsx
@@ -13,6 +13,14 @@ module.exports = React.createClass({
       );
     });
 
+    if (data.Links) {
+      var links = data.Links.map((link) => {
+        return (
+          <a className="secondary-button m-r-2" href={link.url} target="_blank">{link.label || `Website`}</a>
+        );
+      });
+    }
+
     return (
       <div>
         <div className="m-b-3 clearfix">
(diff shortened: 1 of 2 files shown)


pages/gigabit/Project.jsx
```diff
@@ -52,12 +60,14 @@ module.exports = React.createClass({
           </div>
         </div>
 
-        <div className="m-b-3 text-center">
-          <p>Learn more about {data.Project}</p>
-          <div className="middle-button">
-            <a className="secondary-button" href={data.Link} target="_blank">Website</a>
+        { data.Links &&
```
mmmavis: Looks good but I would change this from `data.Links` to `links` since we already checked and initialized `links` based on `data.Links` on line 16.

gvn: `links` won't exist if `data.Links` is null, so that would throw an error.

mmmavis: You are right. I guess it's more of personal coding style. I usually would declare `links` outside of the `if` block.

gvn: I normally would do that as well, but in this case it doesn't need to be declared if `data.Links` is `null`. So yeah, I think it's kind of a style thing and kind of an avoiding declaring unused variables thing... `¯\_(ツ)_/¯`

